PR #2, opened 2018-04-26
Fix travis setup

sileht:


sileht:
Pull-request updated, HEAD is now 566e50bf771d7a328242e8acbcf275a94762c015

PR #26, opened 2018-05-30
Allow to configure the merge_strategy

sileht:


doc/source/configuration.rst
```diff
@@ -43,6 +43,35 @@ A ``Rule`` is a dictionary with the following optional keys:
      - A `label <https://help.github.com/articles/about-labels/>`_ name that
        will disable Mergify if it is applied to a pull request. Default is
        ``no-mergify``.
+   * - ``merge_strategy``
+     - :ref:`Merge Strategy`
+     - The method to use to merge pull requests
```
jd: final `.` missing

doc/source/configuration.rst
```diff
@@ -43,6 +43,35 @@ A ``Rule`` is a dictionary with the following optional keys:
...
+   * - Key name
+     - Value type
+     - Value description
+   * - ``method``
+     - ``string``
+     - Merge method to use. Possible values are ``merge``, ``squash`` or
+       ``rebase``. Default is ``rebase``.
+   * - ``rebase_fallback``
+     - ``string``
+     - If ``method=rebase``, but the pull request can't be rebased, this method
```
jd: `method=rebase` is not really any valid syntax so it looks weird. Rather write:

> If `method` is set to `rebase` and the pull request cannot be rebased, the method defined in `rebase_fallback` will be used instead.

doc/source/configuration.rst
```diff
@@ -43,6 +43,35 @@ A ``Rule`` is a dictionary with the following optional keys:
...
+.. list-table::
+   :header-rows: 1
+
+
+   * - Key name
+     - Value type
+     - Value description
+   * - ``method``
+     - ``string``
+     - Merge method to use. Possible values are ``merge``, ``squash`` or
```
jd: > The method to use to merge the pull request.

Is there any good reason to default to `rebase` where I think GitHub defaults to `merge`?

doc/source/configuration.rst
sileht: done

doc/source/configuration.rst
sileht: done

doc/source/configuration.rst
sileht: Because this is what we use in Gnocchi. rebase make more sense with "strict: True". We can switch back to Github default and update Gnocchi configuration.

jd: Yeah I'd say fix Gnocchi. It's confusing to change the GitHub default IMHO.

PR #28, opened 2018-05-30
doc: explain behavior when .mergify.yml is changed

sileht:


doc/source/getting-started.rst
```diff
@@ -92,6 +92,13 @@ rules.
 .. image:: _static/mergify-status-ko.png
    :alt: status check
 
+.. note::
+
+   When a pull request changes the configuration of Mergify, the ``mergify/pr``
+   status is built with the current configuration (without the pull request
+   change). To validate the Mergify configuration change a additional status is
```
jd: an additional

doc/source/getting-started.rst
```diff
@@ -92,6 +92,13 @@ rules.
...
+   posted named ``mergify/future-config-checker``
```
jd: final point.

PR #43, opened 2018-06-23
Subscription cache expiration

sileht:
Set the expiration to 1 hour. And
allow to clear it via API.


sileht:
Pull-request updated, HEAD is now 54db257c18b46043a1e1abe606fe9ebfa87fd50d

PR #46, opened 2018-06-25
Change approved computing to ignore negative, improve reporting

jd:
It does not seem like GitHub cares about negative review, so it's likely
Mergify should do the same by default. Another option could be added to not
block merge on negative reviews.

Also, send a better description on the number of approval that Mergify is
waiting for.

### DIFF
--- a/mergify_engine/gh_pr_fullifier.py
+++ b/mergify_engine/gh_pr_fullifier.py
@@ -20,7 +20,6 @@ import logging
 import time
 
 import github
-import six.moves
 
 LOG = logging.getLogger(__name__)
 
@@ -43,6 +42,13 @@ def ensure_mergable_state(pull):
 
 
 def compute_approvals(pull, **extra):
+    """Compute approvals.
+
+    :param pull: The pull request.
+    :param extra: Extra stuff
+    :return: A tuple (users_with_review_ok, users_with_review_ko,
+                      number_of_review_required)
+    """
     users_info = {}
     reviews_ok = set()
     reviews_ko = set()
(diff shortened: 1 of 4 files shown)


mergify_engine/gh_pr_fullifier.py
```diff
@@ -114,11 +108,15 @@ def disabled_by_rules(pull, **extra):
 
 def compute_weight_and_status(pull, **extra):
     status_desc = disabled_by_rules(pull, **extra)
+    reviews_ok, reviews_ko, reviews_required = pull.mergify_engine["approvals"]
     if status_desc is not None:
         weight = -1
```
sileht: Maybe we should add, to be clear:

```
if reviews_ko:
    weight = -1
    status_desc = "%d changes requested" % len(reviews_ko)
```

jd: I chose to change here to ignore reviews_ko, as IIUC GitHub does.
Not ignoring them could be another option in the yaml file.

jd:
@sileht Looks like Travis did not run!?

mergify_engine/gh_pr_fullifier.py
jd: I missed you other reply. I did not understood GitHub was counting reviews KO. Does it count it the same way as you did or just one KO blocks the entire PR?

sileht: Like I did, one KO and that blocks the PR

jd:
Pull-request updated, HEAD is now b9a62123da82f0733ede0c51c6bab67fb5296d8f

mergify_engine/gh_pr_fullifier.py
```diff
@@ -114,11 +108,18 @@ def disabled_by_rules(pull, **extra):
 
 def compute_weight_and_status(pull, **extra):
     status_desc = disabled_by_rules(pull, **extra)
+    reviews_ok, reviews_ko, reviews_required = pull.mergify_engine["approvals"]
```
sileht: I just remember that redis may have the old version with 4 items in the tuple.

jd: Do you mean a currently running Redis, or something else here in the code?
I imagine you can empty the cache?

sileht: The currently running redis.

I could empty the cache. But currently clearing the cache have one downside:

When you receive the next "pull request closed/merged" event, mergify-engine will remove this PR from the cache and then looks into the cache to choice the next PR to merge or rebase. If I clear it nothing will happen.

The workaround is the run "refresh.sh" on all installations, but that kind of exhaust github ratelimit on projects with a lot of opened pull requests.

To resume, this cache is a bit more than a cache ;)

sileht: I push https://github.com/Mergifyio/mergify-engine/commit/55a2ff009befede6d4087ee5221a2ae520d6d6ad to invalidate the cache when needed.

PR #47, opened 2018-06-26
Set default status check to pending, report success on merge

jd:


mergify_engine/engine.py
```diff
@@ -229,6 +229,9 @@ def proceed_queue(self, branch, **extra):
         LOG.info("%s selected", p.pretty())
 
         if p.mergify_engine_weight >= 11:
+            p.mergify_engine_github_post_check_status(
+                self._redis, self._installation_id, "success",
+                "Merging")
```
sileht: I would put success for any PR with  weight >= 10. Since all criteria are meet, and we are just waiting for merging them.

Maybe by looking to the weight in mergify_engine_github_post_check_status() when the status is not explicitly set.

jd:
Pull-request updated, HEAD is now 06dd2540cb8ab4bf416aad5a08d723a3276f23be

sileht:
That shouldn't be needed if we merge https://github.com/Mergifyio/mergify-engine/pull/57

sileht:
I don't think we need that anymore.

PR #56, opened 2018-06-28
Fix the selection of the pr to merge

sileht:
This change adds some check to ensure the selected pull request to
proceed is still valid.

### DIFF
--- a/mergify_engine/engine.py
+++ b/mergify_engine/engine.py
@@ -201,37 +201,72 @@ class MergifyEngine(object):
     # State machine goes here #
     ###########################
 
+    @staticmethod
+    def sort_pulls(pulls):
+        """Sort pull requests by weight and updated_at"""
+        sort_key = operator.attrgetter('mergify_engine_weight', 'updated_at')
+        return list(sorted(pulls, key=sort_key, reverse=True))
+
     def build_queue(self, branch, **extra):
+        """Return the pull requests from redis cache ordered by weight"""
+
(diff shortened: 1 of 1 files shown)


mergify_engine/engine.py
```diff
@@ -201,37 +201,72 @@ def handle(self, event_type, data):
     # State machine goes here #
     ###########################
 
+    @staticmethod
+    def sort_pulls(pulls):
```
jd: at some point we'd just need our own PullRequest class. Can we make such a class and make pygithub use it?

mergify_engine/engine.py
```diff
@@ -201,37 +201,72 @@ def handle(self, event_type, data):
...
+
+            expected_weight = p.mergify_engine_weight
+
+            # NOTE(sileht): We refresh it before processing, because the cache
+            # can be outdated, user may have manually merged the PR or weight
+            # may have changed by an event not yet received.
+
+            # FIXME(sileht): This will refresh the first pull request of the
+            # queue on each event. To limit this almost useless refresh, we
+            # should be smarted on when we call proceed_queue()
```
jd: smarter

mergify_engine/engine.py
sileht: Agreed. We currently can't make pygithub returning our object, that why I have done this crappy/lazy monkeypatching stuff. But this crap begin to become a bit fat.

PR #58, opened 2018-06-29
Ignore not required check from combined_status

sileht:
The global status of a combined_status is calculated with all checks,
not only the required one.

As we don't care about not required one. This change ignores them.


sileht:
Pull-request updated, HEAD is now 05fe98e06a2ab59028bb4edc81e03cb437932301

PR #63, opened 2018-06-29
tests: cover web side of refresh

sileht:


sileht:
Pull-request updated, HEAD is now 71008f081c23860939034f1541021a204a7def82

PR #64, opened 2018-07-02
tests: testing and coverage improvements

sileht:
* We now assert on expected events
* Remove some dead code
* Don't log merge exception on already merged PR
* Use merged instead of is_merged() when possible
* Workaround the etag/last_modified github bug
* Add more tests


sileht:
Pull-request updated, HEAD is now 1be504b3f41e1ea8c3cb75a717373eddfb3bbc17

PR #71, opened 2018-07-04
rules: fix restrictions parsing

### LINKED ISSUE #65: Mismatch between schema and docs for restrictions
The documentation for restrictions (https://doc.mergify.io/configuration.html#restriction) reads that a Restriction is "a dictionary with the following optional keys: teams/users."

However in rules.py it seems to be defined as a list (https://github.com/Mergifyio/mergify-engine/blob/master/mergify_engine/rules.py#L45): `'restrictions': voluptuous.Any(None, []),`.

At the moment when I try to define a teams restriction, it fails with: .mergify.yml is invalid: not a valid value for dictionary value @ data['rules']['default']['protection']['restrictions']".

Locally I can get it to pass if I adjust the rule to look like `'restrictions': voluptuous.Any(None, {'teams': [str], 'users': [str]}),`.

Minimum .mergify.yml to reproduce:
```
rules:
  default:
    protection:
      restrictions:
        teams:
          - core
```

jd:
Fixes #65

### DIFF
--- a/mergify_engine/rules.py
+++ b/mergify_engine/rules.py
@@ -42,7 +42,10 @@ Protection = {
             'required_approving_review_count': voluptuous.All(
                 int, voluptuous.Range(min=1, max=6)),
         }),
-    'restrictions': voluptuous.Any(None, []),
+    'restrictions': voluptuous.Any(None, {
+        voluptuous.Required('teams', default=[]): [str],
+        voluptuous.Required('users', default=[]): [str],
+    }),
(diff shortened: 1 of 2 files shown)


jd:
Pull-request updated, HEAD is now 2f1c2991b5d8d16a10d381b83a7736642b395335

jd:
Pull-request updated, HEAD is now ba3a8e8a6cc72df78edfce417d99368a8a20fb52

Issue #70, opened 2018-07-04
Python 3.7 support


jd:
Currently does not work because of:
- https://github.com/rq/rq/issues/972

PR #72, opened 2018-07-04
Pull request for fix-tox

jd:
tox: specify python3.6 for py36
tox: remove Python 3.5 from env list

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py35,py36,pep8
+envlist = py36,pep8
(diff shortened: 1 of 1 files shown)


jd:
I wish but that would not work, see https://github.com/Mergifyio/mergify-engine/issues/70.
So in the end I think it's better to specify which sub version we want :(

sileht:
Oh I got it, so just moving `basepython = python3` from `[testenv]` to `[tox]` should solve the issue for all python 3 version

PR #75, opened 2018-07-05
refresh PR when mergeable_state looks wrong

sileht:


sileht:
Pull-request updated, HEAD is now c2a327f3c152f2c7ed73ebabf23fe8225e0d8608

mergify_engine/gh_pr_fullifier.py
```diff
@@ -149,6 +155,8 @@ class MergifyState(object):
     READY = 30
 
 
+@tenacity.retry(stop=tenacity.stop_after_attempt(3),
+                retry=tenacity.retry_if_exception_type(tenacity.TryAgain))
```
jd: don't you want to set a delay like an exponential one?

PR #85, opened 2018-07-11
Don't throw backtrace on 401

sileht:
When an installation remove the access_token, we got some 401 in logs.

This change replace the backtrace by an error message.


jd:
That looks good though if you leverage daiquiri for logging I'd suggest to ahead and use more `extra` as kwargs.

PR #101, opened 2018-07-17
Pull request for sileht/blocked-gh-bug-workaround

sileht:
Add workaround about not updated mergeable_state
Don't set ALMOST_READY if the CIs fail.


sileht:
Pull-request updated, HEAD is now 7ba786e72cfbf9e9f35818756cf6efe61db90375

PR #102, opened 2018-07-18
Add support to Checks API

sileht:
We was supporting only Statuses API for required_status_check/contexts.

But Github have recently added a new Checks API. And Travis now send
Check Runs and Check Suites to Github in additional to Statuses.

As Github support now Check Runs names for required_status_check/contexts,
we should to support it, too.

To do this we have a couple of things to do:
* Looks at Check Runs to compute out "required_statuses_succeed"
* Subscribes and handles some check_suite/check_run Github events to
  refresh the Mergify status

This change does all of that.


sileht:
We can't merge it yet, see https://github.com/Mergifyio/mergify-engine/pull/102/files#diff-c20883d43327ab2e60651352ef187b4eR103

I have reached the Github support about these two issues. The new API is not really reliable for us.

I will propose a doc change to explicitly say we didn't support the Checks API because of Github bugs.

sileht:
Pull-request updated, HEAD is now 70948753f1848c7c5c3e25c33aebf192fe063dbb

mergify_engine/check_api.py
```diff
@@ -0,0 +1,94 @@
...
+        headers={'Accept': 'application/vnd.github.antiope-preview+json'}
+    )
+    return data
+
+
+@tenacity.retry(wait=tenacity.wait_exponential(multiplier=0.2),
+                stop=tenacity.stop_after_attempt(5),
+                retry=tenacity.retry_never)
+def workaround_for_unfinished_check_suite(g_repo, data):
+    """Workaround for broken Checks API events
```
jd: final dot missing

mergify_engine/check_api.py
```diff
@@ -0,0 +1,94 @@
...
+
+    The Checks API have two major flaws:
```
jd: has

mergify_engine/check_api.py
```diff
@@ -0,0 +1,94 @@
...
+    * We received check_suite/completed too early, some checks are still
+      pending
+    * Once it's completed, even if some check are rerun, we never got
```
jd: checks

mergify_engine/check_api.py
```diff
@@ -0,0 +1,94 @@
...
+      a new event to notify us that it finish again.
```
jd: finished

mergify_engine/check_api.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+    This method workaround the first issue, but the second one is still
```
jd: works around, while the second one

mergify_engine/check_api.py
```diff
@@ -0,0 +1,94 @@
...
+    unsolved
```
jd: final dot missing

sileht:
Pull-request updated, HEAD is now f2eee8a44bf63b13052d7aeed97ecc2f25d80cb9

sileht:
@jd I think we should merge this. The Github "Checks API" is not yet perfect for us, but we should at least compute our "CI combined statuses" correctly. Some CI applications have started to use it, and unlike Travis they haven't done a transition phase with both API.

For example: https://github.com/probot/dco

mergify_engine/check_api.py
```diff
@@ -0,0 +1,92 @@
...
+                stop=tenacity.stop_after_attempt(5),
+                retry=tenacity.retry_never)
+def workaround_for_unfinished_check_suite(g_repo, data):
+    """Workaround for broken Checks API events.
+
+    The Checks API have two major flaws:
+    * We received check_suite/completed too early, some checks are still
+    pending
+    * Once it's completed, even if some check are rerun, we never got
+    a new event to notify us that it finish again.
```
jd: finishes

PR #105, opened 2018-07-19
Enforce regex branch starting with ^

sileht:


sileht:
Pull-request updated, HEAD is now 47eff9d3282d40118a2004b75997f6edb709b7d6

mergify_engine/rules.py
```diff
@@ -133,7 +133,8 @@ def get_branch_rule(g_repo, branch, ref=github.GithubObject.NotSet):
     dict_merge(rule, rules.get("default", {}))
 
     for branch_re in rules.get("branches", []):
-        if re.match(branch_re, branch):
+        if (branch_re == branch or
```
jd: This looks like there might be a problem if a branch has the same name of the regexp.
Something like:

```python
if (branch_re[0] == "^" and re.match(branch_re, branch)) or (branch_re[0] != "^" and branch_re == branch)
```

would be more solid.

sileht: good catch

sileht:
Pull-request updated, HEAD is now a957bf6a879001eb0b26a2b6b3afc7ee7169c815

PR #106, opened 2018-07-19
Allow to put null for default.

sileht:


mergify_engine/rules.py
```diff
@@ -65,7 +65,7 @@
 
 UserConfigurationSchema = {
     voluptuous.Required('rules'): voluptuous.Any({
-        'default': Rule,
+        'default': voluptuous.Any(Rule, None),
```
jd: This should probably be reflected in the doc.

mergify_engine/rules.py
```diff
@@ -130,15 +128,25 @@ def get_branch_rule(g_repo, branch, ref=github.GithubObject.NotSet):
...
 
+    found_at_least_one_match = False
     for branch_re in rules.get("branches", []):
         if re.match(branch_re, branch):
             if rules["branches"][branch_re] is None:
                 LOG.info("Rule for %s branch: %s" % (branch, rule))
                 return None
             else:
                 dict_merge(rule, rules["branches"][branch_re])
+                found_at_least_one_match = True
```
jd: (nit) you could use a `for/else` I guess by replacing this by a `break.`

sileht: I just realised that something suck here, imagine say you have:

```
branches:
  ^stable/.*:
      ..
  stable/foobar:
       ...
```

and the branch is `stable/foobar`. Then both rules are merged.

But because `branches` is a dict and python dict are not ordered, the merge order is kind rolling a dice...

sileht:
Pull-request updated, HEAD is now dadc52a6e3e8d8ca7ebaa8b4f33bbd57764db978

sileht:
Pull-request updated, HEAD is now cfe11182ee9b29a0590ac02a15eacf6446c8a0d2

mergify_engine/rules.py
sileht: I also fixed that.

sileht:
Pull-request updated, HEAD is now 708add38d864686b7e84eedc108b87ffa8f097cd

sileht:
Pull-request updated, HEAD is now 7efde0bc2d57b26e7fccc00a88c4cc9a5f63b7a3

PR #108, opened 2018-07-19
Add enabling_label option

### LINKED ISSUE #107: Add enabling_label option
Some users do not want Mergify to automerge if the number of reviews matching is ok. However, they are happy to see the patch merge if the CI passes + reviews are OK + a label is set.

In the same vein that Mergify offers `disabling_label`, having a `enabling_label` that would be required for Mergify to be active, would help some users.

(Think of this as the +2 Approved of Gerrit)

jd:
Fixes #107


jd:
record-all ran

PR #120, opened 2018-07-24
Pass subscription to processor

sileht:


mergify_engine/engine.py
```diff
@@ -72,7 +72,6 @@ def __init__(self, g, installation_id, installation_token, subscription,
                                             redis=utils.get_redis_for_cache())
         self._g = g
         self._installation_token = installation_token
-        self._subscription = subscription
```
jd: you need to keep this

sileht:
Pull-request updated, HEAD is now d1abb07b50de3b8942830853da9d231e47fb60a2

PR #123, opened 2018-07-24
Remove pygithub monkeypatching

sileht:
Remove pygithub monkeypatching
rename gh_branch to branch_protection


sileht:
Pull-request updated, HEAD is now eab09ff3596f76223c81a05abdd2df4cf9102b8d

PR #133, opened 2018-07-26
Use Enum in Mergify pull request objects

jd:


mergify_engine/mergify_pull.py
```diff
@@ -120,16 +132,16 @@ def complete(self, cache, **context):
             self._approvals = self._compute_approvals(**context)
 
         if "mergify_engine_required_statuses" in cache:
-            self._required_statuses = cache[
-                "mergify_engine_required_statuses"]
+            self._required_statuses = StatusState(cache[
```
sileht: Maybe convert old cache integer to the string (and put need_to_be_saved to True) ? or do you prefer I refresh the cache of all installations once it's merged ?

jd: Since I'd prefer to keep it simple I'd vote for clearing the cache, but if you think it's a burden, I can be convinced otherwise.

jd:
Fix validator

mergify_engine/mergify_pull.py
sileht: Looks like we didn't hit the rate limit anymore, I will refresh the cache

PR #134, opened 2018-07-26
Use daiquiri and extra prepending to log

jd:
This changes the logging mechanism to leverage daiquiri and log a bit better.
- It passes most context information as keyword arguments so they are printed
  correctly on the line and can be embedded to Sentry
- It uses a log object per MergifyPull object so it's automatically prepended
  with information on the pull request logging

### DIFF
--- a/mergify_engine/backports.py
+++ b/mergify_engine/backports.py
@@ -13,15 +13,15 @@
 # under the License.
 
 import functools
-import logging
 import subprocess
 
+import daiquiri
 import github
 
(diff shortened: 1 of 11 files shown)


sileht:
Hum, I have checked the log and extra are not printed

sileht:
I fixed it https://github.com/Mergifyio/mergify-engine/commit/5c891c37aae45c71a88ed7015e163f52a90ca986

PR #136, opened 2018-07-27
Retry on HTTP error 5XX

sileht:
This change retries on HTTP error 5XX with an exponential backoff
starting from 5 seconds.


sileht:
Pull-request updated, HEAD is now 15ec8ee8245a8fd2f7cdf5b0dfff28c32a656c09

PR #140, opened 2018-07-30
Don't store required_reviews in cache

sileht:


sileht:
Pull-request updated, HEAD is now b0b5c7ff28eb24cd958b1c0ee163b68507899f1f

PR #142, opened 2018-08-03
Pull request for sileht/merge-raise-405

sileht:
Don't reuse pull and trust etag
log: don't override message in logging


sileht:
This should make [1] issue occurs less and set the status check as "Merge failed" as before when Github lies to us.

The root cause is still the mergeable_state and the status checks list that are not in sync...
I have reported the issue to Github one month ago, and re-reach them two weeks ago about this.
They still tell me "we are working on it" ....

[1] https://github.com/awesomeWM/awesome/pull/2322

sileht:
Pull-request updated, HEAD is now 0e2f3725ea1308d5a9e71818f783a2c16ca0722c

PR #143, opened 2018-08-13
Enable support for Python 3.7

### LINKED ISSUE #70: Python 3.7 support

jd:
Fixes #70


jd:
merged in https://github.com/Mergifyio/mergify-engine/pull/145

PR #151, opened 2018-08-24
Pull request for delete-branch-on-merge

jd:
engine: implement branch deletion after merge
Factorize branch creation in tests


mergify_engine/mergify_pull.py
```diff
@@ -402,6 +402,21 @@ def _merge_failed(self, e):
                        exc_info=True)
         return False
 
+    def delete_head_branch(self):
```
sileht: I also delete head branch here: https://github.com/Mergifyio/mergify-engine/blob/master/mergify_engine/engine.py#L242

I'm sure we can factorize both.

jd:
Closing as v1's deprecated.

PR #162, opened 2018-08-31
Expose some metrics for prometheus

sileht:


mergify_engine/stats.py
```diff
@@ -17,14 +17,50 @@
 
 import github
 
+import prometheus_client
+
 from mergify_engine import config
 from mergify_engine import utils
 
+GITHUB_EVENTS = prometheus_client.Counter(
+    "github_events", "Number of Github events handled")
```
jd: GitHub

PR #164, opened 2018-09-03
pull: merge() use explicit arguments instead of a branch rule

jd:


jd:
Pull-request updated, HEAD is now 63cd2f91ed83cb08576d81b6f581714b3736d5c2

PR #171, opened 2018-09-12
stats: create a prometheus exporter

sileht:
This change removes the usage of multiprocess system of
prometheus_client. It creates a dedicated prometheus exporter instead.


sileht:
Pull-request updated, HEAD is now bb3cb45b5f3996c10480b86b0c335b336a23041e

sileht:
Pull-request updated, HEAD is now 07669311cd18cbf8e449fc367f9a96dfb0107662

PR #172, opened 2018-09-12
Revert "rules: factorize default config merge"

sileht:
This reverts commit 2832d194224370d96f24023b8e0a4bec1ef04c6c.


mkarp:
If that's related to the `Mergify configuration is invalid: required key not provided @ data['protection']['required_pull_request_reviews']['require_code_owner_reviews']` error, then thank you! :)

PR #173, opened 2018-09-12
rules: factorize default config merge and add missing test

jd:
This reverts commit 6e1af1a7df624a435fe20a524bb0833f35d1c68a with a fix and a
test case for it.


jd:
Fix pep8

PR #177, opened 2018-09-13
Add prom_exporter test

sileht:
To do so, tests have been split into two directories,
a base class have been created for functional testing.


sileht:
Pull-request updated, HEAD is now 12c2ce4e326fb154f6cdde8b98a6001421cdc1a3

PR #180, opened 2018-09-14
prom-exporter: Set all Gauges at the same time

sileht:


mergify_engine/prom_exporter.py
```diff
@@ -80,7 +81,16 @@ def collect_metrics():
...
+            value = len(list(repos))
+            todo.append((c.set, (value, )))
+
+    # Prometheus can scrape data during our loop. So make it fast
+    # to ensure we always have the good values
+    INSTALLATIONS._metrics = []
+    REPOSITORY_PER_INSTALLATION._metrics = []
+    USERS_PER_INSTALLATION._metrics = []
+    for method, args in todo:
+        method(*args)
```
jd: This seems quite complicated. Why don't you just do:

```python
new_metrics = {}

<fill metric as before>

INSTALLATIONS._metrics = new_metrics
```

sileht: _metrics is a list of so complicated object.

PR #190, opened 2018-09-18
rules: use None rather a string in rebase fallback new rules

jd:


jd:
Pull-request updated, HEAD is now 35c20f5719e9fec3aed25d0c1473da2c1f86f0ea

PR #201, opened 2018-10-03
Move to engine v2

sileht:


.mergify.yml
```diff
@@ -1,14 +1,11 @@
...
-
-  branches:
-    gh-pages: null
+pull_request_rules:
+  - name: automatic merge
+    conditions:
+      - base=master
+      - status-success=continuous-integration/travis-ci/pr
+      - "#approved-reviews-by>=1"
+      - label!=no-mergify
```
jd: Can we use something like `work-in-progress` instead?

sileht:
Pull-request updated, HEAD is now a56d7f6f50829e4b2157326846318855f118dd85

PR #209, opened 2018-10-04
convert: remove 'merged' from conditions

jd:
That is not needed since backport only affect merged PRs.


jd:
Pull-request updated, HEAD is now 3d522686c7a4bcdac6269ec078565421b78ceed5

PR #216, opened 2018-10-05
Remove initial request creation

sileht:
This features only generates v1 configuration and will be move on site
side, with wizard. So remove it.


sileht:
Pull-request updated, HEAD is now fdba24048fe3ed9188892d3433dee4486fdc570a

PR #218, opened 2018-10-06
v2: only keep last valid reviews

sileht:


sileht:
Pull-request updated, HEAD is now ec1a1b217c11ddb4c3a917358d2140b460d0be06

PR #219, opened 2018-10-06
Add dismiss_reviews action

sileht:


mergify_engine/actions/dismiss_reviews.py
```diff
@@ -0,0 +1,55 @@
...
+import github
+
+import voluptuous
+
+from mergify_engine import actions
+
+
+class DismissReviewsAction(actions.Action):
+    validator = {
+        voluptuous.Required("approved", default=[]):
```
jd: I'd think we'd set `True` by default so if the action is enabled then it dismisses everything by default.

mergify_engine/actions/dismiss_reviews.py
```diff
@@ -0,0 +1,55 @@
...
+                    self._dismissal_review(pull, review)
+
+    @staticmethod
+    def _dismissal_review(pull, review):
+        try:
+            review._requester.requestJsonAndCheck(
+                "PUT",
+                "%s/reviews/%s/dismissals" % (review.pull_request_url,
+                                              review.id),
+                input={'message': "Pull request have been modified."},
```
jd: has been

mergify_engine/actions/dismiss_reviews.py
```diff
@@ -0,0 +1,55 @@
...
+                headers={'Accept':
+                         'application/vnd.github.machine-man-preview+json'}
+            )
+        except github.GithubException as e:  # pragma: no cover
+            pull.log.error("failed to dismiss review",
+                           error=e.data["message"], exc_info=True)
```
jd: do you really need `exc_info=True`?

PR #220, opened 2018-10-06
Change summary style

sileht:
* Use title as dedicated check
* use GitHub CheckBox instead of emojis


sileht:
Pull-request updated, HEAD is now 98f819d1ae5ef772f545ad5dfe9da1351b6924a4

PR #221, opened 2018-10-07
Add close action

sileht:


mergify_engine/actions/close.py
```diff
@@ -0,0 +1,52 @@
...
+# under the License.
+
+import github
+
+import voluptuous
+
+from mergify_engine import actions
+
+
+MSG = "This pull request have been automatically close by Mergify."
```
jd: has
closed

mergify_engine/actions/close.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __call__(self, installation_id, installation_token, subscription,
+                 event_type, data, pull):
+        if pull.g_pull.state == "close":
+            return
+
+        try:
+            pull.g_pull.edit(state="close")
+        except github.GithubException as e:
+            pull.log.error("fail to close the pull request")
```
jd: maybe log the details as to why?

mergify_engine/actions/close.py
```diff
@@ -0,0 +1,52 @@
...
+            return ("failure", "Pull request can't be closed", "")
+
+        try:
+            pull.g_pull.create_issue_comment(self.config["message"])
```
jd: Actually I'd think it's better to add a custom "comment" action where anyone can add a comment, even for close, with a custom message.

sileht:
Pull-request updated, HEAD is now e7ca4c70d2335d0aa56372c175b41d0da94d7aa1

PR #228, opened 2018-10-09
delete head branch of mergify backport

sileht:


mergify_engine/tasks/engine/v2.py
```diff
@@ -141,6 +141,13 @@ def run_actions(installation_id, installation_token, subscription,
 def handle(installation_id, installation_token, subscription,
            pull_request_rules_raw, event_type, data, pull_raw):
 
+    # Some mandatory rules
```
jd: It makes the engine depends on a plugin (`delete_head_branch`).

mergify_engine/tasks/engine/v2.py
```diff
@@ -141,6 +141,13 @@ def run_actions(installation_id, installation_token, subscription,
...
+    pull_request_rules_raw["rules"].extend([
+        {"name": "delete backport branch (Mergify rule)",
+         "conditions": ["head~=^mergify/bp/", "closed"],
```
jd: Maybe add also author?
Not sure we have a config variable available with that author name available though.

sileht:
Pull-request updated, HEAD is now 80216ca5820f53a8a0f024abd78201dd6d3a319f

mergify_engine/tasks/engine/v2.py
sileht: I have moved this into a configuration file.

sileht:
If I move that into action backport, I must add a hook that exec on every evaluation.

sileht:
Pull-request updated, HEAD is now d574eb481b16da08209e88816cc79b416fbebc0f

PR #235, opened 2018-10-11
rules: fix white space parsing

jd:
This changes the status success/failure to match against any string that follow
the operator. The text parsing is now done with whatever chars are after the
operator.


jd:
Pull-request updated, HEAD is now 3c1e2420d730ebf92ea468192cfc35a98bf1c087

Issue #66, opened 2018-07-04
Only merge if all required reviewers approved a PR


jd:
Stolen from https://github.com/isaacs/github/issues/1205:

> I would like to have a flag on the repo that when enabled requires all assigned reviewers to approve the PR before it can be merged.

> We are currently using the specific number of reviews required, but this is too rigid for our workflow. For us, the number of reviewers required depends on what was touched by the PR.

> We have automation that assigns the necessary reviewers to our PRs. Having a simple check that required all assigned reviewers to approve the PR is the last piece of our workflow puzzle.

PR #254, opened 2018-10-16
Added requested-reviews-approved condition

edahlseng:
This pull request addresses #66 by adding a new boolean rule called `requested-reviews-approved`. A few notes:
* I don't know how to test or debug Mergify (I didn't see any CONTRIBUTING information that explains how to do this – let me know if I just missed it!). Any help with verifying that it works as expected would be helpful!
* Did I update everything that needs to be updated? I added the data in `mergify_engine/mergify_pull.py`, the search rule in `mergify_engine/rules/parser.py` and the docs in `doc/source/conditions.rst`. Anything else?
* Currently this condition ignores teams. (As I understand it, additional lookups would be needed to get the list of users for each team in order to confirm whether or not the entire team has approved).

### DIFF
--- a/doc/source/examples.rst
+++ b/doc/source/examples.rst
@@ -199,3 +199,25 @@ as the pull request gets updated with new commits.
           - base=master
         actions:
           dismiss_reviews:
+
+Require All Requested Reviews to Be Approved
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If all requested reviews have been approved, then the number of
+``review-requested``, ``changes-requested-reviews-by``, and
+``commented-reviews-by`` will all be 0.
+
+.. code-block:: yaml
+
+    pull_request_rules:
+      - name: remove outdated reviews
+        conditions:
+          - "#review-requested=0"
+          - "#changes-requested-reviews-by=0"
+          - "#commented-reviews-by=0"
+        actions:
+            merge:
+              method: merge
+
+Note that if a requested review is dismissed, then it doesn't count as a review
+that would prevent the merge.


sileht:
Thanks, that's look good for me.

About testing, we didn't have unit tests around this method, we only cover it with integration tests that requires special GitHub testing accounts to generate.

About teams, the required API was not available until Monday for GitHup App. The library we use  `pygithub` does not support it yet. So I'm ok to look into that later.

jd:
@edahlseng We just tested and it seems that the variable `review-requested` gets empty once all the reviews requested are done by the users. So you basically just need to use a condition like `#review-requested=0` to merge. Can you test it out and let us know if that's good for you? That'd make this PR useless. :)

jd:
@edahlseng any success? :)

edahlseng:
@jd yes indeed, that works well! I've updated this PR to include an additional example that highlights this!

doc/source/examples.rst
```diff
@@ -199,3 +199,24 @@ as the pull request gets updated with new commits.
...
+``review-requested`` and the number of ``changes-requested-reviews-by`` will
+both be 0.
+
+.. code-block:: yaml
+
+    pull_request_rules:
+      - name: remove outdated reviews
+        conditions:
+          - "#review-requested=0"
+          - "#changes-requested-reviews-by=0"
```
jd: Hum I think that'd merge even with 0 review since if you did not request any review `#review-requested=0` and since nobody asked for a change `#changes-requested-reviews-by=0` is true.

doc/source/examples.rst
```diff
@@ -199,3 +199,24 @@ as the pull request gets updated with new commits.
...
+        actions:
+            merge:
+              method: merge
+
+Note that if a requested review is dismissed, then it does count as a review
+that would prevent the merge.
```
jd: How's that? You mean `changes-requested-reviews-by` contains something?
That'd be weird.

doc/source/examples.rst
edahlseng: That's correct, it would. Though I think for this example, that makes sense – it's only verifying that all requested reviews have been approved. It doesn't make any assumptions about whether a review is required in the first place or not.

doc/source/examples.rst
edahlseng: Awkward, that was a typo. It should say _doesn't_.

edahlseng:
I ran into an interesting edge case today – if a review leaves comments, but does not explicitly approve or request changes, then the PR will still be merged. Does GitHub track that as a separate state as well?

doc/source/examples.rst
jd: Yeah ok, I thought it was confusing but re-reading it seems ok to me. :)

jd:
@edahlseng it makes sense to me, the review request has been completed so the requestee gets out of `review-requested`.

If what you want is a way to say "I want an approve from those 2 guys", I don't think there's a way in GitHub to do that in the end.

I've updated your branch with a fix for the last sentence.

edahlseng:
@jd This example needed one more change – setting #comment-reviews-by to 0 as well. I've just updated the PR.

PR #256, opened 2018-10-16
Move out MergifyPull v1 stuff to v1 engine

sileht:
Many methods and attributes are used only by v1 engine.
This change moves them there.


sileht:
Pull-request updated, HEAD is now 637a732d6c157eec448c793b30458d7efa212f64

PR #260, opened 2018-10-16
Allow to convert config file with API

sileht:


sileht:
Pull-request updated, HEAD is now b74c70e450969a4525eaaeed3d15bbe89e27f19a

PR #284, opened 2018-10-24
s/dictionnary/dictionary/

bowlofeggs:
Signed-off-by: Randy Barlow <randy@electronsweatshop.com>


bowlofeggs:
The CI failure seems unrelated to my change - looks like some linter failures in the code and I only changed the docs.

sileht:
A new flake8 have just been release and break us. I propose a fix here: https://github.com/Mergifyio/mergify-engine/pull/285

PR #289, opened 2018-11-06
doc: fix example name and conditions

jd:


jd:
Rebase

jd:
@sileht you need to approve it

PR #302, opened 2018-11-27
Always get a new token when task start

sileht:
When a task is retried (error 500, ...), we reuse the installation_token
that have good chance to have expired.

This change uses a fresh token at the begin of each tasks.


sileht:
Pull-request updated, HEAD is now ca88f61056e3374f117fb8b6f7b147e4c0edeb02

Issue #304, opened 2018-11-28
Wrong failed status on Checks page


renatocaval:
I'm experimenting with different rules on this repo:
https://github.com/renatocaval/hello

The rules are working as we can see on this screenshot:

![screenshot 2018-11-28 at 14 21 23](https://user-images.githubusercontent.com/502982/49154676-56c0ad80-f319-11e8-81ba-640a49532ebc.png)

However, the Checks page is showing errors. 

![screenshot 2018-11-28 at 14 22 16](https://user-images.githubusercontent.com/502982/49154712-72c44f00-f319-11e8-80da-8940ae08feb2.png)

Yesterday I also did a few tests, but didn't get those errors.

PR #305, opened 2018-11-28
label: fix issue when label have been deleted in the meantime

sileht:


renatocaval:
Is that a fix for #304?

sileht:
Only one part of it.

sileht:
There is this one too: https://github.com/Mergifyio/mergify-engine/pull/307

Issue #310, opened 2018-12-04
GitHub teams references in the approved-reviews-by condition


bowlofeggs:
Greetings!

I am attempting to configure my project with a rule that matches on the @fedora-infra/bodhi team as a reviewer, like this:

```
- approved-reviews-by=[@fedora-infra/bodhi]
```

That is based on my understanding of the [docs](https://doc.mergify.io/conditions.html#attribute-list), which at the time of this writing say "approved-reviews-by | array of string | The list of GitHub user or team login that approved the pull request. Team logins are prefixed with the @ character. This will match only reviewers with admin or write permission on the repository."

When I try this out, the mergify config checker fails and complains about the use of the @ character:

```
Mergify configuration is invalid: extra keys not allowed @ data['pull_request_rules'], Invalid condition 'approved-reviews-by=[@fedora-infra/bodhi]'. Expected end of text (at char 34), (line:1, col:35) @ data['pull_request_rules'][1]['conditions'][2]
```

Am I formatting it incorrectly?

bowlofeggs:
I did try @bodhi and the config checker seems to be OK with it (it's when the / is present that it gets upset it seems, despite the error about @), but then it doesn't seem to match and actually merge PRs when they are reviewed by members of the bodhi team.

bowlofeggs:
I am also having trouble with another rule with this match line:

```
- approved-reviews-by=[bowlofeggs]
```

A pull request I've reviewed is not getting merged, and Mergify says that this rule isn't satisfied. The docs say this can be an "array of string" - is the syntax I used for an array of strings not correct here?

jd:
You should just use `- approved-reviews-by=bowlofeggs`. If the value is an array of string, the `=` operator acts as a `in` operator.

bowlofeggs:
What is the syntax for expressing an array of strings?

bowlofeggs:
I tried ```- approved-reviews-by=@fedora-infra/bodhi``` and Mergify says the [syntax is invalid](https://github.com/fedora-infra/bodhi/pull/2814/checks?check_run_id=37944764):

```
Mergify configuration is invalid: extra keys not allowed @ data['pull_request_rules'], Invalid condition 'approved-reviews-by=@fedora-infra/bodhi'. Expected end of text (at char 33), (line:1, col:34) @ data['pull_request_rules'][1]['conditions'][2]
```

jd:
[...]

PR #316, opened 2018-12-06
Support teams

sileht:
Where we can put github login, we can also put teams now.

This also change the tests setup to use an Organization account instead
of an User account.

Related: #310


mergify_engine/mergify_pull.py
```diff
@@ -168,6 +169,42 @@ def _get_checks(self):
...
+
+        try:
+            g_organization = self.g.get_organization(organization[1:])
+            for team in g_organization.get_teams():
+                if team.slug == team_slug:
+                    return [m.login for m in team.get_members()]
+        except github.GithubException as e:
+            if e.status == 400:
+                return
+            elif e.status >= 500:
```
jd: why do you raise if >= 500 ? If you get 5xx it seems like it's a GitHub issue, I'd say we just want to log something, maybe at a the error level.

```python
if e.status != 400:
    self.log.error(…)


return []
```

mergify_engine/mergify_pull.py
```diff
@@ -168,6 +169,42 @@ def _get_checks(self):
                                for s in self._get_statuses()])
         return generic_checks
 
+    def _get_team_members(self, name):
+        organization, _, team_slug = name.partition("/")
+        if organization[0] != "@" or not team_slug or '/' in team_slug:
+            # Not a team slug
+            return
```
jd: I'd say rename this method to `_resolve_login` and change this line to `return [name]`.
You can then simplify `resolve_teams`.

mergify_engine/mergify_pull.py
```diff
@@ -168,6 +169,42 @@ def _get_checks(self):
...
+
+    def resolve_teams(self, values):
+        if not values:
+            return []
+        if not isinstance(values, (list, tuple)):
+            values = [values]
+
+        gh_logins = []
+        for value in values:
+            team_members = self._get_team_members(value)
```
jd: if you change `_get_team_members` as suggest you can just do
gh_logins.extend(self._resolve_login(value))

mergify_engine/mergify_pull.py
sileht: I reraise any 500 from GitHub because we have generic celery task retrying in such case:

https://github.com/Mergifyio/mergify-engine/blob/master/mergify_engine/worker.py#L63

mergify_engine/mergify_pull.py
sileht: done.

sileht:
Pull-request updated, HEAD is now cdf02fe5b83b1652e6761353a2e1c545e454744c

mergify_engine/mergify_pull.py
```diff
@@ -168,6 +170,33 @@ def _get_checks(self):
...
+                             detail=e.data["message"])
+        return [name]
+
+    def resolve_teams(self, values):
+        if not values:
+            return []
+        if not isinstance(values, (list, tuple)):
+            values = [values]
+        return list(itertools.chain.from_iterable((
+            self._resolve_login(value) for value in values)))
```
jd: you missed a good opportunity to use `map`! :p

PR #324, opened 2018-12-13
merge: allow to have status check about the mergeability

sileht:
This restores a V1 engine functionality. User can configure Mergify to
set the Mergify Summary status with the combined state of all rules with
merge action.


sileht:
Pull-request updated, HEAD is now dae716eb5a35c7d0a33a2998b217f80e31b1c7c6

jd:
@sileht do we still want this?

PR #330, opened 2018-12-18
Pull request for quoted-CI

sileht:
doc: use checks instead of status in some example
rules: allow quoted string for text


mergify_engine/rules/parser.py
```diff
@@ -21,7 +21,9 @@
 github_login = pyparsing.CharsNotIn(" /@")
 github_team = pyparsing.Combine(pyparsing.Literal("@") + github_login +
                                 pyparsing.Literal("/") + github_login)
-text = pyparsing.CharsNotIn("")
+text = (pyparsing.QuotedString('"') |
```
jd: Is this really necessary?
Isn't it adding more confusion? Maybe not. I wonder.

sileht: That's mainly because most people expect to have quote for a string after a `=`, like any programming language.

For me, it's `pyparsing.CharsNotIn("")` that confuses me, but we cannot remove it people is using it.

I even thinking about putting quote everywhere in the documentation.

PR #342, opened 2019-01-29
doc: remove unused images

jd:


sileht:
That's not really obvious but their are used here: https://github.com/Mergifyio/mergify-engine/blob/master/mergify_engine/web.py#L68

jd:
Pull-request updated, HEAD is now 2eb0657b051dcab52f573305180c463b9b33a85f

jd:
Rebased

PR #351, opened 2019-02-20
Pull request for master

sileht:
Encrypt token in cache
remove usage of search API


sileht:
Pull-request updated, HEAD is now 260ce318baadf70f53ee34433a2091ef204b289f

PR #374, opened 2019-04-15
branch_updater: Fallback to unshallowed clone

sileht:
When "git" fails with "unrelated history" error, retries with
a full git clone.

### DIFF
--- a/mergify_engine/branch_updater.py
+++ b/mergify_engine/branch_updater.py
@@ -14,6 +14,8 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import subprocess
+
 import daiquiri
 
 from mergify_engine import config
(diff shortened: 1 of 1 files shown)


sileht:
Pull-request updated, HEAD is now c79dbda1d48cf32ea87bc0b8591fde7bc18a05e2

mergify_engine/branch_updater.py
```diff
@@ -71,18 +85,23 @@ def update(pull, installation_id, method="merge"):
...
-            git("push", "--quiet", "origin", head_branch)
-        elif method == "rebase":
-            git("rebase", "upstream/%s" % base_branch)
-            git("push", "--quiet", "origin", head_branch, "-f")
-        else:
-            raise RuntimeError("Invalid branch update method")
+        try:
+            _do_update(git, method, base_branch, head_branch)
+        except subprocess.CalledProcessError as e:
+            if "unrelated histories" in e.output:
```
jd: I don't think this appears when you use the `rebase` method.

sileht: I haven't reproduced the issue, so I put the only error message we have.

PR #375, opened 2019-04-15
Handle rerequest button of checks

sileht:


sileht:
Pull-request updated, HEAD is now b841baa1bacd1d079799708771c41c5142346284

PR #377, opened 2019-04-16
Remove Mergify prefix from checks title and regroup most checks under Summary

jd:


sileht:
This have good chance to annoy people. We should do something for the transition.

Like getting the old check name if exists we use the prefix, if not we use the new name.

sileht:
Pour une transition en douceur, je pensais dans le cas ou le "name" match pas celui déjà affiché dans Github, le detecté ici:

https://github.com/Mergifyio/mergify-engine/blob/master/mergify_engine/check_api.py#L147-L148

Et rajouter "Mergify - " is besoin. Dans 2/3 mois, on virera juste de petit workaround pour la transition.

jd:
Yeah, that sounds like a good idea. I'll do that once #370 is merged.

PR #378, opened 2019-04-17
deploy: fix bash error

jd:
When deploying, the output shows:
bash: -c: option requires an argument

This does not prevent the command to be executed. This happened because the
first line is empty, so just remove the \n before the first command to make
this disappear.

### DIFF
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -25,8 +25,7 @@ docker push mergifyio/engine:latest
 # TODO(sileht): nginx cache the ip of the engine container, when the
 # engine container is recreated, nginx still try to connect to the old one
 # So we temporary restart the nginx container each time we deploy mergify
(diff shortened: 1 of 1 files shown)


tools/deploy.sh
```diff
@@ -25,8 +25,7 @@ docker push mergifyio/engine:latest
 # TODO(sileht): nginx cache the ip of the engine container, when the
 # engine container is recreated, nginx still try to connect to the old one
 # So we temporary restart the nginx container each time we deploy mergify
-ssh $PRODUCTION_HOST -p $PRODUCTION_PORT bash -c "
-set -euxo pipefail;
+ssh $PRODUCTION_HOST -p $PRODUCTION_PORT bash -c "set -euxo pipefail;
```
sileht: We should just remove -c

sileht: Oh It manually run bash, sorry

PR #382, opened 2019-04-25
Check validity of token in subscription cache

sileht:


sileht:
Pull-request updated, HEAD is now 1feec648dbf50dbb3c46fc995b24f26cede0c182

PR #385, opened 2019-04-29
Handle new tokens format from website

sileht:
This allows to retry the branch update with all tokens available.

### DIFF
--- a/mergify_engine/branch_updater.py
+++ b/mergify_engine/branch_updater.py
@@ -24,8 +24,17 @@ from mergify_engine import utils
 
 LOG = daiquiri.getLogger(__name__)
 
+AUTHENTICATION_FAILURE_MESSAGES = (
+    b"Invalid username or password",
+    b"The requested URL returned error: 403",
+)
 
-def _do_update(git, method, base_branch, head_branch):
+
+class AuthentificationFailure(Exception):
+    pass
+
+
+def _do_update_branch(git, method, base_branch, head_branch):
     if method == "merge":
         git("merge", "--quiet", "upstream/%s" % base_branch, "-m",
             "Merge branch '%s' into '%s'" % (base_branch, head_branch))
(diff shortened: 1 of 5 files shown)


mergify_engine/branch_updater.py
```diff
@@ -24,8 +24,17 @@
 
 LOG = daiquiri.getLogger(__name__)
 
+AUTHENTICATION_FAILURE_MESSAGES = [
```
jd: nit: `set()`:p

mergify_engine/branch_updater.py
```diff
@@ -99,15 +98,39 @@ def update(pull, installation_id, method="merge"):
...
+
+    if not subscription:
+        LOG.error("unable to update branch: subscription is missing")
+        return
+
+    for login, token in subscription["tokens"].items():
+        try:
+            return _do_update(pull, token, method)
+        except AuthentificationFailure as e:
+            LOG.error("authentification failure: %s", e,
```
jd: This is likely to mislead us when reading the log. We'll think something's wrong when it's just trying to re-auth with another token. Maybe use `debug` and make the sentence clearer.

We should log an `error` only if *all* tokens fail.

mergify_engine/branch_updater.py
```diff
@@ -99,15 +98,39 @@ def update(pull, installation_id, method="merge"):
...
         git.cleanup()
+
+
+def update(pull, installation_id, method="merge"):
+    redis = utils.get_redis_for_cache()
+
+    subscription = sub_utils.get_subscription(redis, installation_id)
+
+    if not subscription:
+        LOG.error("unable to update branch: subscription is missing")
```
jd: nit: I feel like `subscription` is misleading because it'd make me think that the problem is that`subscription` is `False` where this just means that there's a 404 or the like on the installation from the dashboard.

sileht: In fact, this cannot occurs, `get_subscription` always returns a `dict`

mergify_engine/branch_updater.py
sileht: done

mergify_engine/branch_updater.py
sileht: or even a tuple ;)

sileht:
Pull-request updated, HEAD is now b971e00e07e23febc7502e69abad3f27fba444b7

mergify_engine/branch_updater.py
jd: Well a `tuple` is ordered, whereas a `set` is not. In that case we don't care about order, a `set` makes that clear. ;)

PR #388, opened 2019-04-30
debug: print the number of cached tokens and catch auth GithubException

jd:
Fixes MERGIFY-ENGINE-KT


mergify_engine/debug.py
```diff
@@ -94,10 +96,15 @@ def report(url):
             except github.BadCredentialsException as e:
                 exception = e
                 continue
```
sileht: This continue looks useless.

PR #389, opened 2019-04-30
debug: fix no token error

jd:
Right now it would be printed as soon as one token fails.

Only print it if ALL tokens fail.


jd:
Fix pep8

PR #390, opened 2019-05-02
Filter some exceptions with sentry

sileht:
We currently send all exceptions to Sentry even those
that we retry with Celery.

This change ensures that exception that need to be retried
will not send event to sentry before all retries are done.


sileht:
Pull-request updated, HEAD is now faa2d730d8a2c3af14e35ae377c67a43398f1341

sileht:
Pull-request updated, HEAD is now 5cb5351ebf195907a046240e9d269df083292d4d

Issue #37, opened 2018-06-17
Allow to customize automatic backports


sileht:
We should allow user to configure associated labels and title.

renatocaval:
We are especially interested in the possibility of having the title configurable. 

But not only for backports. Also for automatic merges. 

The problem that we see is that GitHub UI treats merges different from squashes. 

If it's a squash, by default, GitHub uses the title of the PR + (#num):

```
Added support for feature X (#10) 
``` 

while a merge, will produce something similar to:

```
Merge pull request #11 from renatocaval/to-merge
```

The backport in mergify, follows the later:

```
Merge pull request #12 from renatocaval/mergify/bp/1.0.x/pr-10
```

It will be great if we could configure it to produce:

* Added support for feature X (#10) - for squash
* Added support for feature X (#10) - for merge
* Added support for feature X (#10 backported) - for backport

Or a variation of this, eg:

* Added support for feature X (#10 squashed)
* Added support for feature X (#10 merged)
* Added support for feature X (#10 backported)

Where `Added support for feature X` is the original title of the PR.

WDYT?

ignasi35:
See https://github.com/Mergifyio/mergify-engine/pull/392

leseb:
Being able to also customize the PR body would be nice. Many CIs are parsing the content and if they find something like `[test foo]`, that would run the `foo` pipeline instead. We could do that we labels, if the original PR has a given label then we match this as a message in backport PR.

Thanks!

jd:
@renatocaval you can now specify the commit message from the body https://doc.mergify.io/actions.html#defining-the-commit-message or from the PR itself with `commit_message: title+body` so I think everything's covered in your case?

jd:
> Being able to also customize the PR body would be nice. Many CIs are parsing the content and if they find something like [test foo], that would run the foo pipeline instead. We could do that we labels, if the original PR has a given label then we match this as a message in backport PR.

I've never seen a CI parsing the body for anything. Do you have a live example of that?

I imagine we could an option to duplicate (some) of the labels too.

leseb:
> > Being able to also customize the PR body would be nice. Many CIs are parsing the content and if they find something like [test foo], that would run the foo pipeline instead. We could do that we labels, if the original PR has a given label then we match this as a message in backport PR.
> 
> @leseb I've never seen a CI parsing the body for anything. Do you have a live example of that?
> 
> I imagine we could an option to duplicate (some) of the labels too.

[...]

PR #392, opened 2019-05-07
Replace generic backport title with original PR's title

ignasi35:
This is a first solution for #37 (only tries to solve the issue for backports).

(disclaimer: please note this is my first time navigating this codebase or using the python ecosystem. I think my env is not fully setup as `./run_tests.py` seems to block. Any help will be very welcome.)


ignasi35:
cc @renatocaval

mergify_engine/backports.py
```diff
@@ -161,7 +161,7 @@ def backport(pull, branch, installation_token):
             "checking-out-pull-requests-locally/")
 
     return repo.create_pull(
-        title="Automatic backport of pull request #%d" % pull.g_pull.number,
+        title="[bp] %s" % pull.g_pull.title,
```
ignasi35: This solution is not 100% correct.

GitHub automatically trims the text from the title of the PR in order to append `(#pr-num)`. Since this change appends `[bp]` (to signal it's a `backport`) it is possible the resulting title overflows the [50 characters](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). I'll gladly work on a function that rewrites the title to avoid overflows.

jd: What about just using the original title?
The target branch is different anyway, so that might not be a problem.

You could also use a template system in the action where `{{title}}` would be replaced by the original title. So everyone can use their own backport title.

renatocaval: Great @ignasi35! 

@jd I like the idea of templating. 

I also find it useful to see somewhere that this is a backport of some other PR, but that can be solved by a template. Each project can choose how they want the BP message.

ignasi35: I completely agree that templating would be the long term goal but I think it's a bit out of my reach ATM. I'd like to focus on the baby step first.

I've reviewed the behavior of `title="Automatic backport of pull request #%d" % pull.g_pull.number` by observing an actual [example](https://github.com/lagom/lagom/pull/1891) and noticed `pull.g_pull.number` is the number of the original PR. I am assuming, then, that `pull.g_pull.title` is the title of the original PR. If that's the case, `pull.g_pull.title` will _not_ contain a `(#num)` at the end, so my comment above is wrong.

Taking [this PR](https://github.com/lagom/lagom/pull/1890) as an example:

 * the original PR number is `#1890`
 * the original PR title is `Upgrades to Play 2.7.2`

My intent is for mergify to produce the backport PR with the title `Upgrades to Play 2.7.2 (bp #1890)`. That is `"{} (bp #{})".format(pull.g_pull.title, pull.g_pull.number)` (using [`.format()` notation](https://pyformat.info/)).

@jd It'd be great if you could provide some guidance for me to write and run some tests locally. I'm running a bit blind. Again, first-timer on python-world 😅.

jd: You can run the test locally by using `tox -e py37`, assuming you have Python 3.7 installed.

Since your change did not break any test, I don't think we have a test checking the title of backported PR. That sounds not a big issue for me, and adding more tests is complicated  — the tests have to interact with `github.com` which needs special tests accounts).

So I'm good with that for now. I'll @sileht comment.

sileht:
A template would be better, but yeah, this new title looks better anyways.

ignasi35:
Hi @sileht, does that mean we can remove the `work-in-progress` label?

PR #396, opened 2019-05-09
merged_by can be none event if PR is merged

sileht:
That doesn't look logic, but well we have received some events with
this...

Closes: MERGIFY-ENGINE-PP, MERGIFY-ENGINE-PQ


sileht:
Pull-request updated, HEAD is now e0852bfd531b2379d81f8e7a74a8ea0a79cdd263

jd:
Adding a test would have been easy, just saying… ;)

PR #408, opened 2019-05-16
engine: remove deprecated summary check name support

jd:


sileht:
Déjà ?

jd:
It has been >2 weeks, sounds like most PR that are "active" might be merged by now :)

PR #412, opened 2019-05-19
branch_update: fix some git issues

sileht:
branch_update: really unshallow git clone
branch_update: Catch some new git issues


jd:
Typo brancg -> branch

PR #414, opened 2019-05-20
promexporter: don't stop polling on GitHub server errors

jd:


sileht:
I have hesitated to do this, because reported counters will not be exact. But the loop take so much time now, that It's better to have number that nothing :)

PR #418, opened 2019-05-30
branch_update: use new github API in case of method=merge

sileht:
Many tests have been regenerated because Github does not always sent
our event when we update only the "Summary text".


mergify_engine/actions/dismiss_reviews.py
```diff
@@ -37,12 +38,14 @@ class DismissReviewsAction(actions.Action):
     def run(self, installation_id, installation_token,
             event_type, data, pull, missing_conditions):
         if (event_type == "pull_request" and data["action"] == "synchronize"):
+            # NOTE(sileht): mergify-bot have push the "Update Branch" button.
```
jd: pushed

PR #420, opened 2019-05-31
prom_exporter: ignore 401 status code

jd:
GitHub will return 401 if Mergify does not have access anymore. Just ignore it.


sileht:
401 also when the access token expire too, but anyways it's better with this.

PR #422, opened 2019-06-01
prom: Use threadpool to retrieve data in parallel

sileht:


mergify_engine/prom_exporter.py
```diff
@@ -134,6 +137,11 @@ def collect_metrics():
             if e.status not in (403, 401) and e.status < 500:
                 raise
 
+    with futures.ThreadPoolExecutor() as executor:
```
jd: I think the default is the number of CPU, though in this case it's unlikely to be a good default.

sileht: Number of cpu x5

PR #443, opened 2019-06-27
Handle 422 when we update pull request branch

sileht:


mergify_engine/branch_updater.py
```diff
@@ -168,7 +168,12 @@ def update_with_api(pull):
             input={"expected_head_sha": pull.g_pull.head.sha},
             headers={"Accept": "application/vnd.github.lydian-preview+json"})
     except github.GithubException as e:
-        if e.status < 500:
+        if e.status == 422:
+            LOG.debug("update branch done somewhere in meantime",
```
jd: I'd go for "branch already updated" or "already up-to-date"

sileht: Well maybe not, it can also be some-else that push more commits in the meantime

jd: Ah ok I see what you mean. So maybe "branch updated in the meantime" would be clearer.

sileht: done

Issue #450, opened 2019-07-15
Assign action


edahlseng:
Would it be possible to add an action for adding an assignee to PRs?

PR #452, opened 2019-07-16
actions: Add assign action

edahlseng:
This pull request adds an `assign` action. I have no clue how to test this out, so I suspect that there are issues, but hopefully this gets the ball rolling. This should address #450.


jd:
Thanks @edahlseng!
LGTM. I'll let @sileht approve. He might also want to add some test himself. :)

sileht:
I have added tests here: https://github.com/Mergifyio/mergify-engine/pull/455

PR #459, opened 2019-07-23
actions: Add request_reviews action

edahlseng:
This pull request adds a `request_reviews` action. I still don't quite understand how to test this out, so there may very well be some issues.


jd:
The tests need credentials to run and register requests done against GitHub itself. So you could write the tests, but unless you setup the whole credentials list we use for recording the tests, you can't really run them. :( It sucks but that's also a good guarantee we're not testing against something fake. GitHub changes its API quite often without warning so testing against real endpoint is crucial.

I'll let @sileht comment and add tests as he did before. :)

edahlseng:
Ah, that makes sense @jd, thanks for the info!

And thanks for adding tests and pushing this through @sileht! 🙂

PR #465, opened 2019-08-09
Add simulator

sileht:


mergify_engine/web.py
```diff
@@ -151,6 +160,67 @@ def subscription_cache(installation_id):  # pragma: no cover
     return "Cache cleaned", 200
 
 
+SimulatorSchema = voluptuous.Schema({
+    voluptuous.Required('url'): voluptuous.Url(),
```
jd: `pull_request_url`?

mergify_engine/web.py
```diff
@@ -151,6 +160,67 @@ def subscription_cache(installation_id):  # pragma: no cover
...
+    voluptuous.Required('mergify.yml'): str,
+})
+
+
+def abort(message, code=400):
```
jd: You could rather use something like:

```python
@app.errorhandler(exceptions.MergifyException)
def handle_mergify_exception(error):
    response = flask.jsonify({
        "error": error.__class__.__name__,
        "message": error.message,
        "detail": error.detail,
    })
    response.status_code = error.status_code
    return response
```

And define/raise proper exceptions.

mergify_engine/web.py
```diff
@@ -151,6 +160,67 @@ def subscription_cache(installation_id):  # pragma: no cover
...
+        abort("Mergify not installed")
+
+    token = integration.get_access_token(installation_id).token
+    pull = mergify_pull.MergifyPull.from_number(
+        installation_id, token, owner, repo, pull_number
+    )
+
+    try:
+        mergify_config = rules.validate_user_config(data["mergify.yml"]) or {}
+    except rules.InvalidRules as e:  # pragma: no cover
```
jd: sounds like we could cover this

mergify_engine/web.py
```diff
@@ -151,6 +160,67 @@ def subscription_cache(installation_id):  # pragma: no cover
...
+        message = str(e)
+        if hasattr(e, 'problem_mark'):
+            message += " (position %s:%s)" % (e.problem_mark.line + 1,
```
jd: Rather than `abort()`ing with that message, I'm pretty sure that exposing those position as a proper exception field that would be serialized (as suggested earlier) would ease error reporting on the UI side.

mergify_engine/web.py
```diff
@@ -151,6 +160,67 @@ def subscription_cache(installation_id):  # pragma: no cover
...
+    token = integration.get_access_token(installation_id).token
+    pull = mergify_pull.MergifyPull.from_number(
+        installation_id, token, owner, repo, pull_number
+    )
+
+    try:
+        mergify_config = rules.validate_user_config(data["mergify.yml"]) or {}
+    except rules.InvalidRules as e:  # pragma: no cover
+        message = str(e)
+        if hasattr(e, 'problem_mark'):
```
jd: I don't think `InvalidRules` has `problem_mark`. It sounds to me you pasted code from `validate_user_config` but that does not match here.

mergify_engine/web.py
```diff
@@ -151,6 +160,67 @@ def subscription_cache(installation_id):  # pragma: no cover
...
+            message += " (position %s:%s)" % (e.problem_mark.line + 1,
+                                              e.problem_mark.column + 1)
+        abort(message)
+
+    if not mergify_config.get("pull_request_rules"):
```
jd: Should this be part of `validate_user_config`?

mergify_engine/tests/functional/test_simulator.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class TestEngineV2Scenario(base.FunctionalTestBase):
+    """Mergify engine tests.
+
+    Tests user github resource and are slow, so we must reduce the number
+    of scenario as much as possible for now.
+    """
+
+    def test_simulator(self):
```
jd: That definitely needs some more failure scenarios.

mergify_engine/web.py
sileht: I put just pull_request

mergify_engine/web.py
sileht: This give me the idea to rework a bit the rule validator to remove most of the manual 400 reporting. I just put a handler for mashmallow.Invalid.

mergify_engine/web.py
sileht: done

mergify_engine/web.py
sileht: I created a new voluptuous object to handle this cleanly.

mergify_engine/web.py
sileht: Fixed

mergify_engine/web.py
sileht: I moved this to the voluptuous schema.

mergify_engine/tests/functional/test_simulator.py
sileht: I added a some negative tests here and in test_config.

jd:
@sileht test failed :(

mergify_engine/web.py
```diff
@@ -151,6 +157,70 @@ def subscription_cache(installation_id):  # pragma: no cover
...
+    voluptuous.Required('mergify.yml'): rules.UserConfigurationSchema,
+})
+
+
+@app.errorhandler(voluptuous.Invalid)
+def voluptuous_error(error):
+    payload = flask.jsonify({
+        "message": str(error),
+        "type": error.__class__.__name__,
+        "error": error.msg,
```
jd: Is this really needed? sounds like `message` is enough

sileht: That's not the same information: https://github.com/alecthomas/voluptuous/blob/master/voluptuous/error.py#L27-L36

PR #467, opened 2019-08-12
Prevent requesting review from users who have already reviewed

edahlseng:
This pull request filters out users who have already submitted a review from the `request_reviews` action.

It looks like the GitHub API happily accepts new review requests for users who have already review. Since Mergify runs whenever a PR is edited, this means that with the recently added `request_reviews` action, if a user reviews a PR, then Mergify will request another review from them within moments. This pull request should prevent that from happening. I don't think there's much more that can be done without more extensive configuration.

### DIFF
--- a/mergify_engine/actions/request_reviews.py
+++ b/mergify_engine/actions/request_reviews.py
@@ -1,3 +1,5 @@
+import itertools
+
 import voluptuous
 
 from mergify_engine import actions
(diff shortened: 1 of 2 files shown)


edahlseng:
Thanks @sileht, it looks like I was quite far from the mark with my attempt 🙈.

Is there anything I can do to help push this through?

sileht:
The logic is the same, I just removed an useless API call.

I thought it was already merged.

edahlseng:
Thanks @sileht. It looks like the status check in the Mergify configuration doesn't match the status checks on this PR, so it's not being merged as expected...

jd:
That's likely because I disable the CircleCI status check extension. @sileht should we put it back or should I modify the `.mergify.yml`?

PR #470, opened 2019-09-06
Fix typo on job marketplace events handler

jd:


jd:
This probably needs testing so it does not happen again.

PR #472, opened 2019-09-06
tests: remove unused fake-sub-api

jd:


sileht:
It's used here: https://github.com/Mergifyio/mergify-engine-docker/blob/master/docker-compose.fake.yaml#L21

jd:
ok!

PR #474, opened 2019-09-06
rules: allow to store configuration in .mergify/config.yml

jd:


mergify_engine/debug.py
```diff
@@ -122,13 +122,17 @@ def report(url):
 
     print("* CONFIGURATION:")
     try:
-        mergify_config = rules.get_mergify_config(r)
+        mergify_config_content = rules.get_mergify_config_content(r)
     except rules.NoRules:  # pragma: no cover
         print(".mergify.yml is missing")
+
+    print(mergify_config_content)
```
sileht: meuh!

PR #475, opened 2019-09-06
handle neutral status

### LINKED ISSUE #468: Provide status-neutral-success
Currently status-success means it has to be succeed - green check mark.

However, sometimes if no code change, some services will skip it.

For example, in this PR

https://github.com/Hongbo-Miao/hongbomiao.com/pull/171

image

The rule

```
status-success=LGTM analysis: JavaScript
```

won't let Mergify merge.

Can we provide something like this?

```
status-neutral-success=LGTM analysis: JavaScript
```

so as long as it is not failed, it will still let Mergify merge? Thanks

sileht:
Closes #468


Hongbo-Miao:
Thanks @sileht !

PR #480, opened 2019-09-09
doc: Update about protected branches

sileht:

PR #481, opened 2019-09-09
doc: remove FAQ entry about branch protection settings

jd:
This is now fixed:

https://developer.github.com/changes/2019-09-05-apps-protected-branches-api/


jd:
#480 is better.

PR #484, opened 2019-09-09
rules: fix regex with value is none

sileht:


mergify_engine/rules/filter.py
```diff
@@ -83,7 +83,7 @@ class Filter:
         "!=": (operator.ne, all, _identity),
         "≠": (operator.ne, all, _identity),
 
-        "~=": (lambda a, b: b.search(a), any, re.compile),
+        "~=": (lambda a, b: a and b.search(a) or False, any, re.compile),
```
jd: So you mean `a and b.search(a)`? ;)

sileht: lol

PR #501, opened 2019-09-22
Add status neutral or success

thislooksfun:
Some checks (like LGTM analysis) return a status of Neutral to mean unchanged and Success to mean changed, but ok. Either of these statuses is fine, and should be allowed to pass. Currently (since afaik there is no way to manually "or" rules), the only way is to create two fully separate rules, each of which differs only in that one specifies `status-success` and the other specifies `status-neutral`, which is very redundant and bad for maintainability. This PR adds a new check, `status-neutral-or-success`, which, as it suggests, matches when the status is either `neutral` or `success`.

Alternatives considered:

1. Implement a formal way to "or" conditions together in a single rule -- doable, but could easily raise the parsing complexity a lot
2. Implement a way to have templated rules (see #502), and use that as an "or" instead -- also a good idea for other reasons
3. Just have multiple (slightly different) rules -- works, but is very hard to maintain

Fully implements #468

### DIFF
--- a/doc/source/conditions.rst
+++ b/doc/source/conditions.rst
@@ -155,13 +155,13 @@ Here's the list of pull request attribute that rules can be matched against:
        `continuous-integration/travis-ci/pr` or of a *check run* such as
        `Travis CI - Pull Request`. See `About status check name`_ for more
        details.
-  * - ``status-neutral-or-success``
-    - array of string
-    - The list of status checks that either successfully passed or are neutral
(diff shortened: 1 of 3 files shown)


jd:
I'm afraid that if we go this route it might never end of `or` status.

We never implemented an `or` in the syntax to keep things simple and straightforward. Maybe it's time to think about it?

jd:
Closing this as we'd rather go the `or` operator support rather than this. Thanks for the contribution!

PR #506, opened 2019-10-02
Add copy action

sileht:


doc/source/actions.rst
```diff
@@ -73,6 +73,33 @@ parameter:
 Once the backporting pull request is closed or merged, Mergify will
 automatically delete the backport head branch that it created.
 
+copy
+====
+
+The `copy` action just create a copy of the pull request as soon as the
```
jd: > The `copy` action creates a copy of the pull request targetting other branches.

No need to say more I think all actions are triggered when conditions match :)

doc/source/actions.rst
```diff
@@ -73,6 +73,33 @@ parameter:
...
+     - array of string
+     - ``[]``
+     - The list of branches the pull request should be copied to.
+   * - ``regexes``
+     - array of string
+     - ``[]``
+     - The list of regexes to find branches the pull request should be copied to.
+
+Once the copy pull request is closed or merged, Mergify will
+automatically delete the copy head branch that it created.
```
jd: Shouldn't that be done by GitHub itself (they now have an option for it) or via `delete_head_branch`?

doc/source/actions.rst
```diff
@@ -73,6 +73,33 @@ parameter:
...
+   * - ``branches``
+     - array of string
+     - ``[]``
+     - The list of branches the pull request should be copied to.
+   * - ``regexes``
+     - array of string
+     - ``[]``
+     - The list of regexes to find branches the pull request should be copied to.
+
+Once the copy pull request is closed or merged, Mergify will
```
jd: I'd say `new pull request` rather than `copy pull request`. I find it confusing since I'm not sure you're talking about the one that is the source of the new one.

> Once the new pull requests are closed or merged, Mergify will...

mergify_engine/actions/copy.py
```diff
@@ -0,0 +1,139 @@
...
+    try:
+        re.compile(value)
+    except re.error as e:
+        raise voluptuous.Invalid(str(e))
+    return value
+
+
+class CopyAction(actions.Action):
+    KIND = "copy"
+    SUCCESS_MESSAGE = "Backports have been created"
```
jd: Backports?

doc/source/actions.rst
sileht: I updated the default mergify rules.

mergify_engine/actions/copy.py
```diff
@@ -0,0 +1,139 @@
...
+    try:
+        re.compile(value)
+    except re.error as e:
+        raise voluptuous.Invalid(str(e))
+    return value
+
+
+class CopyAction(actions.Action):
+    KIND = "copy"
+    SUCCESS_MESSAGE = "Pull request copy have been created"
```
jd: "has been"

The message for backports is pural, I guess it should be here too:

`Pull requests copies have been created`

doc/source/actions.rst
jd: My point was more about not deleting it by default.
You can decide to delete it automatically with a rules, but you can't deactivate Mergify default rules.

PR #519, opened 2019-10-18
Fix check API summary limit

jd:
check_api: truncate output summary if bigger than 65535 bytes
tests: retry fetching from forks


sileht:
I'm currious about how many checks the repo have to reach that.

PR #529, opened 2019-11-07
request_review: add support for teams

jd:


mergify_engine/tests/functional/test_engine_v2.py
```diff
@@ -286,16 +285,17 @@ def test_request_reviews_users(self):
             sorted(["mergify-test1"]), sorted([l.login for l in requests[0]])
         )
 
-    @unittest.skip("Github API doesn't behave as the doc say")
     def test_request_reviews_teams(self):
+        # Add a team to the repo with write permissions  so it can review
+        team = list(self.o_admin.get_teams())[0]
+        team.set_repo_permission(self.r_o_admin, "push")
```
sileht: I'm curious to see why we need that.

jd: This is how it works. You can't add a team as a reviewer if it's not part of the repo.

PR #543, opened 2019-11-25
Add basic documentation in created PR

sileht:


mergify_engine/duplicate_pull.py
```diff
@@ -29,6 +29,29 @@
...
+<br />
+You can also trigger Mergify actions by commenting on pull requests:
+
+- `@Mergifyio rebase` will rebase this PR
+- `@Mergifyio backports <destination>` will backport this PR on `<destination>` branch
+
+Additionally, on Mergify [dashboard](https://dashboard.mergify.io/) you can:
+
+- look at your merge queues
+- generate the `.mergify.yml` with the simulator.
```
jd: maybe use `configuration file` rather than `.mergify.yml` since it can also be named `.mergify/config.yml`?

mergify_engine/duplicate_pull.py
```diff
@@ -29,6 +29,29 @@
...
+<details>
+<summary>Mergify commands and options</summary>
+<br />
+More conditions and actions can be found in the [documention](https://doc.mergify.io/).
+<br />
+<br />
+You can also trigger Mergify actions by commenting on pull requests:
+
+- `@Mergifyio rebase` will rebase this PR
+- `@Mergifyio backports <destination>` will backport this PR on `<destination>` branch
```
jd: without `s` to `backport`

PR #548, opened 2019-12-02
Add statsd support

sileht:


mergify_engine/tasks/engine/__init__.py
```diff
@@ -91,6 +92,7 @@ def create_metrics(event_type, data):
     # that will be recorded by celery and exported with celery exporter
     if event_type == "pull_request" and data["action"] == "opened":
         pull_request_opened.apply_async()
+        statsd.increment("engine.pull-request-counter.opened")
```
jd: you can remove the `-counter` suffix

mergify_engine/tasks/engine/__init__.py
```diff
@@ -99,10 +101,12 @@ def create_metrics(event_type, data):
     ):
 
         pull_request_merged.apply_async()
+        statsd.increment("engine.pull-request-counter.merged")
```
jd: ditto

mergify_engine/tasks/engine/__init__.py
```diff
@@ -99,10 +101,12 @@ def create_metrics(event_type, data):
...
         if data["pull_request"]["merged_by"] and data["pull_request"]["merged_by"][
             "login"
         ] in ["mergify[bot]", "mergify-test[bot]"]:
             pull_request_merged_by_mergify.apply_async()
+            statsd.increment("engine.pull-request-counter.merged-by-mergify")
```
jd: ditto

PR #549, opened 2019-12-02
statsd: send runtime

sileht:


mergify_engine/worker.py
```diff
@@ -81,6 +81,9 @@ def statsd_after_task_publish(sender, **kwargs):
 @celery.signals.task_postrun.connect
 def statsd_task_postrun(sender, **kwargs):
     statsd.increment("engine.task_run", tags=[f"task_name:{sender.name}"])
+    statsd.gauge(
```
jd: This is not a gauge as it's the total task of total time you're going to have. You should use `increment` here.

PR #569, opened 2019-12-11
datadog: generate process conf based on DYNOTYPE

jd:


datadog/conf.d/process.yaml
```diff
@@ -1,20 +1,4 @@
...
-  - name: celery-beat
-    # celery-beat
-    search_string: ['[celery beat]']
-  - name: gunicorn-worker
-    search_string: ['^gunicorn: worker']
-    exact_match: false
-  - name: gunicorn-master
-    search_string: ['^gunicorn: master']
-    exact_match: false
+  # Filled by prerun.sh
```
sileht: The path in prerun is not the same.

jd: That's normal, this is how it works.

PR #592, opened 2019-12-13
metering of actions and commands

sileht:


mergify_engine/tasks/engine/actions_runner.py
```diff
@@ -295,6 +296,7 @@ def run_actions(
                     missing_conditions,
                 )
                 if method_name == "run":
+                    statsd.increment("engine.actions.count", tags=["name:%s" % action])
```
jd: More fstrings plz

PR #596, opened 2019-12-14
Pull request for action-tracker

sileht:
## Improve the action tracker

The previous action tracker was relying on the dedicated action report
posted via Checks API.

The issue is that some actions don't post any report (eg: comment,
request_review, assignee). These actions have special code to avoid
reposting the same comment on each event.

The new action tracker will not care about the report and will
write the action conclusions into the Mergify summary. So next
rules evaluation can retrieve it and compare it with the current
expected state to known if we have to run the action again.

## tests: redesign the event tracking

This makes `tox -e record` a bit more solid


mergify_engine/actions/merge/helpers.py
```diff
@@ -72,6 +72,13 @@ def merge_report(pull, strict):
     return conclusion, title, summary
 
 
+WAIT_FOR_CI_REPORT = (
```
jd: I'd rather name it `WAIT_FOR_STATUS_CHECK_REPORT` or the like, but fine

PR #601, opened 2019-12-16
remove unused things

sileht:
* github deployment hook
* docker image


jd:
Missing mergify conf :)

PR #613, opened 2019-12-19
logs: normalize extra= for anything related to a pull request

sileht:


mergify_engine/utils.py
```diff
@@ -113,6 +125,9 @@ def setup_logging():
             ("vcr", "WARN"),
         ]
     )
+
+    github.PullRequest.PullRequest.log = property(GithubPullRequestLog)
```
jd: Why not using a standard class declaration with the decorator?

sileht: I'm a bit lazy on that one. Sometimes we use the MergifyPull, someother directly github.PullRequest(). 

I will try to remove all github.PullRequest() usage, but I'm not sure we can remove them all.

jd: Oh sorry, I read that as being `MergifyPull` and not `github.PullRequest`. As you wish. I think it'd be better to clean the code indeed, especially if we plan on removing `github.PullRequest` long term.

PR #640, opened 2020-01-04
simu: allow to just valid the configuration

sileht:


mergify_engine/web.py
```diff
@@ -216,18 +216,23 @@ def simulator():
...
+            "repository": pull_request.g_pull.base.repo.raw_data,
+            "installation": {"id": pull_request.installation_id},
+            "pull_request": pull_request.g_pull.raw_data,
+        }
+        title, summary = actions_runner.gen_summary(
+            "refresh", raw_event, pull_request, match
+        )
+    else:
+        title = "The configuration is valid"
+        summary = "Provide a pull request to get the summary."
```
jd: I'd rather return a simple `None` and leave the summary be set by the UI?

PR #643, opened 2020-01-07
Use gevent by default

sileht:


tox.ini
```diff
@@ -52,6 +52,7 @@ commands =
   pip uninstall --yes mergify-engine
   bash -c "pip freeze --exclude-editable >| requirements.txt"
   bash -c "echo '-e .' >> requirements.txt"
+  echo "Don't forget to readd extra, pipe freeze remove them."
```
jd:
```suggestion
  echo "Don't forget to re-add the extra flavors in the requirement list, pip freeze removes them."
```

jd:
I guess we forgot to regenerate the requirements for a while :D

PR #646, opened 2020-01-07
Use gevent by default

sileht:
## refactor: rework processes setups

The new processes setup is done by honcho.

celery beat is separated from celery worker.

tox -e test now deploy the same thing as production

## feat: gevent requirements setup

This change just prepare requirements and new processes setup for
gevent. But does not enable gevent yet.

Since it adds a lot of requirements, and heroku is take a lot of times
to setup them.

If we need to revert the real gevent change, requirements will not
change. And the revert will be fastest.

## feat: gevent supports only py3.7 for now.

to reuse python 3.8.1 we must wait 1.5


setup.cfg
```diff
@@ -26,7 +26,7 @@ install_requires =
     python-dotenv
```
jd: You might want to readd 3.7 to the classifiers too in this file.

PR #655, opened 2020-01-15
Log rotate every minute instead of 5 seconds

sileht:


mergify_engine/config.py
```diff
@@ -42,6 +42,7 @@ def CommaSeparatedStringList(value):
         voluptuous.Required("LOG_RATELIMIT", default=False): CoercedBool,
         voluptuous.Required("LOG_STDOUT", default=True): CoercedBool,
         voluptuous.Required("LOG_JSON_FILE", default=None): voluptuous.Any(None, str),
+        voluptuous.Required("LOG_JSON_TCP", default=None): voluptuous.Any(None, str),
```
jd:
```suggestion
```

PR #657, opened 2020-01-16
Add support to datadog logger

sileht:


sileht:
Need release of https://github.com/jd/daiquiri/pull/41 first.

PR #659, opened 2020-01-17
Pull request for datadog-tcp

sileht:
## feat(logging): drop queues contents

This would help for debugging

## feat(logging): allow to config level of STDOUT

It's interesting, to keep error and critical in the console in
additional to external logger that receive everything.

### DIFF
--- a/mergify_engine/config.py
+++ b/mergify_engine/config.py
@@ -16,6 +16,7 @@
 
 import base64
 import distutils.util
+import logging
 import os
 import re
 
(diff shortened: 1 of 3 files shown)


mergify_engine/config.py
```diff
@@ -31,6 +32,13 @@ def CoercedBool(value):
     return bool(distutils.util.strtobool(str(value)))
 
 
+def CoercedLoggingLevel(value):
+    value = value.upper()
+    if value in ("CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"):
+        return getattr(logging, value)
+    raise ValueError
```
jd:
```suggestion
    raise ValueError(value)
```

PR #667, opened 2020-01-17
docs(examples): revamp existing examples and add some more

jd:
Also increase the menu width to be more readable for the examples section.


doc/source/examples.rst
```diff
@@ -4,26 +4,28 @@
...
 ===============
 
-You can define more specific rules based on the large number of criterias
-available: pull request author, base branch, labels, files, etc.
+Mergify allows you to define a lot of specific rules. There is a large number
+of criterias available to define rues: pull request author, base branch,
+labels, files, etc.
 
-Here's a few example that should help you getting started.
+In this section, we build a few eamples that should help you getting started
```
sileht: eamples -> examples

doc/source/examples.rst
```diff
@@ -4,26 +4,28 @@
...
+and cover many common uses cases.
```
sileht: `uses`,  not `use` ?

PR #683, opened 2020-01-27
fix(commands): warns about wrong account with correct case

### LINKED ISSUE #685: Erroneous logging of wrong name in PR messages
When we referred to @Mergifyio here: https://github.com/flux-framework/flux-core/pull/2678#issuecomment-579024146, it replied with:

```
Command rebase: success

    Branch has been successfully rebased

Hey, I reacted but my real name is @Mergifyio
```

Which is confusing since we used the exact string it printed out to refer to it.  This also happened in this PR (that I found via a DDG search): https://github.com/openupm/openupm/pull/16#issuecomment-570160689

Poking at the mergify-engine code, it looks like [this conditional](https://github.com/Mergifyio/mergify-engine/blob/bf46764e5c454a55b0eaf0bb593686162a4520f7/mergify_engine/tasks/engine/commands_runner.py#L119) checks for a lower case reference of `@mergifyio` but [this error message string](https://github.com/Mergifyio/mergify-engine/blob/bf46764e5c454a55b0eaf0bb593686162a4520f7/mergify_engine/tasks/engine/commands_runner.py#L31) contains the capitalized version `@Mergifyio`.  I suggest at minimum changing the error message to use the lowercase form, but I think a more general solution would be to check for `@mergifyio` in a lower-cased version of the commit message (i.e., `if "@mergifyio" not in comment.lower():`.

jd:
Fixes #685

### DIFF
--- a/mergify_engine/tasks/engine/commands_runner.py
+++ b/mergify_engine/tasks/engine/commands_runner.py
@@ -61,7 +61,7 @@ def spawn_pending_commands_tasks(installation_id, event_type, data, g_pull):
 
     for pending in pendings:
         run_command.s(
-            installation_id, event_type, data, "@mergifyio %s" % pending, rerun=True
+            installation_id, event_type, data, "@Mergifyio %s" % pending, rerun=True
         ).apply_async()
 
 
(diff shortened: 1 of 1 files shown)


jd:
@Mergifyio rebase

PR #684, opened 2020-01-27
fix(commands): only report command name without args

jd:
This fixes the statsd tagging which included arguments rather than just the
command name.

### DIFF
--- a/mergify_engine/tasks/engine/commands_runner.py
+++ b/mergify_engine/tasks/engine/commands_runner.py
@@ -32,14 +32,17 @@ WRONG_ACCOUNT_MESSAGE = "_Hey, I reacted but my real name is @Mergifyio_"
 
 
 def load_action(message):
+    """Load an action from a message.
+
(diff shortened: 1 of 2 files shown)


mergify_engine/tasks/engine/commands_runner.py
```diff
@@ -32,14 +32,16 @@
...
+    :return: A tuple with 2 values: the command name and the action."""
     action_classes = actions.get_commands()
     match = COMMAND_MATCHER.search(message)
     if match and match[1] in action_classes:
         action_class = action_classes[match[1]]
         config = action_class.command_to_config(match[2].strip())
         action = voluptuous.Schema(action_class.get_schema())(config)
-        command = "%s%s" % (match[1], match[2])
-        return command, action
+        return match[1], action
```
sileht: We must return the args here, there are used in user reported messages

jd: Ok, I wasn't sure it was really useful to report the arguments too, but you seem opiniated. :)

PR #715, opened 2020-02-14
fix: render correctly markdown content inside HTML entity

calvellido:
As it can be seen in

https://github.com/higherkindness/sbt-compendium/pull/53

the comment set by Mergify fails to render the ~documention~ documentation link correctly. This is due to the need of additional line breaks on HTML entities, to be properly detected through the markdown parser. Here it can be checked a manually fixed comment:

https://github.com/higherkindness/mu-scala/pull/790

---

This PR aims to fix this by modifying the default initial comment content.

* Fix typo.
* Modify markdown content to be rendered properly in GitHub.
* Update tests.


jd:
Thanks a lot for your contribution! This is definitely valuable.
The tests might not be easy to regenerate, we'll take care of it.

calvellido:
Great!  Let me know how would be best to proceed, should I revert all test changes?

Thanks!

jd:
I'm working on regenerating the test results @calvellido, don't worry too much :)

jd:
@calvellido you can double check I did not break your original patch, but that should be good!

calvellido:
@jd all good!  :raised_hands:

jd:
@mergifyio refresh

PR #718, opened 2020-02-18
refactor: prepare event batching

sileht:
## refactor: prepare event batching (actions)

This create a new object called "sources" that will
contains all raw events data assiociated to a pull request.

Actually the sources will still only contains one event. But once we
will be able to batch github events, we can have more events one pull
requests.

This also makes actions interfaces lighter, as we don't have to pass
everything to all actions.

installation_id/installation_token have been removed from actions too,
as their are now always accessible from MerlgifyPull object.

To be able to do that engine.action_runner.handle is no longer a celery
task.


jd:
typos in commit msg:
`assiociated` -> `associated`
`contains` -> `contain`
`MerlgifyPull` -> `MergifyPull`

PR #723, opened 2020-02-19
feat(actions): do not log on error but store failed state in the PR

jd:
When a post comment fails, rather than logging an error on our side, let's
store the failed state in the PR conclusion. That should allow the engine to
retry since the status will be "failure" rather than "sucesss".

We don't post the report to the check API because we don't want to have a
report for some actions like comment. To do that a new flag for Action is
added: "silent_report".

Fixes MERGIFY-ENGINE-ZB
FIxes MERGIFY-ENGINE-12Y

### DIFF
--- a/mergify_engine/actions/__init__.py
+++ b/mergify_engine/actions/__init__.py
@@ -61,6 +61,10 @@ class Action(abc.ABC):
     # If an action can't be twice in a rule this must be set to true
     only_once = False
 
+    # If set to True, does not post the report to the Check API
+    # Only keep it for internal tracking
+    silent_report = False
+
     @property
     @staticmethod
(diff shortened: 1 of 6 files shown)


jd:
@sileht I'm not sure if that'd work. If yes, it'd need to be done to other actions (I only did comment for now just to see if you think this was a valid change).

sileht:
I guess you find the dup comment bug ! well done.

To have mergify retrying on next event, you have to set the state to `None` instead of `Failure`.

And yes `silent_report` must be used in all other actions that doesn't return state but that can be done aside.

jd:
Ok, I've made a few changes there and there but I'd be reassured if you could double-check this is good for all actions.

PR #724, opened 2020-02-19
Pull request for no-g_pull

sileht:
## refactor: start removing usage of g_pull

This is a first step for removing pygithub.

MergifyPull will act as proxy to pygithub.PullRequest attributes

Then we will have almost all pygithub call located here.


mergify_engine/mergify_pull.py
```diff
@@ -342,3 +337,85 @@ def __str__(self):
...
+    @property
+    def merged(self):
+        return self.g_pull.merged
+
+    @property
+    def merged_by(self):
+        return self.g_pull.merged_by
+
+    @property
+    def stat_commits(self):
```
jd: why `stat_commits`?
Is this the number of commits? Should be `num_commits` then?

mergify_engine/mergify_pull.py
```diff
@@ -342,3 +337,85 @@ def __str__(self):
...
+    @property
+    def base_repo_name(self):
+        return self.g_pull.base.repo.name
+
+    @property
+    def head_repo_name(self):
+        return self.g_pull.head.repo.name
+
+    @property
+    def base_repo_owner(self):
```
jd: `base_repo_owner_login`?

mergify_engine/mergify_pull.py
```diff
@@ -342,3 +337,85 @@ def __str__(self):
...
+        return self.g_pull.base.repo.owner.login
+
+    @property
+    def head_repo_owner(self):
```
jd: `head_repo_owner_login`?

mergify_engine/mergify_pull.py
```diff
@@ -342,3 +337,85 @@ def __str__(self):
...
+    @property
+    def user(self):
+        return self.g_pull.user.login
+
+    @property
+    def state(self):
+        return self.g_pull.state
+
+    @property
+    def come_from_same_repo(self):
```
jd: I'd rather do `from_fork` it's simpler as a name, just invert the result

mergify_engine/mergify_pull.py
sileht: This is not even needed.

PR #728, opened 2020-02-25
fix(actions/copy): report final status as failure/success based on all PR

jd:
The current check status that is reported is the one from the latest created
PR. This does not make sense; having "success" if all copies are done correctly
is more logical.

This also add more testing on the check content that is reported by the action.

This also fixes a report error where `self.KIND` and the `branch_name` were
inverted in the string.


mergify_engine/actions/copy.py
```diff
@@ -35,12 +35,52 @@ def Regex(value):
...
     }
 
+    def _copy(self, pull, branch_name):
+        """Copy the PR to a branch.
+
+        Returns a tuple of strings (state, reason).
+        """
+        try:
+            branch = pull.g_pull.base.repo.get_branch(parse.quote(branch_name, safe=""))
+        except github.GithubException as e:
```
sileht: Not directly related to the change, but status >= 500 should be reraised here.

mergify_engine/mergify_pull.py
```diff
@@ -415,3 +415,7 @@ def merged(self):
     @property
     def merged_by(self):
         return self.g_pull.merged_by
+
+    @property
```
sileht: This is not needed `get_existing_duplicate_pull` returns a `pyGithub.PullRequest`, not a `MergifyPull`

mergify_engine/actions/copy.py
jd: I imagine returning `None` as state is better since we can then retry, right?
Doing that.

mergify_engine/mergify_pull.py
jd: Removed!

PR #731, opened 2020-02-25
feat(smart): show queue in check summary

sileht:
Since we refresh all PRs on each push, we can print the queue here.


mergify_engine/actions/merge/helpers.py
```diff
@@ -68,11 +69,19 @@ def merge_report(pull, strict):
...
+    pulls = queue.get_pulls_from_queue(pull)
+    if pulls:
+        links = ", ".join((f"#{pull}" for pull in pulls))
+        queue_status = f"The following pull requests are queued: {links}"
+
+    return (
+        None,
+        "Base branch updates done",
+        "The pull request has been automatically updated to follow its base branch and "
+        f"will be merged soon.\n\n{queue_status}",
```
jd: Do we need the `\n\n` if the `queue_status` is empty?

sileht: It's to have 2 paragraphes

jd: Right my question was, do we need 2 paragraphs when the second one is empty? ;)

PR #736, opened 2020-02-27
chore: log an error if report=None

sileht:
## chore: log an error if report=None

Every should be clean now, all actions should return a report now.

## fix: github exception status_code -> status

Fixes: MERGIFY-ENGINE-192


sileht:
@mergifyio rebase

PR #740, opened 2020-02-27
feat(actions/copy): add ignore_conflicts option

jd:
This adds a new option ignore_conflicts which allows to create copy or
backports of pull requests even if git fails.

This was the default and is now false by default. User will have to specify
that they still want a PR created if it's impossible to do so automatically.


sileht:
Closed in favor of https://github.com/Mergifyio/mergify-engine/pull/741

PR #741, opened 2020-02-27
feat(actions/copy): add ignore_conflicts option

jd:
This adds a new option ignore_conflicts which allows to create copy or
backports of pull requests even if git fails.

This was the default and is now false by default. User will have to specify
that they still want a PR created if it's impossible to do so automatically.


mergify_engine/duplicate_pull.py
```diff
@@ -27,6 +28,11 @@ class DuplicateNeedRetry(Exception):
     pass
 
 
+@attr.s
```
sileht: In python >= 3.7, you can use builtin https://docs.python.org/3/library/dataclasses.html

sileht:
This may need a test, otherwise, it's good to go.

PR #745, opened 2020-02-28
feat: enable http cache

sileht:

### DIFF
--- a/mergify_engine/__init__.py
+++ b/mergify_engine/__init__.py
@@ -12,15 +12,22 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import re
+
+import cachecontrol
+from cachecontrol.caches import RedisCache
 import celery.exceptions
 import daiquiri
+import requests
 import sentry_sdk
 from sentry_sdk.integrations.celery import CeleryIntegration
 from sentry_sdk.integrations.flask import FlaskIntegration
 from sentry_sdk.integrations.redis import RedisIntegration
+import urllib3
 
 from mergify_engine import config
 from mergify_engine import exceptions
+from mergify_engine import utils
 
 
 LOG = daiquiri.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


mergify_engine/__init__.py
```diff
@@ -12,12 +12,16 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import cachecontrol
+from cachecontrol.caches import FileCache
```
jd: I feel like this is going to be lost on restart right?

setup.cfg
```diff
@@ -22,6 +22,8 @@ include_package_data = true
 
 install_requires =
     attrs
+    cachecontrol
+    lockfile
```
jd: `cachecontrol[filecache]` has the right deps

sileht: done

mergify_engine/__init__.py
sileht: It's was for my testing, it's going to be redis at this end, I have already prepared a new redis instance in production. We now have 5 servers, so we can't downgrade redis plan anymore.

sileht:
It could be merged automatically, but I would like to be around when it reaches production :)

mergify_engine/__init__.py
```diff
@@ -48,3 +55,64 @@ def fixup_sentry_reporting(event, hint):  # pragma: no cover
...
+        "DELETE",
+        "POST",
+        "PATCH",
+    ],
+    raise_on_status=False,
+)
+
+
+SINGLE_PULL_API_RE = re.compile(
+    f"^https://api.{config.GITHUB_DOMAIN}:443/repos/[^/]*/[^/]*/pulls/\\d+$"
```
jd:
```suggestion
    f"^https://api.{config.GITHUB_DOMAIN}:443/repos/[^/]+/[^/]+/pulls/\\d+$"
```

requirements.txt
```diff
@@ -17,7 +18,7 @@ Flask==1.1.1
 Flask-Cors==3.0.8
 gevent==1.4.0
 greenlet==0.4.15
-gunicorn[gevent,setproctitle]==20.0.4
+gunicorn==20.0.4
```
jd: looks wrong

sileht: The real requirements are in setup.cfg, requirements.txt can be generated by `tox -egenreq` and it drops extras just like dependabot does.

mergify_engine/__init__.py
sileht: done

mergify_engine/__init__.py
```diff
@@ -48,3 +55,64 @@ def fixup_sentry_reporting(event, hint):  # pragma: no cover
...
+        "DELETE",
+        "POST",
+        "PATCH",
+    ],
+    raise_on_status=False,
+)
+
+
+SINGLE_PULL_API_RE = re.compile(
+    f"^https://api.{config.GITHUB_DOMAIN}:443/repos/[^/]+/[^/]+/pulls/\\d+$"
```
jd: I'd be good to have a few unit tests around this one just in case…

mergify_engine/__init__.py
```diff
@@ -48,3 +55,64 @@ def fixup_sentry_reporting(event, hint):  # pragma: no cover
...
+        self.mount(config.WEBHOOK_APP_FORWARD_URL, dashboard_adapter)
+    if config.WEBHOOK_MARKETPLACE_FORWARD_URL:
+        self.mount(config.WEBHOOK_MARKETPLACE_FORWARD_URL, dashboard_adapter)
+
+    if config.HTTP_CACHE_URL:
+        github_adapter = CustomCacheControlAdapter(
+            max_retries=RETRY, cache=RedisCache(utils.get_redis_for_http_cache())
+        )
+    else:
+        github_adapter = requests.adapters.HTTPAdapter(max_retries=RETRY)
```
jd: this is the same than `dashboard_adapter`. Reuse it?

mergify_engine/__init__.py
```diff
@@ -48,3 +55,64 @@ def fixup_sentry_reporting(event, hint):  # pragma: no cover
...
+
+def cached_session_init(self, *args, **kwargs):
+    real_session_init(self, *args, **kwargs)
+
+    dashboard_adapter = requests.adapters.HTTPAdapter(max_retries=RETRY)
+    self.mount(config.SUBSCRIPTION_URL, dashboard_adapter)
+    if config.WEBHOOK_APP_FORWARD_URL:
+        self.mount(config.WEBHOOK_APP_FORWARD_URL, dashboard_adapter)
+    if config.WEBHOOK_MARKETPLACE_FORWARD_URL:
+        self.mount(config.WEBHOOK_MARKETPLACE_FORWARD_URL, dashboard_adapter)
```
jd: hm why do we need a not-cached adapter for this one?

mergify_engine/__init__.py
sileht: done

mergify_engine/__init__.py
sileht: Cached adapter works only if API server support `etag`, our dashboard api doesn'. So, it's useless.

jd: Is it just useless, as is it has no impact, or it does not work?
I'm just wondering if it's worth having this 4 lines.

jd:
Can we get stats about the number of hit cache/miss?

mergify_engine/__init__.py
sileht: I double checked cachecontrol, that should be safe, only 301 and 304 code have special handling, but client that doesn't send etag should never received them, so.

mergify_engine/__init__.py
sileht: done

mergify_engine/__init__.py
jd: > but client that doesn't send etag should never received them, so.

Ok, that makes sense then!

sileht:
> Can we get stats about the number of hit cache/miss?

I will take a look at it.

PR #753, opened 2020-03-02
feat(datadog): track http_client_cache miss/hit

sileht:


mergify_engine/__init__.py
```diff
@@ -93,7 +95,13 @@ def send(self, request, *args, **kwargs):
         if SINGLE_PULL_API_RE.match(request.url):
             cache_url = self.controller.cache_url(request.url)
             self.cache.delete(cache_url)
-        return super().send(request, *args, **kwargs)
+        resp = super().send(request, *args, **kwargs)
+        hostname = parse.urlsplit(request.url).hostname
+        if getattr(resp, "from_cache", False):
+            statsd.increment("engine.http_client_cache.hit", tags=[f"to:{hostname}"])
```
jd: I'd have used `hostname` as a tag, it's not like we'd have a `from` tag, right?

PR #794, opened 2020-03-14
feat: introduction to httpx

sileht:
## feat: replace static data of MergifyPull, with httpx client

This change replaces data of MergifyPull with data coming from the new
httpx client.

The tokens and data are passed to pygithub to do a smooth transition to
httpx.


mergify_engine/clients/github.py
```diff
@@ -0,0 +1,90 @@
...
+        self.installation_id = installation_id
+        self.access_token = None
+
+    def auth_flow(self, request):
+        request.headers["Authorization"] = f"token {self.get_access_token()}"
+        response = yield request
+        if response.status_code == 401:
+            self.access_token = None
+            request.headers["Authorization"] = f"token {self.get_access_token()}"
+            yield request
```
jd: I feel like that should be a while loop now?
why only 2 requests otherwise? The last one could also fail no?

mergify_engine/clients/github_app.py
```diff
@@ -0,0 +1,85 @@
...
+
+        return self.jwt
+
+    def auth_flow(self, request):
+        request.headers["Authorization"] = f"Bearer {self.get_or_create_jwt()}"
+        response = yield request
+        if response.status_code == 401:
+            self.jwt = None
+            request.headers["Authorization"] = f"Bearer {self.get_or_create_jwt()}"
+            yield request
```
jd: same as previous

mergify_engine/clients/github_app.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class _Client(httpx.Client, common.HttpxRetriesMixin):
+    def __init__(self):
+        super().__init__(
+            base_url=f"https://api.{config.GITHUB_DOMAIN}",
+            headers={"Accept": DEFAULT_ACCEPT, "User-Agent": "Mergify/Python"},
```
jd: maybe don't repeat this string for user agent?

mergify_engine/clients/github.py
sileht: Well, in theory if a freshly created token still return a 401, that means we not have the permissions anymore.

mergify_engine/clients/github_app.py
sileht: Done

mergify_engine/tasks/engine/__init__.py
```diff
@@ -65,24 +66,20 @@ def get_github_pull_from_event(repo, event_type, data):
...
-                pull = repo.get_pull(p["number"])
-            except github.UnknownObjectException:  # pragma: no cover
-                continue
-
-            if pull and not pull.merged:
-                return pull
+        # NOTE(sileht): data here is really incomplete most of the time,
+        # but mergify will complete it automatically if mergeable_state is missing
+        for pull in pulls:
+            return pull
```
jd: This looks suspicious… just `return pulls[0]`?

mergify_engine/tests/functional/test_github_client.py
```diff
@@ -0,0 +1,65 @@
...
+        }
+        self.setup_repo(yaml.dump(rules), test_branches=["other"])
+        self.create_pr()
+        self.create_pr()
+        self.create_pr(base="other")
+
+        client = github.get_client(self.o_integration.login, self.r_o_integration.name)
+
+        pulls = list(client.items("pulls"))
+        self.assertEqual(2, len(pulls))
```
jd: maybe switch to `assert`?
I mean at some point we might want to stop using `unittest` no?

mergify_engine/tasks/engine/__init__.py
sileht: In aA previous change, `pulls` was a iterator, but this is not needed anymore

mergify_engine/duplicate_pull.py
```diff
@@ -112,7 +112,6 @@ def _get_commits_to_cherrypick(pull, merge_commit):
                 return out_commits
 
     elif len(merge_commit.parents) == 2:
-        pull.log.info("Pull request merged with merge commit")
```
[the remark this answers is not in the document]
sileht: readded

mergify_engine/mergify_pull.py
```diff
@@ -275,19 +303,25 @@ def _ensure_mergable_state(self, force=False):
...
+            or self.data["mergeable_state"] not in self.UNUSABLE_STATES
         ):
             return
 
         raise exceptions.MergeableStateUnknown(self)
 
+    def update(self):
+        # TODO(sileht): Remove me,
+        # Don't use it, because consolidated data are not updated after that.
+        # Only used to
```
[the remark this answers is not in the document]
sileht: Completed

mergify_engine/tests/functional/test_github_client.py
sileht: This is even worse, we use testtools. I will switch the whole thing later.

Issue #360, opened 2019-04-02
Removing Stale Reviews and Labels


cipher1024:
From the documentation, I see that the following:

```yaml
pull_request_rules:
  - name: remove outdated reviews
    conditions:
      - base=master
    actions:
      dismiss_reviews:
```

can be used to dismiss reviews when someone pushes a commit to the PR. If I try to use the same logic to dismiss labels at the same time:

```yaml
  actions:
      dismiss_reviews:
        approved: True
      label:
        remove: ['ready-to-merge']
```

the label `ready-to-merge` gets deleted as soon as it is set.

jd:
Yeah, `dismiss_review` is triggered only on new commits. Label removal is triggered on any new events, including adding a label.

You can't trigger Mergify actions based on events, only on conditions (state). It just happens than dismiss_reviews has an event condition builtin.

That's something we could maybe enhance by adding a new fields "events" which would allow to filter conditions based on events received. Food for thought, @sileht?

cipher1024:
That sounds like a nice improvement. And it would also make clearer when `dismiss_review` occurs

0xc0170:
We are looking at something similar. See the mention above. We would like to dismiss reviews and do some other actions (remove labels, add labels).

jd:
IIUC @0xc0170 You want to be able to trigger actions based on a condition which would be "new commits since last review" or something like that?

0xc0170:
> IIUC @0xc0170 You want to be able to trigger actions based on a condition which would be "new commits since last review" or something like that?

Yes. Our use case: if PR is in "needs work" and it was updated(new commits/rebase), it should go into "ready for review" stage (as reviews were dismissed).

adbridge:
An additional tidy up which is similar to what @0xc0170 mentioned is that if a PR is closed rather than merged we would like to be able to removal all labels matching a specific regexp. We use labels to tag which release version a PR would go to E.g. 'release-version: x.y.z' however the label remove function seems to require you to specify exact matching labels which we would not know for any given PR. Is there some way of doing this currently or is that effectively another enhancement ?

jd:
>  Is there some way of doing this currently or is that effectively another enhancement ?

Yeah, that'd be another enhancement.
An easy one would be to delete **all** labels. Would that work for your use case?

(Deleting only certain label swould be also possible, but more expensive — there's a need to list the label first.)

adbridge:
[...]

PR #824, opened 2020-03-19
Add remove all labels option + reorganize cassettes

jd:
## feat(actions/label): add remove_all option

This new option allows to remove all labels of a pull request.

Relates to #360

## test: use class name in cassette directory

That should avoid potential conflicts between different test methods.

## test: rename simulator testing class


mergify_engine/tests/functional/base.py
```diff
@@ -275,7 +275,7 @@ def setUp(self):
         self.pr_counter = 0
         self.git_counter = 0
         self.cassette_library_dir = os.path.join(
-            CASSETTE_LIBRARY_DIR_BASE, self._testMethodName
+            CASSETTE_LIBRARY_DIR_BASE, self.__class__.__name__, self._testMethodName
```
sileht: Thx! I wanted to do that a long times ago and never did.

PR #832, opened 2020-03-20
fix: Always query the installation id

sileht:
This change always query the installation id from owner/repo. This will
raise a 404 if Mergify is not installed anymore, making the event to be
skipped, instead of raise 404/401 later in the process.

fixes MERGIFY-ENGINE-1BD


mergify_engine/clients/github.py
```diff
@@ -58,9 +62,15 @@ class GithubInstallationClient(common.BaseClient):
...
-        )
+        self.installation_id = github_app.get_client().get_installation_id(owner, repo)
+        if self.installation_id != installation_id:
+            LOG.error(
+                "installation id for repository diff from event installation id",
+                gh_owner=owner,
+                gh_repo=repo,
+                installation_id=self.installation_id,
+                expected_installation_id=installation_id,
+            )
```
jd: this will end up in sentry, do we want that?

sileht: Oh yes ! We should never happen, but I want check if I'm right.

PR #843, opened 2020-03-22
feat(check_runs): use httpx only

sileht:


mergify_engine/tests/functional/base.py
```diff
@@ -387,6 +389,15 @@ def setUp(self):
...
 
+        try:
+            self.cli_integration = github.get_client(
+                config.TESTING_ORGANIZATION, self.name, config.INSTALLATION_ID
+            )
+        except httpx.HTTPNotFound:
+            self.cli_integration = github.get_client(
+                config.TESTING_ORGANIZATION, self.name, config.INSTALLATION_ID
+            )
+
```
jd: why?

sileht: The repo have just been created sometimes installation is not yet present.

jd: if that's good enough let's do this

PR #860, opened 2020-03-24
chore: remove dead code and pygithub from runtime

sileht:


requirements.txt
```diff
@@ -35,22 +33,19 @@ jedi==0.16.0
...
-PyGithub==1.47
 Pygments==2.6.1
-PyJWT==1.7.1
 pyparsing==2.4.6
 python-dotenv==0.12.0
 python-json-logger==0.1.11
 pytz==2019.3
 PyYAML==5.3.1
-redis==3.3.11
+redis==3.4.1
```
jd: looks unrelated (and maybe wrong)

sileht: Thx for caching this !!!!

PR #866, opened 2020-03-24
refactor MergifyContext

sileht:


mergify_engine/mergify_context.py
```diff
@@ -119,7 +105,8 @@ def _get_valid_users(self):
         ]
         return bots + valid_collabs
 
-    def _get_consolidated_reviews(self):
+    @functools_bp.cached_property
```
jd: are you sure you want a property knowing it should be async in the future?

mergify_engine/mergify_context.py
```diff
@@ -185,44 +169,40 @@ def _get_consolidated_data(self):
...
+            "status-neutral": [
+                name for name, state in self.checks.items() if state == "neutral"
+            ],
             # NOTE(sileht): Not handled for now
             # cancelled, timed_out, or action_required
         }
 
-    def _get_checks(self):
+    @functools_bp.cached_property
+    def checks(self):
```
jd: ditto

mergify_engine/mergify_context.py
sileht: Yes, it should be cached, when I will render it async, I just have to grep all `functools_bp.cached_property` ;)

PR #867, opened 2020-03-24
feat(datadog): count request on github

sileht:


mergify_engine/clients/github.py
```diff
@@ -112,6 +113,17 @@ def items(self, url, api_version=None, list_items=None, **params):
...
+    def request(self, *args, **kwargs):
+        try:
+            return super().request(*args, **kwargs)
+        finally:
+            self._nb_requests += 1
+
+    def close(self):
+        super().close()
+        statsd.histogram("engine.github_session.size", self._nb_requests)
+        statsd.increment("engine.github_request.count", self._nb_requests)
```
jd: I wonder if we should prefix with `engine` here: if we either track them from the dashboard at some point, it'd be nice to be allowed to group them by app.
I think `github.http_session.size` and `github.http_request.size` maybe better names.

mergify_engine/clients/github.py
```diff
@@ -112,6 +113,17 @@ def items(self, url, api_version=None, list_items=None, **params):
...
                 break
 
+    def request(self, *args, **kwargs):
+        try:
+            return super().request(*args, **kwargs)
+        finally:
+            self._nb_requests += 1
+
+    def close(self):
+        super().close()
```
jd: I feel like this is going to be forgotten. Any chance to use this as a context manager?

mergify_engine/clients/github.py
sileht: Work for me.

mergify_engine/clients/github.py
sileht: It was ugly with context manager because the get_client also raise MergifyNotInstalled.

I look at it and found a cleaner way to do so.

sileht:
Replaced by https://github.com/Mergifyio/mergify-engine/pull/871

PR #887, opened 2020-03-28
feat: cache GET /check_runs

sileht:


mergify_engine/context.py
```diff
@@ -214,11 +215,23 @@ def get_consolidated_data(self, name):
         else:
             raise AttributeError(f"Invalid attribute: {name}")
 
+    def update_check_runs(self, check):
+        self.check_runs = [c for c in self.check_runs if c["name"] != check["name"]]
+        self.check_runs.append(check)
+
+    @functools_bp.cached_property
+    def check_runs(self):
```
jd, line 223:
```suggestion
    def pull_check_runs(self):
```

I think we were going for the `pull` prefix for things related to the PR?

mergify_engine/clients/github.py
```diff
@@ -139,7 +139,7 @@ def get_client(*args, **kwargs):
         reset = datetime.utcfromtimestamp(rate["core"]["reset"])
         now = datetime.utcnow()
         delta = reset - now
-        statsd.increment("engine.rate_limited")
+        statsd.increment("github.rate_limited")
```
jd, line 142: Looks wrong, I'll revert here.

PR #894, opened 2020-04-01
feat: This token never expire, so if we get a 401 that's mean:

sileht:
This is a sporadic Github is issue.
    User revoke it
    Mergify have been uninstalled.

To workaround the sporadic issue, we will retry. If Mergify got
uninstalled, on next strict merge polling all queues will be deleted
anyways.

If user revoke it, the erreur become legit and will be posted after some retry.


jd:
Fixed the commit message for you ;)

PR #901, opened 2020-04-03
fix: take more times before refreshing all PRs

sileht:
When something is merged in a branch, all `mergeable_state` of all opened
PRs of this branch become `unknown` it takes some times for Github to
reevalute them.

This changes the delay before refreshing all opened PRs on our sides.
To avoid many retry due to all mergeable_state being "unknown", just
after a "push" events.


jd:
@mergifyio rebase

PR #917, opened 2020-04-06
fix: revert retry on requests.HTTPError

jd:
This has been removed as part of 7b5879118c103f6ae57551594c769ea46cd704e9 but
requests is still used to getting subs from the dashboard.

Fixes MERGIFY-ENGINE-1CW

### DIFF
--- a/mergify_engine/exceptions.py
+++ b/mergify_engine/exceptions.py
@@ -13,6 +13,7 @@
 # under the License.
 
 import httpx
+import requests
 
 
 class MergifyNotInstalled(Exception):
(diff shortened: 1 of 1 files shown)


mergify_engine/exceptions.py
```diff
@@ -41,13 +42,20 @@ def need_retry(exception):  # pragma: no cover
         return max(exception.countdown, RATE_LIMIT_RETRY_MIN)
     elif isinstance(exception, MergeableStateUnknown):
         return BASE_RETRY_TIMEOUT
-    elif isinstance(exception, httpx.HTTPError) and (
-        exception.response is None or exception.response.status_code >= 500
+    elif (
+        isinstance(exception, httpx.HTTPError)
```
sileht, line 46: looks like a ( ) is missing

PR #920, opened 2020-04-07
Replace requests by httpx

jd:


mergify_engine/__init__.py
```diff
@@ -51,22 +49,3 @@ def fixup_sentry_reporting(event, hint):  # pragma: no cover
...
     )
-
-
-real_session_init = requests.sessions.Session.__init__
-
-
-def retring_session_init(self, *args, **kwargs):
-    real_session_init(self, *args, **kwargs)
-
-    adapter = requests.adapters.HTTPAdapter(max_retries=common.RETRY)
```
sileht, line 62: We must keep this, for testing e.g.: `tox -erecord`

jd: Do  you mean that `vcr` is using requests under the hood or something like that?
Since there's no more `requests` being used in test either AFAICT I'm confused.

sileht: pygithub uses request under the hood and we still using it in tests.

PR #928, opened 2020-04-09
feat: finish batch of engine actions

sileht:
This change to prepare batching for the engine action/command runner.


mergify_engine/context.py
```diff
@@ -52,6 +52,25 @@
 # https://developer.github.com/v4/enum/mergestatestatus/
 
 
+def get_logger(pull):
```
jd, line 55: Why isn't that a method?

mergify_engine/tasks/engine/__init__.py
```diff
@@ -161,24 +161,73 @@ def run(event_type, data):
...
     with github.get_client(owner, repo, installation) as client:
-        _run(client, event_type, data)
+        pull = get_github_pull_from_event(client, event_type, data)
 
+        if not pull:  # pragma: no cover
+            LOG.info(
+                "No pull request found in the event %s, ignoring",
+                event_type,
+                gh_owner=client.owner,
+                gh_repo=client.repo,
```
jd, line 171:
```suggestion
                gh_owner=owner,
                gh_repo=repo,
```

or is there any reason to access the client values?

mergify_engine/tasks/engine/__init__.py
```diff
@@ -161,24 +161,73 @@ def run(event_type, data):
...
+        context.get_logger(pull).debug(
+            "No need to proceed queue (got status of an old commit)", event_type,
         )
-        return
+        return True
+    elif (
+        event_type in ["status", "check_suite", "check_run"] and pull["merged"]
+    ):  # pragma: no cover
+        context.get_logger(pull).debug(
+            "No need to proceed queue (got status of a merged pull request)"
```
jd, line 217: That does not mean there is not a rule for this though. It sounds like a little premature optimization, no?
(I realize this is already in the code though so we can discuss that later)

mergify_engine/tasks/engine/__init__.py
```diff
@@ -161,24 +161,73 @@ def run(event_type, data):
...
+        for source in sources:
+            if source["event_type"] == "issue_comment":
+                issue_comment_sources.append(source)
+            elif not is_source_obsolete(ctxt.pull, source):
+                ctxt.sources.append(source)
+
+        commands_runner.spawn_pending_commands_tasks(ctxt)
+
+        if issue_comment_sources:
+            for source in issue_comment_sources:
```
jd, line 191: Are we sure that the sources are ordered here?

mergify_engine/tasks/engine/actions_runner.py
```diff
@@ -31,13 +31,17 @@
...
-        ):
-            return ""
+    if not (
+        ctxt.pull["merged"]
+        and any(
+            filter(
+                lambda s: s["event_type"] == "pull_request"
+                and s["data"]["action"] == "closed",
+                ctxt.sources,
+            )
```
jd, line 41:
```suggestion
            s for s in ctxt.sources if s["event_type"] == "pull_request" and s["data"]["action'] == "closed"
```
Not a big deal but it's usually easier to read.

mergify_engine/tests/functional/base.py
```diff
@@ -198,11 +198,15 @@ def _match(cls, data, expected_data):
...
+        r = self._session.get(
             f"https://gh.mergify.io/events-testing?counter={self._counter}",
             data=FAKE_DATA,
             headers={"X-Hub-Signature": "sha1=" + FAKE_HMAC},
-        ).json()
+        )
+        try:
+            return r.json()
+        except json.decoder.JSONDecodeError:
+            raise Exception(r.text)
```
jd, line 209:
```suggestion
            raise RuntimeError("Unable to decode JSON", r.text)
```

sileht: That was for debugging, the aioredis bug, I removed it.

mergify_engine/tasks/engine/actions_runner.py
sileht: done

mergify_engine/tasks/engine/__init__.py
sileht: Because of gevent, nothing is ordered already. @sileht

mergify_engine/context.py
sileht: It's also used without the context now, I moved it in utils.py

mergify_engine/tasks/engine/__init__.py
sileht: No that's the same.

mergify_engine/tasks/engine/__init__.py
sileht: I hesitated to removed all of that, it's represent only 1500 events per days. And about this one in particular, maybe people quand status/check rule after pull have been merged.

sileht: I decided to remove them.

mergify_engine/tasks/engine/__init__.py
jd: Alright, it might not be a big deal for now but they should be at some point.

PR #937, opened 2020-04-10
chore: move engine out of celery tasks

sileht:


jd:
Probably needs monitoring :)

PR #949, opened 2020-04-16
fix(github_events): do not reload subscription right away

jd:
When receiving a marketplace event, it should not reload the subscription right
away as the event might not be yet forwarded to the dashboard, which then could
return the old purchase.

Simply retrieve it later, when the dashboard had the chance to clear its own cache.


jd:
No, I don't think so:
- dashboard cache the old subscription
- engine gets the event
- engine queue in celery to forward the event
- engine delete the cache and re-download the subscription from the dashboard, which is old and caches it for an hour
- dashboard gets the event
- dashboard clears its cache

Ultimately we may want to change it so the marketplace events are only received by the dashboard and then the engine cache is cleared remotely by the dashboard. That'd fix it.

Issue #765, opened 2020-03-05
Pull request titles are being dropped from commit messages.


ocharles:
We use the `## Commit Message` feature of Mergify to set a final commit message in our PR, as we squash merge everything. However, recently commits have been losing titles:

See

![image](https://user-images.githubusercontent.com/20878/75966003-8ddd2180-5ec1-11ea-966b-9757c284abb3.png)

But here is how the PR looks:

![image](https://user-images.githubusercontent.com/20878/75966040-9d5c6a80-5ec1-11ea-905c-2b7d77ca40d5.png)

This seems to only happen if there is no blank line after `## Commit Message`. Note that this commit is correct:

![image](https://user-images.githubusercontent.com/20878/75966099-b6fdb200-5ec1-11ea-8fc0-4956f2a1d0c1.png)

![image](https://user-images.githubusercontent.com/20878/75966128-c250dd80-5ec1-11ea-9705-d55ac1548395.png)

jd:
This is indeed a bug. I'll fix it.

ocharles:
Is the correct/intended behavior for the GitHub title to be used in the first line of the commit message?

ocharles:
From reading the tests, it looks like you never want to use GitHub's title. I find this a bit of a shame, as the auto-generated title with the PR number in is really useful, and otherwise my team has to duplicate the title into the `## Commit Message` area.

jd:
Yes, that's indeed the intended behaviour for this feature.

However, we can add a different mode that implements what you want. We just have to find a good name for it.

ocharles:
That would be very much appreciated! How about ## Commit Details?

jd:
I'm wondering if adding support for `## Commit Message` to the `title+body` mode wouldn't be a solution?
Where `## Commit Message` would define the body, the title being taken from the pull request.

ocharles:
I'm not quite understanding. Are you suggesting that a PR with title "Foo" (on GitHub) and description:

```markdown
## Commit Message

Hello
```

Would have a final commit message of

```
Foo

Hello
```

?

If so, yes - that's what would work best for us, and is what used to happen if you had a blank line after `## Commit Message`.

jd:
Yes, that's my suggestion, as long as you have `commit_message: title+body` enabled on the merge action.

The idea would be that in `default`, `Commit Message` defines everything (title+body), whereas in the `title+body` mode, `Commit Message` only defines the message – not the title.

[...]

PR #950, opened 2020-04-16
Add templating feature for commit messages

jd:
## feat(context): add a PullRequest object

This is a wrapper that allows only accessing the pull request data that are
exposed through rules or templates.

## feat(merge): introduce templating system in commit message

This adds a feature that allows to use pull request attributes in commit
messages using Jinja2 template.

Related to #765


jd:
Need some test case for template syntax errors

ocharles:
This is a really nice solution!

jd:
Thanks @ocharles ! I was about to ping you to get your opinion on this and to see if it'd fix your use case. :) I guess that's the case!

mergify_engine/actions/merge/action.py
```diff
@@ -181,16 +239,27 @@ def _merge(self, ctxt):
...
-            }
+        data = {}
+
+        try:
+            commit_title_and_message = self._get_commit_message(
+                ctxt.pull_request, self.config["commit_message"],
+            )
+        except UndefinedVariables as e:
+            return (
+                "failure",
```
sileht, line 250: You can use `action_required` here. This makes the engine retries and put a nice icon in check list.

mergify_engine/actions/merge/action.py
```diff
@@ -35,6 +41,45 @@
...
+@dataclasses.dataclass
+class UndefinedVariables(Exception):
+    variables: set
+
+
+class PullRequestContext(jinja2.runtime.Context):
+    """This is a special Context that resolves any attribute first in the "pull request" object.
+
+
+    This allows to write {{author}} instead of {{pull_request.author}}."""
```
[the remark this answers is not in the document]
jd: If you do that it means it means it will request all values for all items of the pull request which makes it do extra HTTP requests to get e.g. the list of checks, etc.

mergify_engine/actions/merge/action.py
```diff
@@ -164,9 +209,23 @@ def _get_commit_message(pull, mode="default"):
...
+            def _render_template(template):
+                ast = env.parse(template)
+                undefined_vars = jinja2.meta.find_undeclared_variables(ast)
+                if undefined_vars:
+                    raise UndefinedVariables(undefined_vars)
+
+                template = env.template_class.from_code(
+                    env, env.compile(ast), env.make_globals(None)
+                )
+                return template.render().strip()
```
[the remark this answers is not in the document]
jd: See my comment above for `pull_request.items()`

mergify_engine/actions/merge/action.py
```diff
@@ -164,9 +209,23 @@ def _get_commit_message(pull, mode="default"):
         )
 
         if found and message_lines:
+            env = jinja2.sandbox.SandboxedEnvironment()
```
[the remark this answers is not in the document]
jd: That's a little less powerful because it'll raise on the first unknown attribute and won't list all of them… but it's good enough and much simpler so I took that approach.

mergify_engine/actions/merge/action.py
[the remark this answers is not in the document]
sileht: Oh, nice!

ocharles:
Great to see this merged! When do you folks release? Should I go ahead and give this a spin now?

jd:
We are using continuous deployment, so this should work!

doc/source/actions.rst
```diff
@@ -373,16 +373,45 @@ Defining the Commit Message
...
+    talk about them. This has been written by {{author}} and has been reviewed
+    by:
+
+    {% for user in approved_reviews_by %}
+    - {{user}}
+    {% endfor %}
+
+
+You can use any of the pull request :ref:`attributes` as variable — replacing
+``-`` with ``_`` in variables names.
```
ocharles, line 413: It looks like `id` isn't here. I'd like to be able to use this template:

```
{{title}} (#{{id}})

Body
```

to match what GitHub would do by default. Having the PR id means that commit messages link directly to the pull request they came from.

jd: I imagine you mean `number` and not `id`, those are different.

Indeed, it's not exposed as an attribute. That shouldn't be hard to add.

jd: https://github.com/Mergifyio/mergify-engine/pull/967

ocharles: I meant whatever shows up in the URL when I click on a pull request, yea :) Thanks!

PR #951, opened 2020-04-16
fix(doc): add the missing update command

jd:


sileht:
I see the tests failure, but I don't what have changed that make them failed.

PR #954, opened 2020-04-19
fix: ensure genreqs use py37

sileht:


tox.ini
```diff
@@ -12,6 +12,12 @@ deps = -r requirements.txt
 commands =
   pifpaf run -g MERGIFYENGINE_STORAGE_URL redis --port 6363 -- pifpaf run -g MERGIFYENGINE_CELERY_BROKER_URL redis --port 6364 -- pytest -v --pyargs mergify_engine {posargs}
 
+[testenv:py37]
+basepython = python3.7
+
+[testenv:py38]
+basepython = python3.8
```
jd, line 19: This is hardcoded in tox, so not needed AFAIK

tox.ini
```diff
@@ -40,6 +46,7 @@ skip_install = true
 commands = pip check
 
 [testenv:genreqs]
+basepython = python3.7
```
jd, line 49: what's the reason?

tox.ini
sileht: The tox code doesn't always work, even the doc said "can" not "will".

`If that’s not set tox can set a default value from the environment name ( e.g. py37 implies Python 3.7)`
`After resolving this value if the interpreter reports back a different version number than implied from the name a warning will be printed by default.`

On my machine, I always got the warning:
```
$ python3.8 --version
Python 3.8.2

$ python3.7 --version
Python 3.7.7

$ tox -e py37
/usr/lib/python3.8/site-packages/tox/config/__init__.py:592: UserWarning: conflicting basepython version (set 3.8, should be 3.7) for env 'py37';resolve conflict or set ignore_basepython_conflict
```

So I resolved the conflict...

tox.ini
sileht: The requirements.txt generated in python3.7 and python3.8 is different because some of our libraries use requirements filter like (python_version < 3.8).

Since production and circle-ci run python3.7, we must enforce requirements generation with 3.7

tox.ini
jd: I think this is because we have `basepython = python3` set in `testenv`, no?

tox.ini
jd: Make sense!

tox.ini
sileht: Good catch !

PR #989, opened 2020-04-27
Allow to use template in comment action

jd:
## feat(engine): report silent action when they fails

If an action with `silent_report = True` return a "failure" conclusion, we
report it to the user anyway rather than keeping it silent.

## feat(comment): allow to use Jinja2 template in the message


sileht:
I'm not sure about report failure of silent_checks, but we will see where it goes.

PR #1032, opened 2020-05-03
feat: copy the ci statuses to merge check

sileht:
In additional to the merge queue, this change also shows the current
state of CIs.


mergify_engine/actions/merge/action.py
```diff
@@ -96,6 +96,8 @@ def _required_statuses_in_progress(ctxt, missing_conditions):
         need_look_at_checks = []
         for condition in missing_conditions:
             if condition.attribute_name.startswith("status-"):
+                # TODO(sileht): Just return True here, no need to checks
```
jd, line 99: Why not doing it now?
(hint: write the reason in this code comment rather than in a GitHub comment ;))

mergify_engine/actions/merge/helpers.py
```diff
@@ -73,19 +73,22 @@ def merge_report(ctxt, strict):
...
         links = ", ".join((f"#{pull}" for pull in pulls))
-        queue_status = f"The following pull requests are queued: {links}"
+        summary += f"\n\nThe following pull requests are queued: {links}"
+
+    if rule and missing_conditions is not None:
+        summary += "\n\nThe required statuses are:\n"
+        for cond in rule["conditions"]:
+            if cond.attribute_name.startswith("status-"):
+                checked = " " if cond in missing_conditions else "X"
+                summary += f"\n- [{checked}] `{cond}`"
```
jd, line 89: Why only reporting the CI conditions?
I imagine that in 99% of the case it's only going to be `status-*` that are missing, but I wonder if it's a good idea to special-case that anyhow?

mergify_engine/clients/github.py
```diff
@@ -125,7 +125,6 @@ def items(self, url, api_version=None, list_items=None, **params):
 
     def request(self, method, url, *args, **kwargs):
         reply = None
-        LOG.debug("http request start", method=method, url=url)
```
jd, line 128: I'd have loved a second PR. :)

mergify_engine/actions/merge/helpers.py
sileht: If they is something, the check go to state cancelled instead of pending.
I think it's a good idea to put everything, so in case of cancellation, they will see why.

mergify_engine/actions/merge/action.py
```diff
@@ -93,31 +93,13 @@ def _required_statuses_in_progress(ctxt, missing_conditions):
...
 
-        need_look_at_checks = []
         for condition in missing_conditions:
             if condition.attribute_name.startswith("status-"):
-                need_look_at_checks.append(condition)
+                # A CI is missing we need to wait
+                return True
             else:
                 # something else does not match anymore
                 return False
```
jd, line 102: So this is
```python
return condition.attribute_name.startswith("status-")
```
?
This is just using the first missing condition as the answer?

sileht: yup

mergify_engine/clients/github.py
sileht: done

mergify_engine/actions/merge/action.py
sileht: I finally remove this commit, I don't get why yet but it doesn't work, so

PR #1087, opened 2020-05-20
feat(merge): new option to include PR number in commit message

eladb:
Adds another option to the `commit_message` merge action attribute called `pull_request` which is similar to `title+body` but also includes the PR number in the commit title (similar to how GitHub defaults when merging PRs manually).

### DIFF
--- a/doc/source/examples.rst
+++ b/doc/source/examples.rst
@@ -5,7 +5,7 @@
 ===============
 
 Mergify allows you to define a lot of specific rules. There is a large number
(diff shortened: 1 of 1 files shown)


jd:
Thank you! I'll take care of fixing the tests.

jd:
I closed this by mistake by pushing to your branch the wrong one @eladb and I cannot do it anymore.
I opened https://github.com/Mergifyio/mergify-engine/pull/1089 instead (keeping you as commiter).

PR #1089, opened 2020-05-20
feat(merge): include PR number in commit message title

jd:
This makes the feature similar to how GitHub defaults when merging PRs
manually.


jd:
@eladb I went ahead and just made that the default. That seems like a good thing.
We can add another option later if anybody does not want that, but it seems it's a pretty common request to make it the default.

eladb:
> @eladb I went ahead and just made that the default. That seems like a good thing.
> We can add another option later if anybody does not want that, but it seems it's a pretty common request to make it the default.

Sure. Makes sense. Didn't want to be opinionated about, but definitely makes sense that this is the default.

PR #1098, opened 2020-05-22
feat(merge): never cancel a closed PR

sileht:


mergify_engine/actions/merge/action.py
```diff
@@ -76,6 +76,14 @@ def run(self, ctxt, rule, missing_conditions):
                     q.remove_pull(ctxt.pull["number"])
 
     def cancel(self, ctxt, rule, missing_conditions):
+        q = queue.Queue.from_context(ctxt)
+        if ctxt.pull["state"] == "closed":
+            output = helpers.merge_report(ctxt, self.config["strict"])
+            if output:
+                if self.config["strict"] == "smart":
+                    q.remove_pull(ctxt.pull["number"])
```
jd, line 84: I'm wondering if you shouldn't do it anyway, just in case the config changed from `smart` to `false` or `true` in the meantime? Does it matter?

jd:
Typo in commit message `cancelled` -> `cancel`

mergify_engine/actions/merge/action.py
sileht: This have been done within another PR

PR #1099, opened 2020-05-22
bye bye celery [tosplit]

sileht:
Remaining to split:


sileht:
First part https://github.com/Mergifyio/mergify-engine/pull/1105

PR #1107, opened 2020-05-24
feat: move smart queue periodic task to worker

sileht:


mergify_engine/tests/unit/test_forwards.py
```diff
@@ -22,17 +22,10 @@
...
 from mergify_engine import utils
 from mergify_engine import web
 
 
-# NOTE(sileht): Celery magic, this just skip amqp and execute tasks directly
-# So all REST API calls will block and execute celery tasks directly
-tasks.app.conf.task_always_eager = True
-tasks.app.conf.task_eager_propagates = True
-
-
```
jd, line 35: I've a doubt that this should be part of this PR?

sileht: Good catch!

sileht:
I have to dismiss manually is that normal ?

mergify_engine/tests/unit/test_forwards.py
jd: Why I still see it there?

sileht: Because I clean base.py that was having these lines too...

PR #1108, opened 2020-05-24
chore: remove celery

sileht:


sileht:
Depends-On: https://github.com/Mergifyio/mergify-engine/pull/1107

setup.cfg
```diff
@@ -21,13 +21,13 @@ packages =
 include_package_data = true
 
 install_requires =
+    billiard
```
jd, line 24: Fine thought we might want to replace that
```
>>> billiard.current_process().name
'MainProcess'
```
Seems not useful anymore.

sileht: Can be replaced by dyno name from env for example

PR #1109, opened 2020-05-25
Pull request for mergify-config-update

jd:
## ci(mergify): rename no-review-needed to hotfix

Automatic fast merge should only by used for hotfixes. Let's make sure we have
the right label and a Sentry entry to make that happen.

## ci(mergify): only request review if there are no review already done

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -16,17 +16,18 @@ pull_request_rules:
       merge:
         strict: "smart"
         method: rebase
-  - name: automatic merge without review
+  - name: automatic merge for hotfix
     conditions:
       - base=master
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -67,10 +68,19 @@ pull_request_rules:
...
       request_reviews:
         teams:
           - devs
+  - name: warn on no Sentry
+    conditions:
+      - -body~=^Fixes MERGIFY-ENGINE-
+      - label=hotfix
+    actions:
+      comment:
+        message: Your pull request is a hotfix but does not fix a Sentry issue.
```
sileht, line 83: Ahaha :)

PR #1112, opened 2020-05-25
docs: add 'Using RegEx' sub section in Exemple Rules

CamClrt:


jd:
It looks like this is not the change you wanted to send.

PR #1113, opened 2020-05-25
fix(httpx): exception handling have changed

sileht:
httpx exceptions does not inherit from HTTPError anymore.

This may break again in the future...
https://github.com/encode/httpx/issues/949

So this change:
* removes the httpx monkeypatching
* put all http related error in clients/http.py
* adds catch HTTP Connection related Exception again

The rest of the code does not import httpx for caching exception anymore.

Fixes MERGIFY-ENGINE-1KZ MERGIFY-ENGINE-1KY


sileht:
done

PR #1118, opened 2020-05-26
feat: change default strict=smart behavior

sileht:
This change renames `smart` in `smart+fastpath`.

And `smart` becomes an alias of `smart+ordered`.


mergify_engine/actions/merge/action.py
```diff
@@ -50,7 +50,10 @@ class MergeAction(actions.Action):
             "merge", "squash", None
         ),
         voluptuous.Required("strict", default=False): voluptuous.Any(
-            bool, "smart", "smart+ordered"
+            bool,
+            voluptuous.All("smart", voluptuous.Coerce(lambda _: "smart+ordered")),
```
jd, line 54: 👏 that hack
I would have done:
```suggestion
            voluptuous.All("smart+ordered", voluptuous.Coerce(lambda _: "smart")),
```

and keep just `smart` as the canonical name, but as you wish.

sileht: I prefer to keep the long name for the code readability.

PR #1128, opened 2020-05-28
fix: use urllib3 only in tests

sileht:
Fixes MERGIFY-ENGINE-1M4 MERGIFY-ENGINE-1M5


sileht:
@Mergifyio rebase

sileht:
@Mergifyio rebase

PR #1139, opened 2020-05-30
refactor: make marketplace job async

jd:


mergify_engine/web.py
```diff
@@ -315,8 +321,8 @@ def sync_job_marketplace(event_type, event_id, data):
     except exceptions.MergifyNotInstalled:
         return
 
-    r = utils.get_redis_for_cache()
-    r.delete("subscription-cache-%s" % installation["id"])
+    r = await utils.get_aredis_for_cache()
```
sileht, line 324: The `get_client()` and `get_installation()` are still sync. We need to prepare the Async version of all http clients and Auth handlers.

PR #1144, opened 2020-05-31
feat: async http client

sileht:
## feat: make our httpx auth plugin async ready

To make httpx.Auth plugin working with sync and async client their must
not doing any http calls by themselves. Instead they should yield
httpx.Request() and the httpx.Client or httpx.AsyncClient will do the
IO stuff for us.

## feat: add async http client

This change adds a async github client, the base tests and
the vcr-py stubs for httpx.


mergify_engine/clients/github.py
```diff
@@ -67,39 +68,81 @@ def __init__(self, installation_id, owner, repo):
...
+        return self._cached_token.token
+
+    def _get_access_token(self):
+        now = datetime.utcnow()
+        if not self._cached_token:
+            return None
+        elif self._cached_token.expiration <= now:
             LOG.info(
-                "Token expired",
+                "Token expirerd",
```
jd, line 126:
```suggestion
                "Token expired",
```

mergify_engine/clients/github.py
```diff
@@ -67,39 +68,81 @@ def __init__(self, installation_id, owner, repo):
...
-                datetime.fromisoformat(r.json()["expires_at"][:-1]),  # Remove the Z
-            )
-            LOG.info(
-                "New token acquired",
-                gh_owner=self.owner,
-                gh_repo=self.repo,
-                expire_at=self._cached_token.expiration,
-            )
-        return self._cached_token.token
+        """ Legacy method for backport/copy actions"""
```
jd, line 138:
```suggestion
        """Legacy method for backport/copy actions"""
```

mergify_engine/clients/github_app.py
```diff
@@ -51,36 +51,43 @@
...
 
     def __init__(self):
         self.jwt = None
         self.jwt_expiration = None
+        self.lock = threading.Lock()
 
-    def get_or_create_jwt(self):
+    def get_or_create(self, force=False):
         now = int(time.time())
+        if force or self.jwt is None or self.jwt_expiration <= now:
```
jd, line 64: Hm I'd say this needs to be done with the lock to be safe, no?

sileht: yep, even both to fast and safe.

jd: Sorry, I don't understand your reply.

If you don't lock before the `if`, this can happen:

1. Thread 1 evaluates the `if` and it's True
2. Thread 1 locks and starts renewing the JWT
3. Thread 1 changes `self.jwt_expiration` to the future
3. Thread 2 check the if and sees that the expiration time is correct
3. Thread 2 returns the current `self.jwt` which is actually outdated

mergify_engine/clients/github_app.py
```diff
@@ -144,10 +152,76 @@ def _get_installation(self, url):
...
+            raise exceptions.MergifyNotInstalled()
+
+    async def _get_installation(self, url):
+        installation = (await self.get(url)).json()
+        installation["permissions_need_to_be_updated"] = False
+        expected_permissions = EXPECTED_MINIMAL_PERMISSIONS[installation["target_type"]]
+        for perm_name, perm_level in expected_permissions.items():
+            if installation["permissions"].get(perm_name) != perm_level:
+                LOG.debug(
+                    "mergify installation doesn't have required permissions",
```
jd, line 202:
```suggestion
                    "The Mergify installation doesn't have the required permissions",
```

PR #1147, opened 2020-06-01
Pull request for async-http-client-part2

sileht:


mergify_engine/tests/functional/test_engine.py
```diff
@@ -177,7 +175,10 @@ def test_debugger(self):
 
         self.setup_repo(yaml.dump(rules))
         p, _ = self.create_pr()
-        debug.report(p.html_url)
+
+        # NOTE(sileht): Run is thread due to not mess the main asyncio loop
```
jd, line 179:
```suggestion
        # NOTE(sileht): Run in a thread to not mess with the main asyncio loop
```

mergify_engine/tests/functional/actions/test_request_reviews.py
```diff
@@ -82,7 +82,11 @@ def test_request_reviews_already_max(self):
                 {
                     "name": "request_reviews",
                     "conditions": [f"base={self.master_branch_name}"],
-                    "actions": {"request_reviews": {"users": ["mergify-test1"]}},
+                    "actions": {
+                        "request_reviews": {
+                            "users": ["mergify-test1", "mergify-test-2"]
```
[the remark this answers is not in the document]
sileht: I will move out all tests changes.

mergify_engine/utils.py
```diff
@@ -107,6 +108,18 @@ def get_pull_logger(pull):
     )
 
 
+async def async_main(*awaitables):
+    # NOTE(sileht): This looks useless but
+    #   asyncio.run(asyncio.gather(...))
+    # does not work, because when the gather() coroutine is created, it need the current
+    # loop but asyncio.run() haven't create it yet, so it doesn't work.
```
jd, line 115: 👍

jd, line 115:
```suggestion
    # loop but asyncio.run() haven't created it yet, so it doesn't work.
```

mergify_engine/utils.py
```diff
@@ -107,6 +108,18 @@ def get_pull_logger(pull):
     )
 
 
+async def async_main(*awaitables):
+    # NOTE(sileht): This looks useless but
+    #   asyncio.run(asyncio.gather(...))
+    # does not work, because when the gather() coroutine is created, it need the current
```
jd, line 114:
```suggestion
    # does not work, because when the gather() coroutine is created, it needs the current
```

Issue #1145, opened 2020-06-01
Enhance error reporting when organization use SAML SSO


sileht:
Currently, we log that and just report a generic auth failure.
```
authentification failure, will retry another token: remote: The `xxxxx' organization has enabled or enforced SAML SSO. To access
remote: this repository, you must re-authorize the GitHub App `Mergify`.
fatal: unable to access 'https://github.com/xxxx/xxxx/': The requested URL returned error: 403
```

I have no idea how the github SAML SSO works, but I'm sure we can do better to make them aware they have to do something on their side to make rebase and backport working.

jd:
There's a button in your GitHub interface to authorize the app with your SSO access.
I know you can do that for personal tokens as described here: https://help.github.com/en/github/authenticating-to-github/authorizing-a-personal-access-token-for-use-with-saml-single-sign-on

We should link to the GitHub doc if we can find it for apps.
Maybe https://help.github.com/en/github/authenticating-to-github/authorizing-oauth-apps#oauth-apps-and-organizations

I think it might as simpler as relogging to Mergify and/or renewing their token.

https://help.github.com/en/github/authenticating-to-github/about-authentication-with-saml-single-sign-on says:

> To use the API or Git on the command line to access protected content in an organization that uses SAML SSO, you will need to use an authorized personal access token over HTTPS or an authorized SSH key. OAuth App access tokens are authorized by default.

So it should work by default, except maybe if SAML was added after Mergify was enabled?

PR #1148, opened 2020-06-02
fix: report SAML issue as fatal

sileht:
Currently this is catched as AuthentificationFailure making Mergify
retrying for ever. This change mark the SAML message as fatal. So
the error message coming from Github is reported on the PullRequest.

Related #1145
Fixes MERGIFY-ENGINE-1MG


mergify_engine/branch_updater.py
```diff
@@ -44,20 +45,23 @@ class AuthentificationFailure(Exception):
...
-    b"remote contains work that you do": BranchUpdateNeedRetry,
-    b"remote end hung up unexpectedly": BranchUpdateNeedRetry,
-    b"cannot lock ref 'refs/heads/": BranchUpdateNeedRetry,
-    b"Could not resolve host": BranchUpdateNeedRetry,
-    b"Operation timed out": BranchUpdateNeedRetry,
-    b"No such device or address": BranchUpdateNeedRetry,
-    b"Protected branch update failed": BranchUpdateFailure,
-    b"Couldn't find remote ref": BranchUpdateFailure,
-}
+GIT_MESSAGE_TO_EXCEPTION = collections.OrderedDict(
```
jd, line 48: Why does this need to be ordered?

sileht: SAML error also match 'The requested URL returned error, 403'. So first match first raise.

PR #1165, opened 2020-06-05
feat(actions): use cancelled conclusion and report != success by default

jd:


0xc0170:
I am trying to understand this "cancelled" as it changed the behavior in our PRs yesterday (blocking Prs now as they are red).

See https://github.com/ARMmbed/mbed-os/pull/13040/checks?check_run_id=754069345

jd:
We might be reporting too much, added https://github.com/Mergifyio/mergify-engine/pull/1171

0xc0170:
Thanks, but does it fix? I run `refresh` command on PR 13040 referenced above, still it's marked as "red"

jd:
They won't disappear, you can't delete a check run. That'll only work for new PRs.

Issue #358, opened 2019-03-27
Allow to ignorecase for regular expression (and maybe other regex flags?)


sileht:
We currently can't ignore the case for regular expression. We should find a way to allow it.

Lemmingh:
Hi,

What about adding a note to documentation saying that
Mergify uses Python regular expression with [search()](https://docs.python.org/library/re.html#re.search).
Then users can include `(?aiLmsux)` flags in regular expression confidently.

PR #1173, opened 2020-06-10
docs: add RegEx using informations

CamClrt:


jd:
@CamClrt You should add `Fixes #358` in your commit message so the bug will be closed automatically when the PR is merged.

PR #1177, opened 2020-06-12
fix: reword delete head branch

jd:
This uses the summary field a bit better.


mergify_engine/actions/delete_head_branch.py
```diff
@@ -39,15 +39,16 @@ def run(self, ctxt, rule, missing_conditions):
                 )
                 if pulls_using_this_branch:
                     return (
-                        "success",
+                        "neutral",
```
sileht, line 42: Using neutral make the action retring again and again.

sileht: since the summary will not change and the PR is not going to received events anymore, this is not a big deal.

jd: Even if it retries, I don't think it's a big deal since it could be that the branch becomes deletable.

PR #1187, opened 2020-06-17
fix(tests): set enough request reviews

sileht:


sileht:
I'm not sure about what we want to test

`will_exceed_max` currently need to at least make two review requests to get the error message.

sileht:
I think it's fine, just wanted your double checking.

jd:
I was not sure either when I wrote the tests. It's a bit tricky because we don't have much users.

PR #1188, opened 2020-06-18
feat: add sponsor link in summary

sileht:


mergify_engine/debug.py
```diff
@@ -112,7 +112,7 @@ def report(url):
 
     pull_raw = client.item(f"pulls/{pull_number}")
     ctxt = context.Context(
-        client, pull_raw, [{"event_type": "mergify-debugger", "data": {}}]
+        client, pull_raw, cached_sub, [{"event_type": "mergify-debugger", "data": {}}]
```
jd, line 115: This sounds ok but unrelated :)

mergify_engine/web.py
```diff
@@ -233,14 +233,19 @@ def PullRequestUrl(v):
...
             data = client.item(f"pulls/{pull_number}")
         except http.HTTPNotFound:
             raise PullRequestUrlInvalid(message=("Pull request '%s' not found" % v))
 
         return context.Context(
-            client, data, [{"event_type": "mergify-simulator", "data": []}]
+            client,
+            data,
+            subscription,
+            [{"event_type": "mergify-simulator", "data": []}],
```
jd, line 248: Wondering how it was working before?

sileht: Sub and sources was missing but unused, so it worked, now it's used so we catch it with our tests.

PR #1213, opened 2020-06-25
Pull request for no-install-id

sileht:
## chore: reset event testing endpoint

testing redis may have data from other tests

### DIFF
--- a/mergify_engine/clients/github.py
+++ b/mergify_engine/clients/github.py
@@ -62,17 +62,24 @@ class CachedToken:
 
 
 class GithubActionAccessTokenAuth(httpx.Auth):
+    def __init__(self):
+        self._installation = {
+            "id": config.ACTION_ID,
+            "permissions_need_to_be_updated": False,
+        }
+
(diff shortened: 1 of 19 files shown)


mergify_engine/clients/github.py
```diff
@@ -100,18 +116,42 @@ def auth_flow(self, request):
...
-        return httpx.Request(
-            "POST",
-            f"{config.GITHUB_API_URL}/app/installations/{self.installation_id}/access_tokens",
-            headers=headers,
+        return httpx.Request(method, url, headers=headers)
+
+    def _set_installation(self, installation_response):
+        if installation_response.status_code == 404:
+            LOG.debug(
+                "mergify not installed",
```
jd, line 141:
```suggestion
                "Mergify is not installed",
```

sileht:
The annoying thing is that Github API with ID is not documented and not all endpoints have an equivalent with ID.

After I dropped install_id, I plan to pass user_id/repo_id everywhere, and then see where we should use ID and where not.

PR #1219, opened 2020-06-26
refactor(worker): introduce StreamUnused

sileht:
This change introduce StreamUnused, so
`_translate_exception_to_retries` can recast some exception to it.

The first candidate is exceptions.MergifyNotInstalled.


jd:
image

mergify_engine/worker.py
```diff
@@ -78,6 +78,10 @@ class StreamRetry(Exception):
     retry_at: datetime.datetime
 
 
+class StreamUnused(Exception):
```
jd, line 81: That does not match the PR title

mergify_engine/worker.py
```diff
@@ -78,6 +78,10 @@ class StreamRetry(Exception):
...
+    pass
```
jd, line 82: maybe put the stream name here?

mergify_engine/worker.py
```diff
@@ -112,31 +116,25 @@ async def push(redis, owner, repo, pull_number, event_type, data):
...
 async def get_pull_for_engine(owner, repo, pull_number, logger):
-    try:
-        async with await github.aget_client(owner, repo) as client:
-            try:
-                pull = await client.item(f"pulls/{pull_number}")
-            except http.HTTPNotFound:
-                # NOTE(sileht): Don't fail if we received even on repo/pull that doesn't exists anymore
-                logger.debug("pull request doesn't exists, skipping it")
-                return
+    async with await github.aget_client(owner, repo) as client:
```
jd, line 119: Are you sure we don't need to catch `MergifyNotInstalled` anymore? :(

sileht: Yes, it shouldn't be here as it's caught later.

But, I see other kind of exceptions that should also be catched later like this one, that show up in Sentry while they should not.

I start thinking something isn't well catched somewhere in `_translate_exception_to_retries` or the `ThreadRunner`

PR #1225, opened 2020-06-29
fix: catch 404 earlier for /installation

sileht:
Fixes: MERGIFY-ENGINE-1PT


mergify_engine/clients/github.py
```diff
@@ -101,6 +101,15 @@ def auth_flow(self, request):
                     installation_response = yield self.build_installation_request(
                         url=installation_response.headers["Location"],
                     )
+                if installation_response.status_code == 404:
+                    LOG.debug(
+                        "mergify not installed or repository not found",
```
jd, line 106:
```suggestion
                        "Mergify not installed or repository not found",
```

PR #1231, opened 2020-06-30
Pull request for fix-logging-issue

sileht:


sileht:
> Do you think it's a bug in daiquiri/logging or the way it's handled?
> I'm surprised because that ought to work. I'm wondering what we do wrong here.

For me, it's related the memleak we have, but i'm not 100% sure.
(I will rewrite the commit message, the statement is a bit wrong)

sileht:
Damn, I just reread my patch, and it's due to a mock.patch I miss to remove in my testing...

sileht:
So I just push the cleanup

PR #1232, opened 2020-06-30
feat(worker): set user in Sentry

jd:


sileht:
I'm merging it for testing, but I'm not sure it works with asyncio.

PR #1235, opened 2020-07-01
Pull request for no-install-id

sileht:
## feat: use owner id for subscription

The subscription endpoint config name have changed to help
to migrate to the new endpoint.

The new config is already set everywhere.

### DIFF
--- a/mergify_engine/clients/github.py
+++ b/mergify_engine/clients/github.py
@@ -64,10 +64,12 @@ class CachedToken:
 
 class GithubActionAccessTokenAuth(httpx.Auth):
     def __init__(self):
-        self._installation = {
+        self.installation = {
             "id": config.ACTION_ID,
             "permissions_need_to_be_updated": False,
         }
+        # TODO(sileht): To be defined when we handle subscription for GitHub Action
+        self.owner_id = 0
 
     def auth_flow(self, request):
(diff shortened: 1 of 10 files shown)


mergify_engine/clients/github.py
```diff
@@ -64,10 +64,12 @@ def invalidate(self):
 
 class GithubActionAccessTokenAuth(httpx.Auth):
     def __init__(self):
-        self._installation = {
+        self.installation = {
             "id": config.ACTION_ID,
             "permissions_need_to_be_updated": False,
         }
+        # TODO(sileht): To be defined when we handle subscription for Github Action
```
jd, line 71:
```suggestion
        # TODO(sileht): To be defined when we handle subscription for GitHub Action
```

mergify_engine/config.py
```diff
@@ -83,7 +83,7 @@ def CommaSeparatedIntList(value):
         voluptuous.Required("GITHUB_API_URL", default="https://api.github.com"): str,
         # Mergify website for subscription
         voluptuous.Required(
-            "SUBSCRIPTION_URL", default="http://localhost:5000/engine/installation/%s"
```
jd, line 86: TODO clear that in prod :)

PR #1273, opened 2020-07-16
fix: detection of pull request closed by GitHub

sileht:


jd:
Should we have a test at some point?

sileht:
I will do another version, while this is needed, the aws bug is not due to this.

PR #1333, opened 2020-08-20
feat(config): support `.github/mergify.yml`

GuillaumeOj:


jd (changes requested):
Documentation should be updated too :)

GuillaumeOj:
The documentation is now update

jd (approved):

PR #1336, opened 2020-08-20
feat(config): update the doc for support '.github/mergify.yml'

GuillaumeOj:


jd (approved):
You can use `docs` as a conventional commit header.

PR #1342, opened 2020-08-26
fix(engine): don't shallow unexpected jinja issue

GuillaumeOj:
We need to reimplement  3e6a737 correctly and more importantly, we should have caught the issue with our tests.


jd:
This is good. I also don't understand why it works and how we break it in the first place.
I suggest keeping this as-is for now. I'll try to spend more time digging into it.

sileht:
I found what I missed, I change `PullRequestRulesSchema` and missed to update`PullRequestRules.from_list`.

I still wonder why no test have catch that.

sileht:
Can we just have the added tests in PR ? Even if they don't catch the issue we are looking for, they are a good addition to the test suite.

GuillaumeOj:
> Can we just have the added tests in PR ? Even if they don't catch the issue we are looking for, they are a good addition to the test suite.

Done

jd (approved):

PR #1345, opened 2020-08-27
fix(black): update files with the new Black's version

GuillaumeOj:


jd (approved):
You'll have to fix your commit message though.

jd (approved):

PR #1350, opened 2020-08-28
feat(dashboard): send 'github_app_authorization' to the dahsboard

GuillaumeOj:


mergify_engine/github_events.py
```diff
@@ -74,6 +74,7 @@ def get_ignore_reason(event_type, data):
         "check_suite",
         "check_run",
         "refresh",
+        "github_app_authorization",
```
jd, line 77: I don't think the engine is going to handle those events, so it's fine it returns "unexpected"

jd (changes requested):
This is not going to forward them

mergify_engine/github_events.py
GuillaumeOj: Ok, i'm going to look elsewhere

GuillaumeOj:
No change to be down on the engine, we just need to update the env right env variables: `WEBHOOK_FORWARD_EVENT_TYPES`

PR #1351, opened 2020-08-28
fix(queue): add bot_account in default config

sileht:
Fixes MERGIFY-ENGINE-1Q1

### DIFF
--- a/mergify_engine/actions/merge/queue.py
+++ b/mergify_engine/actions/merge/queue.py
@@ -78,11 +78,21 @@ class Queue:
         """
         config = self.redis.get(self._config_cache_key(pull_number))
         if config is None:
+            # FIXME(sileht): We should never ever pass here in theory, but
+            # Currently we can have race condition like:
+            # * smart queue coro: get next PR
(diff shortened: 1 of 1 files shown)


mergify_engine/actions/merge/queue.py
```diff
@@ -78,11 +78,21 @@ def get_config(self, pull_number: int) -> dict:
         """
         config = self.redis.get(self._config_cache_key(pull_number))
         if config is None:
+            # FIXME(sileht): We should never ever pass here in theorie, but
```
jd, line 81:
```suggestion
            # FIXME(sileht): We should never ever pass here in theory, but
```

mergify_engine/actions/merge/queue.py
```diff
@@ -78,11 +78,21 @@ def get_config(self, pull_number: int) -> dict:
...
+            # Currently we can have race like:
```
jd, line 82:
```suggestion
            # Currently we can have race condition like:
```

jd (approved):

PR #1352, opened 2020-08-28
fix(engine): add a test for the stream processor after a ReadError

GuillaumeOj:
Fixes MERGIFY-ENGINE-1RT

### DIFF
--- a/mergify_engine/tests/unit/test_worker.py
+++ b/mergify_engine/tests/unit/test_worker.py
@@ -18,6 +18,7 @@ import time
 from unittest import mock
 
 from freezegun import freeze_time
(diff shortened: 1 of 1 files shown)


sileht:
I think `ReadError` inherit from `RequestError`

GuillaumeOj:
I have changed the PR title since its not really a fix but just a test improvement

jd (changes requested):
Great!

For Sentry you can use:
`Fixes MERGIFY-ENGINE-1RT` in your commit so it triggers the close on Sentry side.

sileht (approved):

PR #1354, opened 2020-08-31
fix(worker): never fail to do monitoring

sileht:


mergify_engine/worker.py
```diff
@@ -584,22 +584,25 @@ async def smart_queue_processing_task(self):
...
+                if streams:
+                    latency = now - streams[0][1]
+                    statsd.timing("engine.streams.latency", latency)
+                else:
+                    statsd.timing("engine.streams.latency", 0)
 
-            statsd.gauge("engine.workers.count", self.worker_count)
+                statsd.gauge("engine.workers.count", self.worker_count)
+            except Exception:
+                LOG.warning("monitoring task failed", exc_info=True)
```
jd, line 605: do you want to send that back to sentry?

PR #1361, opened 2020-09-03
fix: handle case or merge is still pending

sileht:
Fixes MRGFY-58

### DIFF
--- a/mergify_engine/actions/merge/action.py
+++ b/mergify_engine/actions/merge/action.py
@@ -110,9 +110,13 @@ class MergeAction(actions.Action):
 
         if self._should_be_merged(ctxt):
             try:
-                return self._merge(ctxt)
-            finally:
+                conclusion, title, summary = self._merge(ctxt)
+                if conclusion is not None:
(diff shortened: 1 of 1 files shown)


mergify_engine/actions/merge/action.py
```diff
@@ -110,9 +110,13 @@ def run(self, ctxt, rule, missing_conditions):
...
             try:
-                return self._merge(ctxt)
-            finally:
+                conclusion, title, summary = self._merge(ctxt)
+                if conclusion is not None:
+                    q.remove_pull(ctxt.pull["number"])
+                return conclusion, title, summary
+            except Exception:
                 q.remove_pull(ctxt.pull["number"])
+                raise
```
jd, line 119: how is that different than `finally`?

jd (commented):
I guess we can't have test for those :(

mergify_engine/actions/merge/action.py
jd: Oh right, I misread the code. Good catch.

jd (approved):

PR #1366, opened 2020-09-09
feat: don't hidden rules with unmatched title/body/files

sileht:


jd:
Title should be "don't hide" :)

jd (approved):

PR #1379, opened 2020-09-14
refactor(subscription): implement Subscription class

jd:


mergify_engine/sub_utils.py
```diff
@@ -30,106 +32,139 @@
...
-digest = hashes.Hash(hashes.SHA256(), backend=default_backend())
-digest.update(config.CACHE_TOKEN_SECRET.encode())
-
-SECRET_KEY = digest.finalize()
-IV_BYTES_NEEDED = 12
-BLOCK_SIZE = 16
-TAG_SIZE_BYTES = BLOCK_SIZE
-
-
-def _encrypt(sub):
```
sileht, line 42: I would have not mixed the crypto stuff with the Subscription class.
the encrypt/decrypt method are generic, we can/should use it for anything stored in redis.

jd: Ok, that was my thinking too. I'll split it elsewhere.

sileht (approved):

PR #1381, opened 2020-09-15
chore: upgrade Python on Heroku

jd:


sileht:
Maybe we should set it for circle ci images ?

sileht:
I let you merge it, in case of it break the production again.

sileht (approved):

PR #1386, opened 2020-09-16
refactor(sub): move crypto functions into their own module

jd:


mergify_engine/tests/unit/test_crypto.py
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+from mergify_engine import crypto
+
+
+def test_encrypt():
+    x = "this is an amazing string, right? 🙄".encode()
+    assert crypto.decrypt(crypto.encrypt(x))
```
sileht, line 21: maybe we need also `== x` ?

mergify_engine/tests/unit/test_crypto.py
```diff
@@ -0,0 +1,23 @@
...
+    assert crypto.decrypt(crypto.encrypt(x))
```
sileht, line 22: Duplicate line

mergify_engine/tests/unit/test_crypto.py
jd: 🤦

mergify_engine/tests/unit/test_crypto.py
jd: This was actually voluntary to test there was no side effect on the crypto module itself.

sileht (approved):

PR #1390, opened 2020-09-18
test(debug): improve debug test

jd:
We now check the output to be sure it is accurate.


mergify_engine/debug.py
```diff
@@ -203,6 +199,9 @@ def report(url):
             )
 
     if pull_number:
+        # FIXME queues could also be printed if no pull number given
```
sileht, line 202: I don't get why this is moved

jd: Hehe. Because otherwise the output of the 2 URLs in the test looks like that:

```
test1:
<start of debug>
<end of debug>

test2:
<start of debug>
QUEUES:
<end of debug>
<pull request details>
```

so it's hard to check if the output is the same. With this change it becomes:
```
test1:
<start+end of debug>

test2:
<start+end of debug>
QUEUES:
<pull request details>
```

And I can test the output is the same with `startswith` ;)

sileht (approved):

PR #1397, opened 2020-09-21
feat: check subscription Features to enable engine features

jd:
Fixes MRGFY-83


sileht:
Maybe you should rerecord one of the priority queues tests, to be sure is still works.

sileht:
Otherwise it's good to go!

jd:
> Maybe you should rerecord one of the priority queues tests, to be sure is still works.

I managed to see them broken before https://github.com/Mergifyio/mergify-engine/pull/1396 was merged, so I think there's not even the need to re-record :)

sileht (approved):

PR #1398, opened 2020-09-21
refactor(auth): make auth repo-free

jd:
This makes it possible to authenticate without specifying a repository.


jd:
FWIW I tested it with a record of a functional test and it worked fine.

sileht (approved):

PR #1406, opened 2020-09-22
test: use requirements to run tests

jd:
That makes sure requirements are installed just before running the test and
after the project has been setup with its "wild" dependency list.

Related to MERGIFY-ENGINE-1TH


sileht (approved):
I'm a bit sad to merge that :(

jd:
Why is that?

PR #1407, opened 2020-09-22
refactor(engine): remove unused code

jd:


sileht (approved):
I was wondering how mergify can works without this :) And then I remember we have a new version of this called `extract_pull_numbers_from_event()`

PR #1417, opened 2020-09-23
build(deps): bump httpx and its deps

sileht:


jd (changes requested):
Merge #1406 first

sileht:
arf, we also have to wait the next pyvcr version...

sileht:
\o/

PR #1438, opened 2020-09-30
chore(action): rename check-runs to post_check

jd:
Fixes MRGFY-135


sileht:
The functional tests shoould be rerecorded to be sure we missed nothing.

sileht (approved):

PR #1440, opened 2020-09-30
feat: add +1 reaction when user post comment to mergify

sileht:
Related: MRGFY-70


mergify_engine/engine/commands_runner.py
```diff
@@ -46,6 +47,20 @@ def load_action(message):
...
+async def on_each_event(owner, repo, event_type, data):
+    if event_type == "issue_comment":
+        action = load_action(data["comment"]["body"])
+        if action:
+            async with await github.aget_client(owner) as client:
+                await client.post(
+                    f"/repos/{owner}/{repo}/issues/comments/{data['comment']['id']}/reactions",
+                    json={"content": "+1"},
+                    headers={
+                        "Accept": "application/vnd.github.squirrel-girl-preview+json"
```
jd, line 59: Why not using `api_version`?

jd (dismissed):

jd (changes requested):
same comment

mergify_engine/engine/commands_runner.py
sileht: fixed

jd (approved):

PR #1442, opened 2020-09-30
feat(worker): move worker states to redis

sileht:
Instead of storing pending stream in a local array, we lock
them with redis.

This is the last step to be able to spawn multiple dyno of the worker.

monitoring run twice but this should be not an issue it records only
gauge and timing.


mergify_engine/worker.py
```diff
@@ -209,38 +209,86 @@ def run(self):
...
-                self._pending_streams.remove(streams[0])
+                lock_refresher.cancel()
         else:
             yield
 
+    async def refresh_lock(self, stream):
+        while True:
+            try:
+                await self.redis.eval(
+                    self.STREAM_LOCK_UPDATE, 1, stream, self._hash, self._expire
```
jd, line 285: Looks like `STREAM_LOCK_UPDATE` return value is not checked. Maybe it should be so in case of collision at least we log or something?

jd:
What about using a lock free mechanism?
You could select the stream based on a simple hash of the string name modulo the worker number.
That would spread the workload more or less evenly but would make sure it is lock free.

The only question to answer is how to get the number of instances. That should be solvable.

mergify_engine/worker.py
```diff
@@ -209,38 +209,86 @@ def run(self):
...
 @dataclasses.dataclass
 class StreamSelector:
-    worker_count: int
+    local_worker_count: int
+    process_count: int
     redis: Any
 
-    _pending_streams: Set = dataclasses.field(init=False, default_factory=set)
+    _hash: str = dataclasses.field(init=False, default_factory=lambda: uuid.uuid4().hex)
+    _expire: int = 10
```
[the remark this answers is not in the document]
sileht: I have increase to 30 seconds.

mergify_engine/worker.py
sileht: > What about using a lock free mechanism?
> You could select the stream based on a simple hash of the stream name modulo the worker number.
> That would spread the workload more or less evenly but would make sure it is lock free.

The spread will really depend on the used hash, the thing to hash is currently the org name.
Do you have any recommendation/lead for the hash method ?
I will read some stuff on this to find something in the meantime.

> The only question to answer is how to get the number of instances. That should be solvable.

Simple version is just the env with the expected workers. Other just a key continually refreshed in redis by each worker.

sileht:
> What about using a lock free mechanism?
> You could select the stream based on a simple hash of the stream name modulo the worker number.
> That would spread the workload more or less evenly but would make sure it is lock free.

The spread will really depend on the used hash, the thing to hash is currently the org name.
Do you have any recommendation/lead for the hash method ?
I will read some stuff on this to find something in the meantime.

> The only question to answer is how to get the number of instances. That should be solvable.

Simple version is just the env with the expected workers. Other just a key continually refreshed in redis by each worker.

mergify_engine/worker.py
```diff
@@ -208,37 +208,36 @@ def run(self):
...
 class StreamSelector:
-    worker_count: int
     redis: Any
+    worker_id: int
+    worker_count: int
+
+    _pending_streams: List = dataclasses.field(init=False, default_factory=list)
 
-    _pending_streams: Set = dataclasses.field(init=False, default_factory=set)
+    def _forme(self, stream: bytes) -> bool:
```
jd, line 217:
```suggestion
    def _is_stream_for_me(self, stream: bytes) -> bool:
```

mergify_engine/worker.py
```diff
@@ -534,10 +533,20 @@ async def _consume_pulls(self, stream_name, pulls):
                 logger.error("failed to process pull request", exc_info=True)
 
 
+def get_process_index_from_env():
+    dyno = os.getenv("DYNO", None)
+    if dyno:
+        return int(dyno.split(".")[1]) - 1
```
jd, line 539:
```suggestion
        return int(dyno.rsplit(".", 1)[-1]) - 1
```
might be safer

mergify_engine/worker.py
```diff
@@ -551,35 +560,38 @@ class Worker:
     _worker_tasks: List = dataclasses.field(init=False, default_factory=list)
     _stream_monitoring_task: Any = dataclasses.field(init=False, default=None)
 
-    async def stream_worker_task(self, worker_id):
+    @property
+    def worker_count(self):
+        return self.worker_per_process * self.process_count
```
jd, line 565: Sounds that it might be a property of the `StreamSelector` itself, with `process_count` and `worker_per_process`?
Not sure, just suggesting in case it makes sense.

mergify_engine/worker.py
```diff
@@ -601,37 +613,51 @@ async def monitoring_task(self):
                     num=1,
                     withscores=True,
                 )
+                # NOTE(sileht): May not be always true with the next StreamSelector based
+                # on hash+modulo
```
jd, line 617: What might not be true?

jd (commented):
👍

jd (approved):
Approving as you tagged manual merge anyway.
LGTM, just a few suggestion.

mergify_engine/worker.py
sileht: It's also used by the monitoring task.

PR #1455, opened 2020-10-01
feat: rename status-X in check-X

sileht:
Closes #MRGFY-136


mergify_engine/context.py
```diff
@@ -240,11 +240,11 @@ def _get_consolidated_data(self, name):
         # tracked, and on big projects it can reach the rate limit very
         # quickly.
         # NOTE(sileht): Not handled for now: cancelled, timed_out, or action_required
-        elif name == "status-success":
+        elif name in ("status-success", "check-success"):
```
jd, line 243: if we replace it at parsing, is there's a chance we still encounter it here?

sileht: For Jinja templating, I will try to find a way to do the same I have done with pyparing, with Jinja. It would be cleaner.

sileht:
Using `pyparsing.replaceWith()` may confusing user. 
The user has `status-success` in its configuration, but it will see `check-success` inside the `Summary`.

Do you think it's a big deal ?

jd:
Good point. Yeah it might be better not to replace those in the end, or it might be confusing.

jd (approved):

PR #1460, opened 2020-10-02
refactor: use class for action result

sileht:


mergify_engine/actions/__init__.py
```diff
@@ -46,6 +47,32 @@ def get_commands():
     return dict((name, obj) for name, obj in get_classes().items() if obj.is_command)
 
 
+@dataclasses.dataclass  # type: ignore
```
jd, line 50: sounds like it's not needed here since it's an enum

mergify_engine/actions/__init__.py
```diff
@@ -46,6 +47,32 @@ def get_commands():
...
+    PENDING = None
+    CANCELLED = "cancelled"
+    SUCCESS = "success"
+    FAILURE = "failure"
+    NEUTRAL = "neutral"
+    ACTION_REQUIRED = "action_required"
+
+    def get_status(self):
+        if self == Conclusion.PENDING:
+            return "in_progress"
```
jd, line 61: Can we have an `Enum` for that too?

mergify_engine/actions/__init__.py
```diff
@@ -46,6 +47,32 @@ def get_commands():
...
+            return "completed"
+
+
+@dataclasses.dataclass  # type: ignore
+class Result:
+    conclusion: Conclusion
+    title: str
+    summary: str
+
+    def __repr__(self):
```
jd, line 72: this is generated by `dataclass`

mergify_engine/actions/__init__.py
```diff
@@ -46,6 +47,32 @@ def get_commands():
...
+    ACTION_REQUIRED = "action_required"
+
+    def get_status(self):
+        if self == Conclusion.PENDING:
+            return "in_progress"
+        else:
+            return "completed"
+
+
+@dataclasses.dataclass  # type: ignore
```
jd, line 66:
```suggestion
@dataclasses.dataclass
```

mergify_engine/actions/assign.py
```diff
@@ -51,10 +53,14 @@ def run(self, ctxt, rule, missing_conditions):
...
-                return (
-                    None,
+                return actions.Result(
+                    actions.Conclusion.PENDING,
                     "Unable to add assignees",
                     f"GitHub error: [{e.status_code}] `{e.message}`",
                 )
 
-        return ("success", "Assignees added", ", ".join(self.config["users"]))
+        return actions.Result(
```
jd, line 62: you might want to add type hint to the return type of the method

mergify_engine/actions/merge/queue.py
```diff
@@ -188,20 +189,19 @@ def handle_first_pull_in_queue(self, ctxt):
...
+        result = helpers.merge_report(ctxt, True)
+        if result:
             ctxt.log.info(
                 "pull request closed in the meantime",
-                conclusion=conclusion,
-                title=title,
-                summary=summary,
+                conclusion=result.conclusion,
+                title=result.title,
+                summary=result.summary,
```
jd, line 198: could do `result=result` though

jd (dismissed):

mergify_engine/actions/merge/queue.py
sileht: You read in my mind ;)

jd: Not really I guess?

mergify_engine/actions/post_check.py
```diff
@@ -86,13 +87,16 @@ def _post(self, ctxt, rule, missing_conditions):
...
+                check_api.Conclusion.FAILURE,
                 "Invalid summary template",
                 str(rmf),
             )
 
-        return ("failure" if missing_conditions else "success", title, summary)
+        if missing_conditions:
+            return check_api.Result(check_api.Conclusion.FAILURE, title, summary)
+        else:
+            return check_api.Result(check_api.Conclusion.SUCCESS, title, summary)
```
jd, line 99: This seems to be outside the scope of this patch?

mergify_engine/tests/functional/base.py
```diff
@@ -553,6 +553,8 @@ def tearDown(self):
 
         loop = asyncio.get_event_loop()
         loop.run_until_complete(web.shutdown())
+        for task in asyncio.all_tasks(loop):
+            task.cancel()
```
jd, line 557: looks unrelated :(

jd (changes requested):
A couple of changes seems unrelated though
Otherwise it looks beautiful 👍

jd:
My question around logging is, can we just do `log(result=result` rather than logging each field individually?

mergify_engine/actions/merge/queue.py
sileht: Oh I change `set_check_run` and `set_summary_check` signature to take `result` directly. But I didn't do it for logging to keep these attributes searcheable in datadog.

mergify_engine/actions/post_check.py
sileht: Hum, it's the same code as before with the new enum/dataclass

sileht:
> My question around logging is, can we just do `log(result=result)` rather than logging each field individually?

I wanted to keep it searcheable in datadog, but after some though this morning, I never ever use these attributes to filter logs, so.

mergify_engine/actions/post_check.py
jd: My bad I misread the original code with the hidden `if`

jd (approved):

PR #1484, opened 2020-10-07
fix: really return head sha stored in redis

sileht:


mergify_engine/tests/unit/test_synchronise.py
```diff
@@ -0,0 +1,70 @@
+# -*- encoding: utf-8 -*-
```
jd, line 1: this uses the American version of `synchronize` with a z, you should rename the file :)

mergify_engine/tests/unit/test_synchronise.py
```diff
@@ -0,0 +1,70 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from unittest import mock
+
+from mergify_engine import context
+from mergify_engine.engine import actions_runner
+
+
+def test_summary_synchronisation_cache():
```
jd, line 23:
```suggestion
def test_summary_synchronization_cache():
```

jd (changes requested):
rofl

jd (dismissed):
rofl

mergify_engine/engine/actions_runner.py
```diff
@@ -167,7 +167,7 @@ def delete_last_summary_head_sha(ctxt):
 
 def get_last_summary_head_sha(ctxt):
```
jd, line 168: adding a type hint would have prevented it ;)

jd (approved):

PR #1497, opened 2020-10-12
ci(mypy): install dependencies to get access to all type definitions

jd:


mergify_engine/web/__init__.py
```diff
@@ -46,20 +46,26 @@
 app.mount("/badges", badges.app)
 
 
+_AREDIS_STREAM: aredis.StrictRedis = None
+_AREDIS_CACHE: aredis.StrictRedis = None
```
sileht, line 50: sniff...

tox.ini
```diff
@@ -75,7 +75,6 @@ commands =
   black .
 
 [testenv:pep8]
-commands_pre =
```
sileht, line 78: ?

mergify_engine/web/__init__.py
jd: one could argue it's slightly cleaner that attaching things to an object like a cow-boy

tox.ini
jd: Read the commit title; we need the deps to be installed to have the definition of the libraries we depends on. :)

mergify_engine/web/__init__.py
sileht: That's fastapi recommendation !

sileht (approved):

PR #1501, opened 2020-10-14
fix: ignore 404 on branches

sileht:
Fixes MERGIFY-ENGINE-1V

### DIFF
--- a/mergify_engine/exceptions.py
+++ b/mergify_engine/exceptions.py
@@ -60,6 +60,12 @@ def should_be_ignored(exception):
         ):
             return True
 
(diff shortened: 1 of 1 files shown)


mergify_engine/exceptions.py
```diff
@@ -60,6 +60,12 @@ def should_be_ignored(exception):
         ):
             return True
 
+        # NOTE(sileht): branch is gone since we started to handle a PR
+        elif exception.status_code == 404 and "/branches/" in str(
```
jd, line 64: This seems really hackish brrr

jd (approved):

PR #1533, opened 2020-10-29
feat: post summary as soon as possible for new PR

sileht:
Close MRGFY-70


mergify_engine/engine/__init__.py
```diff
@@ -205,7 +206,63 @@ def run(client, pull, sub, sources):
...
+    if event_type != "pull_request":
+        return
+
+    if data["action"] != "opened":
+        return
+
+    redis = await utils.get_aredis_for_cache()
+    config_location_cache = (
+        f"config-location~{owner}~{data['pull_request']['base']['repo']['name']}"
+    )
```
jd, line 225: Could we have a method to construct this?
It looks to me it's in 2 different places now.

jd (changes requested):

jd (approved):

PR #1551, opened 2020-11-06
fix: rework ignored exception handling

sileht:
The current exception handling was expecting exact message match, but
some error have some noise around.

So just check the payload contains the expected message.

### DIFF
--- a/mergify_engine/exceptions.py
+++ b/mergify_engine/exceptions.py
@@ -12,6 +12,8 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import typing
+
 from mergify_engine.clients import http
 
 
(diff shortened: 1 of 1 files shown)


jd (approved):

mergify_engine/exceptions.py
```diff
@@ -30,34 +32,31 @@ def __init__(self, ctxt):
...
-            and exception.message == MISSING_REPOSITORY_DATA_MESSAGE
-        ):
-            return True
+def should_be_ignored(exception: Exception) -> bool:
+    if isinstance(exception, (http.HTTPClientSideError, http.HTTPServerSideError)):
+        for code, errors in IGNORED_HTTP_ERRORS.items():
+            if exception.status_code == code:
+                for error in errors:
+                    if error in exception.message:
+                        return True
```
jd, line 56:
```suggestion
                for error in IGNORED_HTTP_ERRORS.get(exception.status_code, []):
                    if error in exception.message:
                        return True
```
no?

jd (changes requested):

mergify_engine/exceptions.py
sileht: It's same and keeping `if exception.status_code == code:` looks clearer to me.

sileht: I prefer your finally :p, I just see it's also remove a loop

PR #1561, opened 2020-11-10
refactor: reorganize the merge action

sileht:
This reorganize and add type to merge action.


sileht:
I just move out the refactor from the queue PR

jd (approved):

PR #1582, opened 2020-11-17
chore: Fix typo in rebase section

colmose:
This is just a tiny typo fix 
`the that => that`


jd:
We apply semantic on the commit too. Let me know if you want to amend your commit or can do it for you. :)

Thanks for the PR!

colmose:
AH, so stick "chore: " or whatever onto the commit?

jd:
You can also use `docs:` as a prefix

jd (approved):

PR #1584, opened 2020-11-17
feat: add support for user with maintain permission

sileht:


jd (changes requested):
This is an important ACL change. I don't think `maintain` people can merge PRs right?
So this is a lot of power given. That might be something we offer as configurable?

sileht:
image

jd (approved):
I mixed `maintain` with `triage`, my bad.

PR #1585, opened 2020-11-17
fix: add support to new dashboard token format

sileht:
Some part of the engine expect the token to be the full github dict, some
other place expect it to be the `access_token` string.

So depending on how the Subscription is created, we may cache the dict
instead of the string.

This updates from_dict() to handle both case. The dashboard will be
updated to return the access_token only.

And when the cache we not contain dict anymore, we will drop this
workaround.

### DIFF
--- a/mergify_engine/subscription.py
+++ b/mergify_engine/subscription.py
@@ -69,11 +69,20 @@ class Subscription:
 
     @classmethod
     def from_dict(cls, owner_id, sub):
+        # FIXME(sileht): Remove me in January 2021, we should not support two kinds of
+        # payload, we have wrongly sent the token as dict instead of string, so we have
+        # to handle it for a while, in the meantime the dashboard will be updated to return
+        # only the access_token.
+        tokens = dict(
+            (login, token["access_token"] if isinstance(token, dict) else token)
+            for login, token in sub["tokens"].items()
+        )
+        # END FIXME
         return cls(
             owner_id,
             sub["subscription_active"],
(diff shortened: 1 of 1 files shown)


mergify_engine/subscription.py
```diff
@@ -69,11 +69,20 @@ def missing_feature_reason(owner):
...
     @classmethod
     def from_dict(cls, owner_id, sub):
+        # FIXME(sileht): Remove me in Janvier 2021, we should not support two kind of
+        # payload, we have wrongly token the token as dict instead of string, so we have
+        # to handle for a while, in the meantime the dashboard will be updated to return
+        # only the access_token.
+        tokens = dict(
+            (login, token["access_token"] if isinstance(token, dict) else token)
+            for login, token in sub["tokens"].items()
+        )
```
jd, line 79:
```suggestion
        # FIXME(sileht): Remove me in January 2021, we should not support two kinds of
        # payload, we have wrongly sent the token as dict instead of string, so we have
        # to handle it for a while, in the meantime the dashboard will be updated to return
        # only the access_token.
        tokens = dict(
            (login, token["access_token"] if isinstance(token, dict) else token)
            for login, token in sub["tokens"].items()
        )
```
The tokens are only cached for one hour, so I don't think we need to wait that long, right?

jd (changes requested):

mergify_engine/subscription.py
sileht: I just put a random date.

PR #1586, opened 2020-11-18
fix: always refresh queued pulls when queue changed

sileht:
In some case, like a merge is cancel because test fail. We was
refreshing only the next one in queue. Other queued pulls may have their
summary outdated, if the pull have no activity.


jd (approved):

sileht:
@Mergifyio update

PR #1589, opened 2020-11-19
fix(rebase): improve error handling of rebase command

### LINKED ISSUE #1588: @Mergify rebase runs slow and finally fails with F7978
### Context

Take this PR as an example: https://github.com/gradle/gradle/pull/14818#issuecomment-730095499

It takes Mergify ~30 minutes to perform the rebase action. And it finally failed with no reason, only a error code `F7978`. What shall I do now?

sileht:
In many case, we don't need to retry the rebase or even try it.
We known the none of the token we have will be able to push the modified
branches.

This case was handled by the merge action, but not the rebase one.

This change shares the logic between the merge and queue action.

Fixes #1588


jd (commented):
Is this going to be clear what the problem is for users?
In the case of a fork not giving the right permissions for example.

mergify_engine/branch_updater.py
```diff
@@ -61,6 +62,37 @@ class AuthenticationFailure(Exception):
 GIT_MESSAGE_TO_UNSHALLOW = set([b"shallow update not allowed", b"unrelated histories"])
 
 
+def pre_rebase_check(ctxt):
```
jd, line 65:
```suggestion
def pre_rebase_check(ctxt: context.Context) -> typing.Optional[check_api.Result]:
```

mergify_engine/branch_updater.py
```diff
@@ -259,4 +291,6 @@ def update_with_git(ctxt, method="merge", user=None):
             "Rebasing a branch for a forked private repository is not supported by GitHub"
         )
 
-    raise AuthenticationFailure("No valid OAuth tokens")
+    raise AuthenticationFailure(
+        f"No registered token allows Mergify to push to `{ctxt.pull['head']['label']}`"
```
jd, line 295:
```suggestion
        f"No registered tokens allow Mergify to push to `{ctxt.pull['head']['label']}`"
```
We should use plural form here since we're talking about the registered tokens.

mergify_engine/branch_updater.py
```diff
@@ -61,6 +62,37 @@ class AuthenticationFailure(Exception):
...
+        return check_api.Result(
+            check_api.Conclusion.FAILURE,
+            "Pull request can't be updated with latest base branch changes",
+            "Mergify needs the permission to update the base branch of the pull request.\n"
+            f"{ctxt.pull['base']['repo']['owner']['login']} needs to "
+            "[authorize modification on its base branch]"
+            "(https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).",
+        )
+    # If PR from a private fork but cannot be edited:
+    # NOTE(jd): GitHub removed the ability to configure `maintainer_can_modify` on private fork we which make strict mode broken
```
jd, line 81:
```suggestion
    # NOTE(jd): GitHub removed the ability to configure `maintainer_can_modify` on private fork which make rebase impossible
```

jd (dismissed):
nitpick :)

jd (approved):

PR #1594, opened 2020-11-20
fix(summary): enhance the way we display errors on summary

GuillaumeOj:


mergify_engine/rules/__init__.py
```diff
@@ -262,7 +262,14 @@ class InvalidRules(Exception):
...
     @staticmethod
     def _format_error(error):
-        msg = str(error)
+        msg = str(error.msg)
+
+        if error.error_type:
+            msg += f" for {error.error_type}"
+
+        if error.path:
+            path = " > ".join(map(str, error.path))
```
jd, line 271:
```suggestion
            path = " → ".join(map(str, error.path))
```
😬 ?

jd (commented):
👍

jd (approved):

PR #1598, opened 2020-11-23
fix(simulator): change the payload sent to the dashboard

GuillaumeOj:


mergify_engine/tests/functional/test_simulator.py
```diff
@@ -78,6 +76,57 @@ def test_simulator_with_token(self):
...
+            json={"pull_request": None, "mergify.yml": mergify_yaml},
+            headers={
+                "Authorization": f"token {config.EXTERNAL_USER_PERSONAL_TOKEN}",
+                "Content-type": "application/json",
+            },
+        )
+        assert r.status_code == 400, r.json()
+        assert r.json() == {
+            "errors_list": [
+                "expected str @ pull_request_rules → 0 → actions → label → remove → 0",
```
jd, line 125: unrelated but i'm wondering if we could replace 0 by `item 0` to make it more understandable

mergify_engine/web/simulator.py
```diff
@@ -64,33 +64,26 @@ def PullRequestUrl(v):
...
 async def voluptuous_errors(
     request: requests.Request, exc: voluptuous.Invalid
 ) -> responses.JSONResponse:
-    # FIXME(sileht): remove error at payload root
-    payload = voluptuous_error(exc)
-    payload["errors"] = []
+    # Replace payload by our own
+    payload: dict = {}
+    # FIXME(guillaume): rename key to "errors" when dahsboard transition is done
+    payload["errors_list"] = []
```
jd, line 80:
```suggestion
    # Replace payload by our own
    # FIXME(guillaume): rename key to "errors" when dahsboard transition is done
    payload = {"errors_list": []}
```

mergify_engine/web/simulator.py
```diff
@@ -64,33 +64,26 @@ def PullRequestUrl(v):
...
     if isinstance(exc, voluptuous.MultipleInvalid):
-        payload["errors"].extend(map(voluptuous_error, sorted(exc.errors, key=str)))
+        payload["errors_list"].extend(
+            map(voluptuous_error, sorted(exc.errors, key=str))
+        )
```
jd, line 84: or just
```suggestion
        payload = {"errors_list": list(map(voluptuous_error, sorted(exc.errors, key=str)))
```
and the same below.
keep it simple :)

jd (changes requested):
nitpick :)

mergify_engine/tests/functional/test_simulator.py
GuillaumeOj: It seems a good idea, I will do it in another PR.

jd (approved):

PR #1601, opened 2020-11-24
fix(rules): add a prefix before the index in the formated error

GuillaumeOj:


mergify_engine/rules/__init__.py
```diff
@@ -261,14 +261,19 @@ class InvalidRules(Exception):
     filename: str
 
     @staticmethod
-    def _format_error(error):
+    def _format_path_item(path_item):
+        if path_item == 0:
```
jd, line 265: hmmm and what if the error is in the item number 46738?

jd (changes requested):

jd (approved):

PR #1606, opened 2020-11-30
feat(actions): add the keyword add_users to assign

GuillaumeOj:


mergify_engine/actions/assign.py
```diff
@@ -28,14 +29,35 @@ class AssignAction(actions.Action):
...
+
+        return check_api.Result(
+            check_api.Conclusion.SUCCESS,
+            "Assignees added",
+            ", ".join(users),
+        )
+
+    def _add_assignees(
+        self, ctxt: context.Context, users_to_add: typing.List[str]
+    ) -> None:
```
sileht, line 58: I don't think mypy will like that, `check_api.Result` s returned sometimes.

mergify_engine/actions/assign.py
```diff
@@ -28,14 +29,35 @@ class AssignAction(actions.Action):
...
 
     def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Result:
+        users = list()
+        if self.config["users"]:
+            users = self._check_users(ctxt, self.config["users"])
+            self._add_assignees(ctxt, users)
+
+        if self.config["add_users"]:
+            users = self._check_users(ctxt, self.config["add_users"])
+            self._add_assignees(ctxt, users)
```
sileht, line 48: it can return a `check_api.Result` failure that we need to return to the caller too.

docs/source/actions.rst
```diff
@@ -30,12 +30,12 @@ The ``assign`` action assigns users to the pull request.
...
      - Default
      - Value Description
-   * - ``users``
+   * - ``add_users``
      - list of :ref:`data type template`
      -
      - The users to assign to the pull request.
 
-As the list of ``users`` is based on :ref:`data type template`, you can use
+As the list of ``add_users`` is based on :ref:`data type template`, you can use
```
jd, line 38:
```suggestion
The list of users in ``add_users`` is based on :ref:`data type template`, you can use
```

mergify_engine/actions/assign.py
```diff
@@ -28,26 +30,37 @@ class AssignAction(actions.Action):
     validator = {
         voluptuous.Required("users", default=[]): [
             voluptuous.Coerce(types.Jinja2, types.GitHubLogin)
-        ]
+        ],
```
jd, line 33: Add a note that this is deprecated and just legacy code.

mergify_engine/actions/assign.py
```diff
@@ -28,26 +30,37 @@ class AssignAction(actions.Action):
...
-            else:
-                wanted.add(user)
+        if self.config["users"]:
+            return self._add_assignees(ctxt, self.config["users"])
+
+        if self.config["add_users"]:
+            return self._add_assignees(ctxt, self.config["add_users"])
+
+        return check_api.Result(
+            check_api.Conclusion.FAILURE,
```
jd, line 49: Not sure it's worth failing here. Maybe just a silent succes is enough.

jd (changes requested):

mergify_engine/actions/assign.py
```diff
@@ -26,28 +28,41 @@
...
+        if self.config["users"]:
+            return self._add_assignees(ctxt, self.config["users"])
+
+        if self.config["add_users"]:
+            return self._add_assignees(ctxt, self.config["add_users"])
+
+        return check_api.Result(
+            check_api.Conclusion.SUCCESS,
+            "",
+            "",
```
jd, line 53: You can still add some text here. It's not going to be sent to the users, because the action has `silent_report = True`, but it's still good to see it in the logs.

mergify_engine/actions/assign.py
```diff
@@ -61,8 +76,29 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
...
+            return check_api.Result(
+                check_api.Conclusion.SUCCESS,
+                "Assignees added",
+                ", ".join(assignees),
+            )
         return check_api.Result(
-            check_api.Conclusion.SUCCESS,
-            "Assignees added",
-            ", ".join(self.config["users"]),
+            check_api.Conclusion.FAILURE,
```
jd, line 85: That should not be a failure. The only way this happens is that the users passed an empty list to `add_users`, and in that case, if we don't want to support it, we need to block it in the configuration.

I think it's ok to accept a list of empty users since we'll add `remove_users` at some point, so in that case just report a success.

jd (changes requested):

mergify_engine/actions/assign.py
```diff
@@ -26,28 +28,41 @@
...
+
+    def _add_assignees(
+        self, ctxt: context.Context, users_to_add: typing.List[str]
+    ) -> check_api.Result:
+        try:
+            assignees = self._wanted_users(ctxt, users_to_add)
+        except context.RenderTemplateFailure as rmf:
+            return check_api.Result(
+                check_api.Conclusion.FAILURE, "Invalid assignee", str(rmf)
+            )
```
jd, line 64: This cannot happen, right?
You catch this exception in `_wanted_users` already

jd (dismissed):

PR #1607, opened 2020-12-01
feat(actions): add the keyword remove_users to assign

GuillaumeOj:
We need to merge https://github.com/Mergifyio/mergify-engine/pull/1606 first


mergify_engine/actions/assign.py
```diff
@@ -78,6 +82,41 @@ def _add_assignees(
...
 
+    def _remove_assignees(
+        self, ctxt: context.Context, users_to_remove: typing.List[str]
+    ) -> check_api.Result:
+        try:
+            assignees = self._wanted_users(ctxt, users_to_remove)
+        except context.RenderTemplateFailure as rmf:
+            return check_api.Result(
+                check_api.Conclusion.FAILURE, "Invalid assignee", str(rmf)
+            )
```
jd, line 93: this cannot happen

jd (changes requested):

mergify_engine/actions/assign.py
```diff
@@ -43,6 +44,9 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
         if self.config["add_users"]:
             return self._add_assignees(ctxt, self.config["add_users"])
 
+        if self.config["remove_users"]:
```
jd, line 47: This won't work if you have both `add_users` and `remove_users` in the same action.

jd (changes requested):

jd (approved):

PR #1609, opened 2020-12-02
fix(simulator): handle errors with no path

GuillaumeOj:
Fixes MERGIFY-ENGINE-1XX


mergify_engine/web/simulator.py
```diff
@@ -75,10 +76,11 @@ async def voluptuous_errors(
     request: requests.Request, exc: voluptuous.Invalid
 ) -> responses.JSONResponse:
     # Replace payload by our own
+    print(type(exc))
```
jd, line 79:
```suggestion
```

jd (changes requested):

jd (approved):

PR #1613, opened 2020-12-04
test: move to github actions

sileht:


jd (approved):

jd:
@sileht looks like Redis is too old?

tox.ini
```diff
@@ -50,10 +50,10 @@ basepython = python3.9
 recreate = true
 skip_install = true
 deps =
-  -c requirements.txt
-  .
 commands_pre =
 commands =
+  bash -c "sed -e '/.\/uvloop.*/d' -e '/^-e ./d' requirements.txt > constraints.txt"
```
jd, line 55: can you add `constraints.txt` to `.gitignore` then?

jd (changes requested):

PR #1614, opened 2020-12-07
fix: ignore 404 for /access_token

sileht:
Fixes MERGIFY-ENGINE-1Y6

### DIFF
--- a/mergify_engine/clients/github.py
+++ b/mergify_engine/clients/github.py
@@ -173,7 +173,9 @@ class GithubAppInstallationAuth(httpx.Auth):
             if auth_response.status_code == 401:  # due to jwt
                 auth_response = yield self.build_access_token_request(force=True)
 
-            if auth_response.status_code == 403:
+            if auth_response.status_code == 404:
+                raise exceptions.MergifyNotInstalled()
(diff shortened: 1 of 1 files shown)


mergify_engine/exceptions.py
```diff
@@ -61,6 +61,12 @@ def should_be_ignored(exception: Exception) -> bool:
         ):
             return True
 
+        # NOTE(sileht): an installation return 404, mergify have been uninstalled
```
jd, line 64:
```suggestion
        # NOTE(sileht): an installation return 404, mergify has been uninstalled
```

mergify_engine/exceptions.py
```diff
@@ -61,6 +61,12 @@ def should_be_ignored(exception: Exception) -> bool:
...
+        elif exception.status_code == 404 and str(exception.request.url).endswith(
+            "/access_tokens"
+        ):
```
jd, line 67: Can't we check that in the place where we do the request? That sounds hackish to do that here. What's the good reason?

sileht: Yes, this call should be at a single place.

jd (approved):

PR #1630, opened 2020-12-09
feat(actions): add review_bot_account

GuillaumeOj:


mergify_engine/actions/review.py
```diff
@@ -39,13 +40,42 @@ class ReviewAction(actions.Action):
         voluptuous.Required("message", default=None): voluptuous.Any(
             types.Jinja2, None
         ),
+        voluptuous.Required("review_bot_account", default=None): voluptuous.Any(
```
jd, line 43:
```suggestion
        voluptuous.Required("bot_account", default=None): voluptuous.Any(
```
We only prefixed it in the merge action because it'd do multiple "actions", but here there's only one.

docs/source/actions.rst
```diff
@@ -189,6 +189,15 @@ The ``review`` action reviews the pull request.
      - :ref:`data type template`
      -
      - The message to write as a comment.
+   * - ``review_bot_account``
+     - string
+     -
+     - Mergify can impersonate a GitHub user to review a pull request.
+       If no ``review_bot_account`` is set, Mergify will review the pull request
+       itself. The user account **must** have already been
+       logged in Mergify dashboard once and have **write** or **maintain** permission.
```
jd, line 198:
```suggestion
       itself.
```

mergify_engine/actions/review.py
```diff
@@ -39,13 +40,42 @@ class ReviewAction(actions.Action):
...
+            if permission not in ("write", "maintain"):
+                return check_api.Result(
+                    check_api.Conclusion.ACTION_REQUIRED,
+                    (
+                        f"`{self.config['review_bot_account']}` account used as "
+                        "`review_bot_account` must have `write` or `maintain` permission, "
+                        f"not `{permission}`"
+                    ),
+                    "",
+                )
```
jd, line 77: I don't think that's a requirement. If I'm not mistaken a read-only user could post a review, it's just that it as less weight. And also `admin` could do it there.
cc @sileht for confirmation

jd (changes requested):
👍

mergify_engine/actions/review.py
```diff
@@ -95,8 +110,22 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
             ):
                 break
 
+        bot_account = self.config["bot_account"]
+        if bot_account:
```
jd, line 114:
```suggestion
        if bot_account := self.config["bot_account"]:
```

jd (dismissed):

jd (approved):

PR #1631, opened 2020-12-09
refactor: use Enum for strict merge action config

CamClrt:
Fixes MRGFY-182


mergify_engine/actions/merge_base.py
```diff
@@ -99,7 +107,9 @@ def get_strict_status(
     ) -> check_api.Result:
 
         summary = ""
-        if self.config["strict"] in ("smart+fasttrack", "smart+ordered"):
+        if self.config["strict"] in [
```
sileht, line 110: This now loop across all possible values instead of just "smart+fasttrack" and "smart+ordered".

sileht (dismissed):
You should also change the voluptuous schema to return an enum value instead of a bool/string with `voluptuous.Coerce()`, here:

https://github.com/Mergifyio/mergify-engine/blob/master/mergify_engine/actions/merge.py#L44-L52

mergify_engine/actions/merge_base.py
CamClrt: Is it better ?
```
if (
            self.config["strict"] == StrictMergeParameter["FASTTRACK"].value
            or StrictMergeParameter["ORDERED"].value
        ):
```

jd: 1. That's not a valid syntax
2. That being said, I understand what you mean, and yes that's what the logic you want indeed
3. You should not use the `.value` attribute but e.g. `StrictMergeParameter.ORDERED` and that's it

jd (dismissed):

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +54,27 @@ class PriorityAliases(enum.Enum):
...
         return PriorityAliases[v].value
     except KeyError:
         return v
 
 
+def strict_merge_parameter(v):
+    for name, member in StrictMergeParameter.__members__.items():
+        if v == member.value:
+            return member
+
```
jd, line 76: add a `raise ValueError` just in case if not found

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +54,27 @@ class PriorityAliases(enum.Enum):
     high = 3000
 
 
+class StrictMergeParameter(enum.Enum):
+    true = True
+    false = False
+    smart = "smart"
```
jd, line 60: if you re-read the code you see `smart` is actually `smart+ordered` so you should remove it

mergify_engine/actions/merge.py
```diff
@@ -41,14 +41,9 @@ class MergeAction(merge_base.MergeBaseAction):
...
         ),
-        voluptuous.Required("strict", default=False): voluptuous.Any(
-            bool,
-            voluptuous.All("smart", voluptuous.Coerce(lambda _: "smart+ordered")),
-            voluptuous.All(
-                "smart+fastpath", voluptuous.Coerce(lambda _: "smart+fasttrack")
-            ),
-            "smart+fasttrack",
-            "smart+ordered",
+        voluptuous.Required("strict", default=False): voluptuous.All(
```
jd, line 44:
```suggestion
        voluptuous.Required("strict", default=False): voluptuous.All(
```
the default is not part of the `Enum`

jd (changes requested):

mergify_engine/actions/merge.py
```diff
@@ -41,14 +41,11 @@ class MergeAction(merge_base.MergeBaseAction):
...
-            voluptuous.All("smart", voluptuous.Coerce(lambda _: "smart+ordered")),
-            voluptuous.All(
-                "smart+fastpath", voluptuous.Coerce(lambda _: "smart+fasttrack")
-            ),
-            "smart+fasttrack",
-            "smart+ordered",
+        voluptuous.Required(
+            "strict", default=merge_base.StrictMergeParameter.false
+        ): voluptuous.All(
+            voluptuous.Any("smart", "smart+fasttrack", "smart+ordered", bool),
```
jd, line 47: so `smart` won't work here since it's not valid to convert it

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +54,29 @@ class PriorityAliases(enum.Enum):
     high = 3000
 
 
+class StrictMergeParameter(enum.Enum):
+    true = True
+    false = False
+    # smart = "smart"
```
jd, line 60:
```suggestion
```

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +54,29 @@ class PriorityAliases(enum.Enum):
...
     except KeyError:
         return v
 
 
+def strict_merge_parameter(v):
+    for name, member in StrictMergeParameter.__members__.items():
+        if v == member.value:
+            return member
+        else:
+            raise ValueError(f"{v} is not an enum 'StrictMergeParameter'")
```
jd, line 77:
```suggestion
            raise ValueError(f"{v} is an unknown strict merge parameter")
```

jd (changes requested):

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +54,29 @@ class PriorityAliases(enum.Enum):
...
+
+
 def Priority(v):
     try:
         return PriorityAliases[v].value
     except KeyError:
         return v
 
 
+def strict_merge_parameter(v):
```
jd, line 72: what happens if I pass `smart+ordered` here as an argument?

jd (changes requested):

mergify_engine/actions/merge_base.py
CamClrt: 😓!

mergify_engine/actions/merge.py
```diff
@@ -41,14 +41,11 @@ class MergeAction(merge_base.MergeBaseAction):
...
-            voluptuous.All("smart", voluptuous.Coerce(lambda _: "smart+ordered")),
-            voluptuous.All(
-                "smart+fastpath", voluptuous.Coerce(lambda _: "smart+fasttrack")
-            ),
-            "smart+fasttrack",
-            "smart+ordered",
+        voluptuous.Required(
+            "strict", default=merge_base.StrictMergeParameter.false
+        ): voluptuous.All(
+            voluptuous.Any("smart+fasttrack", "smart+ordered", bool),
```
jd, line 47: Err so no we can't use `smart` anymore? :(

jd (changes requested):

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +54,33 @@ class PriorityAliases(enum.Enum):
...
```
jd, line 71: 👍

mergify_engine/tests/unit/actions/test_merge.py
```diff
@@ -271,3 +272,23 @@ def test_queue_summary_subscription(active, summary):
     with mock.patch.object(merge.queue.Queue, "from_context", return_value=q):
         action = merge.MergeAction(voluptuous.Schema(merge.MergeAction.validator)({}))
         assert summary == action.get_queue_summary(ctxt, q)
+
+
+def test_strict_merge_parameter():
```
jd, line 277: That's good.
You could use `pytest.mark.parametrize` to run all the scenario there but no big deal 😄

mergify_engine/actions/merge.py
```diff
@@ -41,14 +41,14 @@ class MergeAction(merge_base.MergeBaseAction):
         voluptuous.Required("rebase_fallback", default="merge"): voluptuous.Any(
             "merge", "squash", None
         ),
-        voluptuous.Required("strict", default=False): voluptuous.Any(
-            bool,
-            voluptuous.All("smart", voluptuous.Coerce(lambda _: "smart+ordered")),
-            voluptuous.All(
-                "smart+fastpath", voluptuous.Coerce(lambda _: "smart+fasttrack")
+        voluptuous.Required(
+            "strict", default=merge_base.StrictMergeParameter.false
```
jd, line 45: here's the catch: the default value will by passed to all your validators after.
Which mean the enum `false` will get through L46-L51.
However, it's not going to match the first one L48.

Your validators are correct though, so you need to pass a default value that works fine through all validator like a user provided value would 😄

jd:
This looks good but it's not enough. It fails at least for this:
```
Traceback (most recent call last):
  File "/home/runner/work/mergify-engine/mergify-engine/mergify_engine/engine/actions_runner.py", line 211, in exec_action
    return method(ctxt, rule)
  File "/home/runner/work/mergify-engine/mergify-engine/mergify_engine/actions/merge_base.py", line 217, in run
    q.add_pull(ctxt, typing.cast(queue.QueueConfig, self.config))
  File "/home/runner/work/mergify-engine/mergify-engine/mergify_engine/queue.py", line 105, in add_pull
    json.dumps(config),
  File "/opt/hostedtoolcache/Python/3.9.0/x64/lib/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/opt/hostedtoolcache/Python/3.9.0/x64/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/hostedtoolcache/Python/3.9.0/x64/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/hostedtoolcache/Python/3.9.0/x64/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
```

The configuration is serialized to be stored in Redis. We use `json.dumps` to serialize it and `json.dumps` does not know how to serialize an `Enum`. We need a custom JSON encoder/decoder to support that.

docs/source/_templates/navbar.html
```diff
@@ -16,6 +16,9 @@
       <li class="nav-item">
         <a class="nav-link" href="https://dashboard.mergify.io" target="_blank">Dashboard</a>
       </li>
+      <li class="nav-item">
+        <a class="nav-link" href="https://github.com/Mergifyio/mergify-engine/discussions" target="_blank">Discussions</a>
+      </li>
```
jd, line 21: I guess your rebase is wrong :(

jd (dismissed):

docs/source/_templates/navbar.html
CamClrt: 😕

mergify_engine/json.py
```diff
@@ -25,6 +25,7 @@ def register_type(enum_cls):
 class Encoder(json.JSONEncoder):
     def default(self, v):
         if isinstance(v, enum.Enum):
+            register_type(type(v))
```
jd, line 28: Again, the goal is not to register every type multiple times. Each type should be registered when declared.

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +54,33 @@ class PriorityAliases(enum.Enum):
     high = 3000
 
 
+class StrictMergeParameter(enum.Enum):
+    true = True
+    false = False
+    fasttrack = "smart+fasttrack"
+    ordered = "smart+ordered"
+
+
```
jd, line 63: Just register the the json type here?

mergify_engine/tests/unit/test_json.py
```diff
@@ -19,7 +19,7 @@
 from mergify_engine import json as mergify_json
 
 
-class Colour(enum.Enum):
+class Color(enum.Enum):
```
jd, line 22: Why changing this in this PR?

jd (changes requested):

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +55,38 @@ class PriorityAliases(enum.Enum):
...
+class StrictMergeParameter(enum.Enum):
+    true = True
+    false = False
+    fasttrack = "smart+fasttrack"
+    ordered = "smart+ordered"
+
+    @classmethod
+    def __init__(cls, v):
+        if cls.__name__ not in mergify_json.JSON_TYPES:
+            mergify_json.register_type(cls)
```
jd, line 67: You don't need to do it at `__init__`, you can just do it when the module is loaded.

mergify_engine/actions/merge_base.py
```diff
@@ -186,7 +215,10 @@ def run(
             q.remove_pull(ctxt)
             return report
 
-        if self.config["strict"] in ("smart+fasttrack", "smart+ordered"):
+        if (
+            self.config["strict"] is StrictMergeParameter.fasttrack
+            or self.config["strict"] is StrictMergeParameter.ordered
```
jd, line 220:
```suggestion
            self.config["strict"] in (StrictMergeParameter.fasttrack,
            StrictMergeParameter.ordered)
```

mergify_engine/actions/merge_base.py
```diff
@@ -99,7 +125,10 @@ def get_strict_status(
     ) -> check_api.Result:
 
         summary = ""
-        if self.config["strict"] in ("smart+fasttrack", "smart+ordered"):
+        if (
+            self.config["strict"] is StrictMergeParameter.fasttrack
+            or self.config["strict"] is StrictMergeParameter.ordered
```
jd, line 130:
```suggestion
            self.config["strict"] in (StrictMergeParameter.fasttrack,
            StrictMergeParameter.ordered)
```

mergify_engine/tests/unit/test_json.py
```diff
@@ -41,11 +41,10 @@ class Colour(enum.Enum):
...
 
 def test_encode_enum():
-    mergify_json.register_type(Colour)
     assert json.loads(mergify_json.dumps(with_enum)) == with_enum_encode
 
 
 def test_decode_enum():
-    mergify_json.register_type(Colour)
     json_file = mergify_json.dumps(with_enum)
+    mergify_json.register_type(type(Colour.BLUE))
```
jd, line 49:
```suggestion
    mergify_json.register_type(Colour)
```

mergify_engine/tests/unit/test_json.py
```diff
@@ -41,11 +41,10 @@ class Colour(enum.Enum):
 
 
 def test_encode_enum():
-    mergify_json.register_type(Colour)
     assert json.loads(mergify_json.dumps(with_enum)) == with_enum_encode
```
jd, line 44: I don't see how this test can work if you don't register the type?

jd (changes requested):

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +55,37 @@ class PriorityAliases(enum.Enum):
...
 
 
+class StrictMergeParameter(enum.Enum):
+    true = True
+    false = False
+    fasttrack = "smart+fasttrack"
+    ordered = "smart+ordered"
+
+
+if StrictMergeParameter.__name__ not in mergify_json.JSON_TYPES:
```
jd, line 65: That check is more likely to be done by `register_type`. It'd be stupid to have each caller of `register_type` doing this check!

jd (dismissed):

jd:
```
Traceback (most recent call last):
  File "/home/runner/work/mergify-engine/mergify-engine/mergify_engine/engine/actions_runner.py", line 211, in exec_action
    return method(ctxt, rule)
  File "/home/runner/work/mergify-engine/mergify-engine/mergify_engine/actions/merge_base.py", line 232, in run
    if self._should_be_merged(ctxt, q):
  File "/home/runner/work/mergify-engine/mergify-engine/mergify_engine/actions/merge.py", line 106, in _should_be_merged
    raise RuntimeError("Unexpected strict")
RuntimeError: Unexpected strict
```

mergify_engine/actions/merge.py
```diff
@@ -78,31 +76,31 @@ class MergeAction(merge_base.MergeBaseAction):
...
-        elif self.config["strict"] == "smart+fasttrack":
+        elif self.config["strict"] is merge_base.StrictMergeParameter.fasttrack:
             return ctxt.is_behind
-        elif self.config["strict"] is True:
+        elif self.config["strict"] is merge_base.StrictMergeParameter.true:
             return ctxt.is_behind
-        elif self.config["strict"]:
-            raise RuntimeError("Unexpected strict")
-        else:
+        elif self.config["strict"] in ("", merge_base.StrictMergeParameter.false):
```
jd, line 85: how could it be `""`?

jd:
morning advice:

`git grep strict_method`

There's at least one place I saw that needs to be updated for this new enum usage that you missed.

jd (approved):

PR #1635, opened 2020-12-10
docs: use lower case boolean values

jd:
This looks like what a default YAML serializer would write.

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -82,7 +82,7 @@ parameter:
        to.
    * - ``ignore_conflicts``
      - Boolean
-     - ``True``
+     - ``true``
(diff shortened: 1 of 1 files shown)


docs/source/actions.rst
```diff
@@ -616,7 +616,7 @@ benefit from any pull request attributes e.g. ``{{author}}`` and also these
 additional variables:
 
     * ``{{ check_rule_name }}`` the name of the rule that triggered this action.
-    * ``{{ check_succeed }}`` will be ``True`` if all conditions matches otherwise ``False``
```
sileht, line 619: Not sure about that one, because it's the Python bool

jd: Good catch!

sileht (approved):

PR #1637, opened 2020-12-11
refactor(actions): add typing

jd:


mergify_engine/actions/__init__.py
```diff
@@ -79,10 +80,8 @@ class Action(abc.ABC):
     # not create something that endup with a infinite loop of events
     allow_retrigger_mergify = False
 
-    @property
-    @staticmethod
-    @abc.abstractmethod
-    def validator():  # pragma: no cover
+    @abc.abstractproperty
```
sileht, line 83: abstractproperty is deprecated ?

sileht: https://docs.python.org/3/library/abc.html#abc.abstractproperty

sileht (approved):

PR #1644, opened 2020-12-14
refactor: simplify internal queue mechanism

sileht:
This change ensures we always pass context.Context to Queue API, instead
of just the pull request number.

### DIFF
--- a/mergify_engine/queue.py
+++ b/mergify_engine/queue.py
@@ -70,7 +70,7 @@ class Queue:
         )
 
     @classmethod
-    def from_context(cls, ctxt):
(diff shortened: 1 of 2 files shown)


mergify_engine/queue.py
```diff
@@ -219,7 +203,7 @@ def delete(self):
         self.redis.delete(self._redis_queue_key)
 
     @contextlib.contextmanager
-    def pull_request_auto_refresher(self, pull_number):
+    def pull_request_auto_refresher(self, pull_number: int) -> typing.Iterator:
```
jd, line 206: Could you type the `Iterator`?

jd (approved):

PR #1653, opened 2020-12-17
add json enum encoder and decoder

CamClrt:


mergify_engine/json_enum.py
```diff
@@ -0,0 +1,49 @@
+# -*- encoding: utf-8 -*-
```
jd, line 1: We want to support more than just `Enum` in the future, so that should just be `json` not `json_enum`.

mergify_engine/json_enum.py
```diff
@@ -0,0 +1,49 @@
...
+
+    raise ValueError(f"{v} is an unknown strict merge parameter")
+
+
+class StrictMergeParameterEncoder(json.JSONEncoder):
+    def default(self, v):
+        if isinstance(v, StrictMergeParameter):
+            return {"__enum__": True, "value": v.value}
+        else:
+            return super().default(v)
```
jd, line 43: This must work with any `Enum`, not only `StrictMergeParameter`.

mergify_engine/json_enum.py
```diff
@@ -0,0 +1,49 @@
...
+    if v == "smart":
+        return StrictMergeParameter.ordered
+    elif v == "smart+fastpath":
+        return StrictMergeParameter.fasttrack
+    else:
+        for name, member in StrictMergeParameter.__members__.items():
+            if v == member.value:
+                return member
+
+    raise ValueError(f"{v} is an unknown strict merge parameter")
```
jd, line 35: This should not be there. It has nothing to do with JSON encoding.

mergify_engine/tests/unit/test_json_enum.py
```diff
@@ -0,0 +1,38 @@
...
+with_enum = {
+    "name": "hello",
+    "conditions": [],
+    "actions": {"merge": {"strict": json_enum.StrictMergeParameter.fasttrack}},
+}
+
+
+def test_strict_merge_parameter_encoder():
+    json_file = json.dumps(with_enum, cls=json_enum.StrictMergeParameterEncoder)
+    assert isinstance(json_file, str) is True
```
jd, line 30: This actually test `json.dumps`, which *always* returns a `str`. So you're not really testing your code but more `json.dumps`.
I think you can just get rid of this test TBH.

mergify_engine/tests/unit/test_json_enum.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def test_decode_strict_merge_parameter():
+    json_file = json.dumps(with_enum, cls=json_enum.StrictMergeParameterEncoder)
+    assert (
+        json.loads(json_file, object_hook=json_enum.decode_strict_merge_parameter)
+        == with_enum
+    )
```
jd, line 38: 👍

jd (dismissed):

mergify_engine/json_enum.py
```diff
@@ -0,0 +1,49 @@
...
+        if isinstance(v, StrictMergeParameter):
+            return {"__enum__": True, "value": v.value}
+        else:
+            return super().default(v)
+
+
+def decode_strict_merge_parameter(dct):
+    if "__enum__" in dct:
+        return strict_merge_parameter(dct["value"])
+    return dct
```
jd, line 49:
```
def dumps(o):
   return json.dumps(o, cls=…)

def loads(s):
    return json.loads(s, object_hook=decode_strict_merge_parameter)
````

Comment (jd) [reply mergify_engine/json_enum.py:49]: goal:
```
from mergify_engine import json

json.dumps()
```

File: mergify_engine/actions/merge_base.py [hunk 7]
```diff
@@ -54,13 +54,33 @@ class PriorityAliases(enum.Enum):
     high = 3000
 
 
+class StrictMergeParameter(enum.Enum):
+    true = True
+    false = False
+    fasttrack = "smart+fasttrack"
+    ordered = "smart+ordered"
+
+
```
jd, line 63: Please don't add this yet, this is not needed in _this_ PR.

mergify_engine/actions/merge_base.py
```diff
@@ -54,13 +54,33 @@ class PriorityAliases(enum.Enum):
...
+    elif v == "smart+fastpath":
+        return StrictMergeParameter.fasttrack
+    else:
+        for name, member in StrictMergeParameter.__members__.items():
+            if v == member.value:
+                return member
+
+    raise ValueError(f"{v} is an unknown strict merge parameter")
+
+
```
jd, line 83: ditto

mergify_engine/json.py
```diff
@@ -0,0 +1,56 @@
...
+import json
+import enum
+
+from mergify_engine.actions.merge_base import strict_merge_parameter
+
+
+JSON_TYPES = {}
+
+
+def register_types(cls):
```
jd, line 23:
```suggestion
def register_type(cls):
```

mergify_engine/json.py
```diff
@@ -0,0 +1,56 @@
...
+                "__enum__": True,
+                "class": type(v).__name__,
+                "name": v.name,
+            }
+        else:
+            return super().default(v)
+
+
+def decode_enum(dct):
+    if "StrictMergeParameter" in dct:
```
jd, line 41: Why is that here?

mergify_engine/json.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+import json
+import enum
+
+from mergify_engine.actions.merge_base import strict_merge_parameter
```
jd, line 17:
```suggestion
```

mergify_engine/json.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def register_types(cls):
+    JSON_TYPES[cls.__name__] = cls
+
+
+class EnumEncoder(json.JSONEncoder):
+    def default(self, v):
+        if isinstance(v, enum.Enum):
+            register_types(type(v))
```
jd, line 30: This shouldn't be here. This looks smart, but it's actually error-prone. Take a look at that test:

```python
def test_decode_only():
    json.loads({"__enum__": True, "class": "StrictMergeParameter", "name": "smart"})
```

You think it'll work? No, because `json.dumps` has not been called and the type has not been registered before.

So it's better to make the user explicitely use `register_type` before they can use these JSON functions.

Plus, it's a little bit sub-optimal to call register each time, especially considering this function might be called thousands of times!

mergify_engine/json.py
```diff
@@ -0,0 +1,56 @@
...
+            return {
+                "__enum__": True,
```
jd, line 32: I'd rather use something like:
```suggestion
                "__pytype__": "enum",
```
As we're likely going to extend this to more than just `Enum` in the future.

mergify_engine/tests/unit/test_json.py
```diff
@@ -0,0 +1,53 @@
...
+    "conditions": [],
+    "actions": {"merge": {"strict": Colour.BLUE}},
+}
+
+with_strict_merge_parameter_enum = {
+    "name": "hello",
+    "conditions": [],
+    "actions": {"merge": {"strict": StrictMergeParameter.fasttrack}},
+}
+
```
jd, line 45: Your `Colour` test is enough.

jd (dismissed):

jd (dismissed):
Hmm you resolved comments but did not fix them. Mistake?

mergify_engine/tests/unit/test_json.py
```diff
@@ -0,0 +1,37 @@
...
+with_enum = {
+    "name": "hello",
+    "conditions": [],
+    "actions": {"merge": {"strict": Colour.BLUE}},
+}
+
+
+def test_decode_enum():
+    json.register_type(type(Colour.BLUE))
```
jd, line 35:
```suggestion
    json.register_type(Colour)
```

mergify_engine/json.py
```diff
@@ -0,0 +1,51 @@
...
+                "__pytype__": "enum",
+                "class": type(v).__name__,
+                "name": v.name,
+            }
+        else:
+            return super().default(v)
+
+
+def decode_enum(dct):
+    if "enum" in dct.values():
```
jd, line 38: This is going to load ALL the values of the dict and check in ALL of them… There's a better way to do this!

mergify_engine/json.py
```diff
@@ -0,0 +1,51 @@
...
+        cls_name = dct["class"]
+        for name, member in JSON_TYPES[cls_name].__members__.items():
+            if name == dct["name"]:
```
jd, line 41: You're going to iterate over all value of a dict to find a key… Again, there's a better way to do this.

mergify_engine/json.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+JSON_TYPES = {}
+
+
+def register_type(cls):
+    JSON_TYPES[cls.__name__] = cls
+
+
+class EnumEncoder(json.JSONEncoder):
```
jd, line 25:
```suggestion
class JSONEncoder(json.JSONEncoder):
```
This won't support only Enum in the future :)

mergify_engine/tests/unit/test_json.py
```diff
@@ -0,0 +1,37 @@
...
+    BLUE = 3
+
+
+with_enum = {
+    "name": "hello",
+    "conditions": [],
+    "actions": {"merge": {"strict": Colour.BLUE}},
+}
+
+
```
jd, line 33: You should also add a test where `dumps` is not used, only `loads`.

jd (changes requested):

mergify_engine/json.py
```diff
@@ -0,0 +1,52 @@
...
+                "__pytype__": "enum",
+                "class": type(v).__name__,
+                "name": v.name,
+            }
+        else:
+            return super().default(v)
+
+
+def decode_enum(v):
+    if isinstance(v, dict):
```
jd, line 38: I though this was only called when `v` was a dict by `json.loads`? 🤔

mergify_engine/tests/unit/test_json.py
```diff
@@ -0,0 +1,44 @@
...
+    "conditions": [],
+    "actions": {"merge": {"strict": Colour.BLUE}},
+}
+
+with_enum_encode = '{"name": "hello", "conditions": [], "actions": {"merge": {"strict": {"__pytype__": "enum", "class": "Colour", "name": "BLUE"}}}}'
+
+
+def test_encode_enum():
+    json.register_type(Colour)
+    assert json.dumps(with_enum) == with_enum_encode
```
jd, line 38: I think it's fine but be aware this can fail in theory.
A dict is not ordered, so the order in which the strings are printed in the string `with_enum_encode` might change. Which would break the test.

CamClrt: if I use `assert data.get("actions").get("merge").get("strict") == Colour.BLUE` is it better ?

mergify_engine/json.py
jd: That's actually a question… is that check necessary?

mergify_engine/tests/unit/test_json.py
jd: what's `data` in your example? 🤔

CamClrt: sorry :
```
data = json.loads(json_file)
assert data.get("actions").get("merge").get("strict") == Colour.BLUE
```

CamClrt: or I can write it like this:

```
assert (
        json.loads(json_file).data.get("actions").get("merge").get("strict")
        == Colour.BLUE
    )
```

mergify_engine/json.py
CamClrt: nop, I have checked the documentation & I have already corrected it in the new PR.

mergify_engine/tests/unit/test_json.py
jd: If you use `mergify.json.loads` therefore your just writing the same test as `test_decode_enum`.

What you should rather do is use `json.loads` (not the Mergify one) and check the returned data, so:

```python
assert json.loads(json_file) == {"name": "hello", "conditions": [], "actions": {"merge": {"strict": {"__pytype__": "enum", "class": "Colour", "name": "BLUE"}}}}
```

mergify_engine/json.py
jd: What new PR? Why isn't that fixed in this PR?

mergify_engine/tests/unit/test_json.py
```diff
@@ -0,0 +1,51 @@
...
+    "conditions": [],
+    "actions": {
+        "merge": {"strict": {"__pytype__": "enum", "class": "Colour", "name": "BLUE"}}
+    },
+}
+
+
+def test_encode_enum():
+    mergify_json.register_type(Colour)
+    assert mergify_json.dumps(with_enum) == json.dumps(with_enum_encode)
```
jd, line 45: I'm not sure `json.dumps` is supposedly idempotent so that might work by luck also.
Rather do:
```suggestion
    assert json.loads(mergify_json.dumps(with_enum)) == with_enum_encode
```

jd (changes requested):

mergify_engine/tests/unit/test_json.py
CamClrt: ok

jd (approved):

PR #1661, opened 2020-12-22
refactor(command_runner): tidy code, adds types

sileht:
This removes case that can't happen anymore, add types and split the big
method.

### DIFF
--- a/mergify_engine/engine/commands_runner.py
+++ b/mergify_engine/engine/commands_runner.py
@@ -13,6 +13,7 @@
 # under the License.
 
 import re
+import typing
 
 import daiquiri
 from datadog import statsd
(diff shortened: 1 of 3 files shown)


mergify_engine/engine/commands_runner.py
```diff
@@ -30,11 +34,31 @@
 COMMAND_MATCHER = re.compile(r"@Mergify(?:|io) (\w*)(.*)", re.IGNORECASE)
 COMMAND_RESULT_MATCHER = re.compile(r"\*Command `([^`]*)`: (pending|success|failure)\*")
 
+MERGE_QUEUE_COMMAND_MESSAGE = "Command not allowed on merge queue pull request."
```
jd, line 37: I don't see that used anywhere?

jd (approved):

PR #1663, opened 2020-12-22
refactor: renames Queue.get_queue() to Queue.from_queue()

sileht:


mergify_engine/queue.py
```diff
@@ -60,6 +60,17 @@ def from_context(cls, ctxt: context.Context) -> "Queue":
             ctxt.pull["base"]["ref"],
         )
 
+    def from_queue(self, ref: str) -> "Queue":
```
jd, line 63:
```suggestion
    def from_queue_name(self, ref: str) -> "Queue":
```

jd (dismissed):

mergify_engine/queue.py
sileht: Why not

mergify_engine/queue.py
```diff
@@ -135,7 +146,7 @@ def _remove_pull_from_other_queues(self, ctxt: context.Context) -> None:
                 score = self.redis.zscore(queue_name, ctxt.pull["number"])
                 if score is not None:
                     old_branch = queue_name.split("~")[-1]
-                    old_queue = self.get_queue(old_branch)
+                    old_queue = self.from_queue_name(old_branch)
```
jd, line 149: Actually, now that I reread the, I'm not sure it makes sense since it's a ref and not a queue name.

`get_queue` was better wrt this.

If we want `from_something` I'd suggest `from_ref` or `get_queue_from_ref`

jd (changes requested):
sorry about the ping pong

PR #1674, opened 2020-12-29
refactor: add SHAType for GitHub events

jd:


mergify_engine/check_api.py
```diff
@@ -46,7 +51,13 @@ class Result:
     annotations: typing.Optional[typing.List[str]] = None
 
 
-def get_checks_for_ref(ctxt, sha, **kwargs):
+def get_checks_for_ref(
+    ctxt: "context.Context", sha: github_types.SHAType, check_name: str = None
```
sileht, line 55: I would even use the `*` and `/` separator to ensure check_name is always passed as kwargs

sileht (approved):

PR #1686, opened 2020-12-31

jd:
new commits are pushed.

### DIFF
--- a/mergify_engine/github_events.py
+++ b/mergify_engine/github_events.py
@@ -133,7 +133,7 @@ async def filter_and_dispatch(
         if event["repository"]["archived"]:
             ignore_reason = "repository archived"
 
-        elif event["action"] == "opened":
+        elif event["action"] in ("opened", "synchronize"):
             try:
                 await engine.create_initial_summary(event)
             except Exception as e:


sileht (approved):

sileht:
If we do that, we also need to update: https://github.com/Mergifyio/mergify-engine/blob/master/mergify_engine/engine/__init__.py#L91-L94

sileht:
Hum I re self-requested my review to avoid automerge but it hasn't worked.

sileht:
> Hum I re self-requested my review to avoid automerge but it hasn't worked.

We don't use `#review-requested=0` so Mergify didn't care of my self-review-request.

jd:
@sileht no you need to change your review to -1

I'm not sure about why we need to check the summary. IIUC when synchronize happens there's 0 check run so we can post this first one as soon as we have the event being queued in the worker.
What wrong can happen?

sileht:
We store the state of the PR in the summary, on synchronize we copy the last known summary to the last head sha. We can't rely on event type to do the copy because engine may run after the git push but before we get the event. So the method a point copy it only when the summary is missing. We we create an empty summary the copy won't happen and we lost the state of the PR.

sileht:
This change is a good idea, but if we do it we must update `ensure_summary_on_head_sha()` to take care of this case to copy the last know summary also when the content of the current one is the temporary one.

PR #1688, opened 2021-01-01
Revert "fix(github_events): also send initial summary on synchronize events"

sileht:
Reverts Mergifyio/mergify-engine#1686


jd:
@sileht why?!

sileht:
I put the explanation there: https://github.com/Mergifyio/mergify-engine/pull/1686

PR #1689, opened 2021-01-01
feat: faster shutdown

sileht:
Don't wait for the engine thread to finish after 25 seconds

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -610,6 +610,7 @@ def get_process_index_from_env() -> int:
 @dataclasses.dataclass
 class Worker:
     idle_sleep_time: float = 0.42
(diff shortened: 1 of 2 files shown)


mergify_engine/worker.py
```diff
@@ -663,6 +664,11 @@ async def stream_worker_task(self, worker_id: int) -> None:
                 else:
                     LOG.debug("worker %s has nothing to do, sleeping a bit", worker_id)
                     await self._sleep_or_stop()
+            except asyncio.CancelledError:
+                # NOTE(sileht): We don't wait for the thread and just return, the thread
+                # will be kill when the program exit.
```
jd, line 669: killed
exits

jd (approved):

PR #1692, opened 2021-01-01
fix(mypy): enable warn_redundant_casts and strict_equality

jd:
This catches a few bugs and and more strict checks.

### DIFF
--- a/mergify_engine/actions/close.py
+++ b/mergify_engine/actions/close.py
@@ -32,7 +32,7 @@ class CloseAction(actions.Action):
     validator = {voluptuous.Required("message", default=MSG): types.Jinja2}
 
     def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Result:
-        if ctxt.pull["state"] == "close":
+        if ctxt.pull["state"] == "closed":
             return check_api.Result(
                 check_api.Conclusion.SUCCESS, "Pull request is already closed", ""
             )
--- a/mergify_engine/github_events.py
+++ b/mergify_engine/github_events.py
@@ -165,7 +165,9 @@ async def filter_and_dispatch(
         event = typing.cast(github_types.GitHubEventIssueComment, event)
         owner = event["repository"]["owner"]["login"]
         repo = event["repository"]["name"]
-        pull_number = event["issue"]["number"]
+        pull_number = typing.cast(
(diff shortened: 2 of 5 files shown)


mergify_engine/github_events.py
```diff
@@ -289,7 +298,7 @@ async def _get_github_pulls_from_sha(client, repo, sha, pulls):
     elif pull_number == -1:
```
jd, line 298: @sileht I think there was a bug here because AFAIU `redis.get` returns a `str` so it's never `== -1`. :(

sileht: yes, but by chance, since this occurs only when `pull_number is None`, we just don't cache the fact we known that no open pull request have this sha.

Also this code is useless since we don't do GitHub API call in this method anymore. I will remove it.

mergify_engine/actions/close.py
```diff
@@ -32,7 +32,7 @@ class CloseAction(actions.Action):
     validator = {voluptuous.Required("message", default=MSG): types.Jinja2}
 
     def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Result:
-        if ctxt.pull["state"] == "close":
+        if ctxt.pull["state"] == "closed":
```
sileht, line 35: thx mypy!

sileht (approved):

mergify_engine/github_events.py
```diff
@@ -274,10 +276,17 @@ async def filter_and_dispatch(
...
+    repo_name: str,
+    sha: github_types.SHAType,
+    pulls: typing.List[github_types.GitHubPullRequest],
+) -> typing.List[github_types.GitHubPullRequestNumber]:
     redis = await utils.get_aredis_for_cache()
-    cache_key = f"sha~{client.auth.owner}~{repo}~{sha}"
-    pull_number = await redis.get(cache_key)
+    cache_key = f"sha~{client.auth.owner}~{repo_name}~{sha}"
+    pull_number = typing.cast(
+        github_types.GitHubPullRequestNumber, int(await redis.get(cache_key))
```
sileht, line 288: `redis.get` can return `None`

mergify_engine/github_events.py
sileht: https://github.com/Mergifyio/mergify-engine/pull/1693

PR #1698, opened 2021-01-04
fix(github_events): use right event key name

jd:
Fixes MERGIFY-ENGINE-1ZB


sileht:
Why mypy didn't catch that ?

jd:
> Why mypy didn't catch that ?

I don't know, I'm also confused: (

PR #1700, opened 2021-01-04
test(github_events): add sample events and tests for filter_and_dispatch

jd:


sileht (approved):
I guess you pick the GitHub doc, it's better than nothing, but in my experience, there are some attributes that can be missing or present and that's not the case in real life.

jd:
Yeah, it's at least at have something to start with. We can add custom events as we need.
This would have caught the issues we had with the user permissions caching PR.

PR #1702, opened 2021-01-04
fix(github_types): add GitHubBranch types

sileht:

### DIFF
--- a/mergify_engine/actions/copy.py
+++ b/mergify_engine/actions/copy.py
@@ -25,6 +25,7 @@ from mergify_engine import check_api
 from mergify_engine import config
 from mergify_engine import context
 from mergify_engine import duplicate_pull
+from mergify_engine import github_types
 from mergify_engine import rules
 from mergify_engine.clients import http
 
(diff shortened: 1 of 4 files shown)


jd (approved):

mergify_engine/actions/copy.py
```diff
@@ -121,8 +123,16 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
...
                     branch["name"]
-                    for branch in ctxt.client.items(f"{ctxt.base_url}/branches")
-                    if any(map(lambda regex: regex.match(branch["name"]), regexes))
+                    for branch in typing.cast(
+                        typing.List[github_types.GitHubBranch],
+                        ctxt.client.items(f"{ctxt.base_url}/branches"),
+                    )
+                    if any(
+                        map(
+                            lambda regex: regex.match(typing.AnyStr(branch["name"])),
```
jd, line 132: Did you mean to cast here?

sileht: I'm a bit confused about why `str` needs to be cast as `AnyStr`, `AnyStr` seems to be `str|bytes`

sileht: Cast won't work: 
`mergify_engine/actions/copy.py:133: error: Redundant cast to typing.AnyStr?`

The original error I try to fix is: 

`mergify_engine/actions/copy.py:130: error: Argument 1 to "match" of "Pattern" has incompatible type "str"; expected "AnyStr"`

jd: Agreed you shouldn't need cast here 🤔

sileht: Looks like because it can't guess the type of the string in `re.compile`, since it uses generic it can't guess if `T` is `str` or `bytes`.  I fix it by setting type for `re.compile`

jd: Funny thing I don't see the error on my env.

PR #1710, opened 2021-01-05
test: remove double redis_stream.flushall()

jd:


sileht (commented):
The repository setup generates events we don't really care and we don't want to record.

jd:
Is doing it only at the end acceptable then?

sileht (approved):

PR #1712, opened 2021-01-05

sileht:
new commits are pushed.

Fixes  MRGFY-241


jd (commented):
Did you forget to call `create_initial_summary` on `synchronize`?
Can we add a test for this?

sileht:
Looks like I miss your commit ...

jd (dismissed):

jd (approved):

PR #1722, opened 2021-01-06
fix: replace async_run_one by asyncio.run

jd:
The former does not raise the Exception but returns it.


sileht:
I started two new patches to help to migrate to everything async and will avoid using everywhere `asyncio.run()`, because you can't run `asyncio.run()` in another `asyncio.run()`.

sileht (approved):

jd:
pep8 action never ran…

PR #1724, opened 2021-01-07
fix: don't crash on shutdown if we receive SIGTERM twice.

sileht:


jd (approved):
No associated Sentry?

PR #1725, opened 2021-01-07
feat(actions): limit rebase action to repo less than 512mo

CamClrt:
MRGFY-216


mergify_engine/actions/rebase.py
```diff
@@ -59,20 +60,39 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
...
-                    "",
-                )
-            except (
-                branch_updater.AuthenticationFailure,
-                branch_updater.BranchUpdateFailure,
-            ) as e:
-                return check_api.Result(
-                    check_api.Conclusion.FAILURE, "Branch rebase failed", str(e)
-                )
+            kind = "REBASE"
```
jd, line 63:
```suggestion
            kind = "rebase"
```

mergify_engine/actions/rebase.py
```diff
@@ -59,20 +60,39 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
...
+                    subscription.Features.LARGE_REPOSITORY
+                ):
+                    ctxt.log.warning(
+                        "repository too big and no subscription active, refusing to %s",
+                        kind,
+                        size=repo_info["size"],
+                    )
+                    raise RuntimeError(
+                        f"{kind} fail: repository is too big and no subscription is active"
+                    )
```
jd, line 78: This is going to crash the engine. You're supposed to return a `check_api.Result` with a failure state here.

mergify_engine/actions/rebase.py
```diff
@@ -59,20 +60,39 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
...
+
+            if repo_info["size"] > config.NOSUB_MAX_REPO_SIZE_KB:
+                if not ctxt.subscription.has_feature(
+                    subscription.Features.LARGE_REPOSITORY
+                ):
+                    ctxt.log.warning(
+                        "repository too big and no subscription active, refusing to %s",
+                        kind,
+                        size=repo_info["size"],
+                    )
```
jd, line 75: There should not be need for logging.
```suggestion
```

mergify_engine/actions/rebase.py
```diff
@@ -59,20 +60,39 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
...
+                    raise RuntimeError(
+                        f"{kind} fail: repository is too big and no subscription is active"
+                    )
+                ctxt.log.info("running %s on large repository", kind)
```
jd, line 79:
```suggestion
```

mergify_engine/actions/rebase.py
```diff
@@ -59,20 +60,39 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
...
+                        "",
+                    )
+                except (
+                    branch_updater.AuthenticationFailure,
+                    branch_updater.BranchUpdateFailure,
+                ) as e:
+                    return check_api.Result(
+                        check_api.Conclusion.FAILURE, "Branch rebase failed", str(e)
+                    )
         else:
```
jd, line 96: Hmmm, did you try to execute the program in your head first?
That `else` is now triggered by the condition on L67 being `False`, which means that when `repo_info["size"] <= config.NOSUB_MAX_REPO_SIZE_KB` this is going to return "Branch already up to date" without doing anything…

jd (changes requested):

GuillaumeOj:
> MRGFY-216

You should type `Fixes MRGFY-216` in your comment for linking with Linear issue

GuillaumeOj (commented):
LGTM. But is there any test for this action ?
Poke @sileht.

jd (dismissed):
Chatted with @CamClrt about this, need some rework :)

mergify_engine/actions/rebase.py
```diff
@@ -59,6 +60,19 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
             if output:
                 return output
 
+            repo_full_name = ctxt.pull["base"]["repo"]["full_name"]
+            repo_info = ctxt.client.item(f"/repos/{repo_full_name}")
```
jd, line 64: Not a big deal, but I think this would be slightly more accurate:
```suggestion
            repo_info = ctxt.client.item(f"{ctxt.pull["base"]["repo"]["url"]")
```

jd (commented):
@sileht There's a new call to GitHub REST API on L64 but no test fails and nothing has been recorded.
Am I crazy to think that even the base test for the rebase action might be broken?

mergify_engine/actions/rebase.py
```diff
@@ -59,6 +60,18 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
             if output:
                 return output
 
+            repo_info = ctxt.client.item(ctxt.pull["base"]["repo"]["url"])
+
+            if repo_info["size"] > config.NOSUB_MAX_REPO_SIZE_KB:
+                if not ctxt.subscription.has_feature(
+                    subscription.Features.LARGE_REPOSITORY
+                ):
```
jd, line 68:
```suggestion
            if repo_info["size"] > config.NOSUB_MAX_REPO_SIZE_KB and not ctxt.subscription.has_feature(subscription.Features.LARGE_REPOSITORY):
```
no need for 2 if there

jd (changes requested):
We just need a test otherwise that looks good to me.

CamClrt:
@Mergifyio refresh

mergify_engine/tests/functional/actions/test_rebase.py
```diff
@@ -0,0 +1,58 @@
...
+
+        self.setup_repo(
+            yaml.dump(rules),
+            test_branches=[
+                ahead_branch,
+            ],
+        )
+
+        p, commits = self.create_pr()
+        tmp_sha = commits[-1].sha
```
jd, line 43:
```suggestion
        pr_initial_sha = commits[-1].sha
```

jd (changes requested):

mergify_engine/tests/functional/actions/test_rebase.py
```diff
@@ -0,0 +1,58 @@
...
+
```
jd, line 44: is this true?
```python
assert tmp_sha == p.head.sha
```
?
(I think it is and it should but just making sure)

CamClrt: yes, it's True

jd:
Why did you close this?

PR #1727, opened 2021-01-07
feat(action): add bot_account for comments

GuillaumeOj:


mergify_engine/actions/comment.py
```diff
@@ -39,9 +56,22 @@ def run(self, ctxt: context.Context, rule: rules.EvaluatedRule) -> check_api.Res
                 str(rmf),
             )
 
+        bot_account = self.config["bot_account"]
+        if bot_account:
```
jd, line 60:
```suggestion
        if bot_account := self.config["bot_account"]:
```
😉

mergify_engine/tests/functional/actions/test_comment.py
```diff
@@ -20,6 +20,8 @@
 
 
 class TestCommentAction(base.FunctionalTestBase):
+    SUBSCRIPTION_ACTIVE = True
```
jd, line 23: Can you only set this in a different test class so we're able to test both scenario, with and without sub enabled?

```python
class TestCommentActionWithSub(base.FunctionalTestBase):
    SUBSCRIPTION_ACTIVE = True

     def test_comment_with_oauth_token(self):
         [...]
```

and add a test with no sub enabled 😄

jd (dismissed):
👍 
Perfect, I just wish we could add one more test for failure if possible.

jd (approved):

PR #1728, opened 2021-01-07
chore(tests): isolates asyncio main loop between tests.

sileht:


mergify_engine/tests/conftest.py
```diff
@@ -19,3 +20,9 @@ def logger_checker(request, caplog):
             if rec.levelname in ("CRITICAL", "ERROR")
         ]
         assert [] == messages
+
+
+@pytest.fixture(autouse=True)
+def setup_new_event_loop(request):
```
jd, line 26: Do you have to pass request here if it's unused?

mergify_engine/tests/functional/actions/test_post_check.py
```diff
@@ -128,8 +128,7 @@ def test_checks_feature_disabled(self):
             if self.SUBSCRIPTION_ACTIVE
             else frozenset(),
         )
-        loop = asyncio.get_event_loop()
-        loop.run_until_complete(self.subscription.save_subscription_to_cache())
+        utils.async_run(self.subscription.save_subscription_to_cache())
```
jd, line 131:
```suggestion
        asyncio.run(self.subscription.save_subscription_to_cache())
```
should work since there's no `gather()` involved.

jd (dismissed):
I think `utils.async_run` can be replaced by `asyncio.run` when you're calling only one function.

jd (approved):

jd:
@mergifyio update

jd:
@Mergifyio update

PR #1731, opened 2021-01-08
chore: remove unused uvloop dependency

jd:


jd:
Used by uvicorn actually.

PR #1732, opened 2021-01-08
feat: remove uvloop

sileht:


jd (dismissed):
Remove the hack in tox.ini at least?

jd (approved):

PR #1752, opened 2021-01-09
fix(github_types): set correct type for redis stream returns

sileht:


mergify_engine/worker.py
```diff
@@ -149,7 +152,7 @@ async def push(
         gh_pull=pull_number,
         event_type=event_type,
     )
-    return (ret[0], payload)
+    return (T_MessageID(ret[0]), payload)
```
jd, line 155: I'd rather do:
```suggestion
    return (typing.cast(T_MessageID, ret[0]), payload)
```
no?
I'm surprised this is not making mypy complain.

sileht: I guess it works because aredis return Any, not sure

sileht: According the doc, when you use `NewType` it's equivalent at mypy time as :

```
class T_MessageID(str):
    def __init__(self, x: str) -> None:
         pass
```
and at runtime to:
```
def T_MessageID(self, x: str) -> str:
    return x
```

In our case redis return `Any`, so it can't check anything and just trust us.

sileht: I just push a third syntax to do the same for thing returning Any ;)

jd (approved):

PR #1753, opened 2021-01-09
feat: add owner id to stream name

sileht:


mergify_engine/worker.py
```diff
@@ -111,13 +111,15 @@ class T_PayloadEvent(typing.TypedDict):
 )
 async def push(
     redis: aredis.StrictRedis,
-    owner: str,
-    repo: str,
+    # NOTE(sileht): remove Optional when all stream have the new name format
```
jd, line 114: there's no `Optional` here

jd (approved):

PR #1754, opened 2021-01-09
chore: remove unused github client

sileht:


mergify_engine/github_events.py
```diff
@@ -338,13 +337,13 @@ async def filter_and_dispatch(
 
 
 async def _get_github_pulls_from_sha(
-    client: github.AsyncGithubInstallationClient,
-    repo_name: str,
+    owner_login: github_types.GitHubLogin,
```
jd, line 340: Would be simpler to pass a GitHubRepository entirely.

jd (changes requested):

mergify_engine/github_events.py
sileht: We don't have one at this point.

jd: Fair enough.

jd (approved):

sileht:
@Mergifyio update

PR #1757, opened 2021-01-09
test: checkout correct right ref

sileht:
Workaround for https://github.com/actions/checkout/issues/299


jd (approved):
Do you want to test it somehow?

PR #1774, opened 2021-01-11
chore(github_types): use GitHubAccountIdType instead of int

sileht:


mergify_engine/clients/github.py
```diff
@@ -117,7 +121,9 @@ def auth_flow(self, request):
                     "GET", f"{config.GITHUB_API_URL}/users/{self.owner}"
                 )
                 http.raise_for_status(user_response)
-                self.owner_id = user_response.json()["id"]
+                self.owner_id: github_types.GitHubAccountIdType = user_response.json()[
```
jd, line 124: I think it's a better practice to cast the `user_response`, not just the field.

jd (dismissed):

jd (approved):

PR #1776, opened 2021-01-11
feat: only get subscription once per stream

sileht:


jd (approved):

jd (changes requested):
going with the first one

sileht:
@Mergifyio rebase

jd (approved):

PR #1779, opened 2021-01-11
refactor: introduce context.Installation

sileht:


sileht:
GitHub and conflict handling ...

mergify_engine/context.py
```diff
@@ -53,6 +53,15 @@ class PullRequestAttributeError(AttributeError):
     name: str
 
 
+@dataclasses.dataclass
+class Installation:
+    stream_name: str
```
jd, line 58: use a type for this?

jd (changes requested):

jd (approved):

PR #1781, opened 2021-01-11
fix(worker): logging should use owner_login

sileht:
installation may not yet exists at this point


jd (approved):

sileht:
I wonder if there is a mypy/flake8 check to avoid this kind of mistake. (I didn't find one at first glance)

PR #1729, opened 2021-01-07
feat: add queue action

sileht:

PR #1785, opened 2021-01-12
refactor: introduce Repository context

sileht:


mergify_engine/context.py
```diff
@@ -63,25 +63,74 @@ class Installation:
     subscription: subscription.Subscription
     client: github.GithubInstallationClient
 
+    repositories: "typing.Dict[github_types.GitHubRepositoryName, Repository]" = (
+        dataclasses.field(default_factory=dict)
+    )
```
jd, line 68: Do you really need to store them?

mergify_engine/context.py
```diff
@@ -63,25 +63,74 @@ class Installation:
...
+    installation: Installation
+    name: str
+    pull_contexts: "typing.Dict[github_types.GitHubPullRequestNumber, Context]" = (
+        dataclasses.field(default_factory=dict)
+    )
+
+    @property
+    def base_url(self) -> str:
+        """The URL prefix to make GitHub request."""
+        return f"/repos/{self.installation.owner_login}/{self.name}"
```
jd, line 94: This is actually `github_types.GitHubRepository['url']` so I'm wondering if we really need this.
If it's possible to use the GitHubRepository URL that makes it easier to support different URL scheme etc :)

mergify_engine/context.py
```diff
@@ -63,25 +63,74 @@ class Installation:
...
+        return self.repositories[name]
+
+
+@dataclasses.dataclass
+class Repository(object):
+    installation: Installation
+    name: str
+    pull_contexts: "typing.Dict[github_types.GitHubPullRequestNumber, Context]" = (
+        dataclasses.field(default_factory=dict)
+    )
```
jd, line 89: Do we need to store this either?

mergify_engine/context.py
sileht: This is planned to be replace this Context construction here: https://github.com/Mergifyio/mergify-engine/pull/1729/files#diff-e709bf7f7a0ceb20e0cfafd3b8ab1e163a0c04e84262742607dc6828cf868770R258

mergify_engine/context.py
sileht: This is planned to be replace this Context construction here: https://github.com/Mergifyio/mergify-engine/pull/1729/files#diff-e709bf7f7a0ceb20e0cfafd3b8ab1e163a0c04e84262742607dc6828cf868770R258

mergify_engine/context.py
sileht: This url is currently not forwarded from the webhook to the worker.

sileht:
@jd ?

jd (approved):

PR #1786, opened 2021-01-12
chore(github_types): GitHubPullRequestNumber must be int

sileht:
Using typing.cast(GitHubPullRequestNumber, "123") is valid for mypy as
it just trust us.

GitHubPullRequestNumber must be int, so we can use
GitHubPullRequestNumber("123") and got the expected:

Argument 1 to "GitHubPullRequestNumber" has incompatible type "str"; expected "int"

GitHubPullRequestNumber(GitHubIssueNumber(123)) will still works as
there are both int

### DIFF
--- a/mergify_engine/github_types.py
+++ b/mergify_engine/github_types.py
@@ -137,8 +137,8 @@ GitHubPullRequestMergeableState = typing.Literal[
     "has_hooks",
 ]
 
-GitHubPullRequestId = typing.NewType("GitHubPullRequestId", GitHubIssueId)
-GitHubPullRequestNumber = typing.NewType("GitHubPullRequestNumber", GitHubIssueNumber)
+GitHubPullRequestId = typing.NewType("GitHubPullRequestId", int)
+GitHubPullRequestNumber = typing.NewType("GitHubPullRequestNumber", int)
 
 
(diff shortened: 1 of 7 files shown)


mergify_engine/github_types.py
```diff
@@ -137,8 +137,8 @@ class GitHubIssue(GitHubIssueOrPullRequest):
     "has_hooks",
 ]
 
-GitHubPullRequestId = typing.NewType("GitHubPullRequestId", GitHubIssueId)
-GitHubPullRequestNumber = typing.NewType("GitHubPullRequestNumber", GitHubIssueNumber)
+GitHubPullRequestId = typing.NewType("GitHubPullRequestId", int)
+GitHubPullRequestNumber = typing.NewType("GitHubPullRequestNumber", int)
```
jd, line 141: The problem I think is that if you have a method like
```python
def retrieve_comments(issue_id: GitHubIssueId):
```
and that works on both issues and PRs, then you can't pass a `GitHubPullRequestId` to it because they are different.

sileht: This method should take `typing.Union[GitHubIssueId, GitHubPullRequestId]` then.

jd (approved):

PR #1788, opened 2021-01-12
refactor: drop our custom PullRequestJinjaContext

sileht:
It parses the template twice, but that worses it to make the template
rendering async later.


jd (dismissed):
Did you mean to also remove the class?

jd (approved):

PR #1789, opened 2021-01-12
refactor: use asyncio.run once by engine run

sileht:
With this change asyncio.run() is used into two places only,
at the beginning of engine thread and the simulator thread.

### DIFF
--- a/mergify_engine/rules/types.py
+++ b/mergify_engine/rules/types.py
@@ -39,7 +39,7 @@ class LineColumnPath(object):
 
 class DummyContext(context.Context):
 
-    # This is only used to check Jinja2 syntax validity
+    # This is only used to check Jinja2 syntax validity and must be sync
     @staticmethod
     def _get_consolidated_data(key):
         if key in context.PullRequest.ATTRIBUTES:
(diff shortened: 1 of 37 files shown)


sileht:
5 tests remaining :smile:

mergify_engine/rules/types.py
```diff
@@ -54,7 +54,28 @@ def _ensure_complete():
         pass
 
 
-_DUMMY_PR = context.PullRequest(
+class DummaryPullRequest(context.PullRequest):
```
jd, line 57:
```suggestion
class DummyPullRequest(context.PullRequest):
```

mergify_engine/rules/types.py
```diff
@@ -54,7 +54,28 @@ def _ensure_complete():
...
+            infos = {}
+            for k in used_variables:
+                if extra_variables and k in extra_variables:
+                    infos[k] = extra_variables[k]
+                else:
+                    infos[k] = getattr(self, k)
+            return env.from_string(template).render(**infos)
+
+
+_DUMMY_PR = DummaryPullRequest(
```
jd, line 78:
```suggestion
_DUMMY_PR = DummyPullRequest(
```

mergify_engine/rules/types.py
```diff
@@ -54,7 +54,28 @@ def _ensure_complete():
...
+        )
+        with self._template_exceptions_mapping():
+            used_variables = jinja2.meta.find_undeclared_variables(env.parse(template))
+            infos = {}
+            for k in used_variables:
+                if extra_variables and k in extra_variables:
+                    infos[k] = extra_variables[k]
+                else:
+                    infos[k] = getattr(self, k)
+            return env.from_string(template).render(**infos)
```
jd, line 75: Is this duplicated from `context`?

jd (dismissed):

mergify_engine/rules/types.py
sileht: One is sync the other is async.

jd (approved):

PR #1800, opened 2021-01-15
chore: create one redis client per stream

sileht:
Fixes MRGFY-267

### DIFF
--- a/mergify_engine/actions/dismiss_reviews.py
+++ b/mergify_engine/actions/dismiss_reviews.py
@@ -20,7 +20,6 @@ from mergify_engine import actions
 from mergify_engine import check_api
 from mergify_engine import context
 from mergify_engine import rules
-from mergify_engine import utils
 from mergify_engine.clients import http
 from mergify_engine.rules import types
 
@@ -51,13 +50,12 @@ class DismissReviewsAction(actions.Action):
             # his access_token instead of the Mergify installation token.
             # As workaround we track in redis merge commit id
             # This is only true for method="rebase"
-            async with utils.aredis_for_cache() as redis:
-                if await redis.get(f"branch-update-{ctxt.pull['head']['sha']}"):
-                    return check_api.Result(
-                        check_api.Conclusion.SUCCESS,
-                        "Updated by Mergify, ignoring",
-                        "",
-                    )
+            if await ctxt.redis.get(f"branch-update-{ctxt.pull['head']['sha']}"):
+                return check_api.Result(
+                    check_api.Conclusion.SUCCESS,
+                    "Updated by Mergify, ignoring",
+                    "",
+                )
 
             try:
                 message = await ctxt.pull_request.render_template(
--- a/mergify_engine/actions/merge_base.py
+++ b/mergify_engine/actions/merge_base.py
@@ -14,6 +14,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 import abc
+import asyncio
 import enum
 import itertools
 import re
@@ -26,6 +27,7 @@ from mergify_engine import branch_updater
 from mergify_engine import check_api
 from mergify_engine import config
 from mergify_engine import context
+from mergify_engine import github_types
 from mergify_engine import json as mergify_json
 from mergify_engine import queue
 from mergify_engine import subscription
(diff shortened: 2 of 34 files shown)


mergify_engine/actions/dismiss_reviews.py
```diff
@@ -51,13 +50,12 @@ async def run(
...
             # As workaround we track in redis merge commit id
             # This is only true for method="rebase"
-            async with utils.aredis_for_cache() as redis:
-                if await redis.get(f"branch-update-{ctxt.pull['head']['sha']}"):
-                    return check_api.Result(
-                        check_api.Conclusion.SUCCESS,
-                        "Updated by Mergify, ignoring",
-                        "",
-                    )
+            if await ctxt.redis.get("branch-update-%s" % ctxt.pull["head"]["sha"]):
```
jd, line 53: I liked the f string :p

mergify_engine/actions/dismiss_reviews.py
```diff
@@ -51,13 +50,12 @@ async def run(
...
+                return check_api.Result(
+                    check_api.Conclusion.SUCCESS,
+                    "Rebased/Updated by us, nothing to do",
```
jd, line 56: I liked the old msg

mergify_engine/actions/merge_base.py
```diff
@@ -607,8 +611,9 @@ def get_queue_summary(self, ctxt: context.Context, q: queue.Queue) -> str:
         priorities_configured = False
 
         summary = "\n\nThe following pull requests are queued:"
+        pulls_priorities = {p: (await q.get_config(p))["priority"] for p in pulls}
```
jd, line 614: something like this should be better to do multiple calls at once:
```suggestion
        async def _get_config(p):
             return p, (await q.get_config(p))["priority"]
        pulls_priorities = dict(await asyncio.gather(*(_get_config(p) for p in pulls)))
```

mergify_engine/context.py
```diff
@@ -117,20 +117,20 @@ class Context(object):
     SUMMARY_NAME = "Summary"
     USER_PERMISSION_EXPIRATION = 3600  # 1 hour
 
+    @property
+    def redis(self) -> aredis.StrictRedis:
+        return self.repository.installation.redis
+
     @property
     def subscription(self) -> subscription.Subscription:
-        # TODO(sileht): remove me when context split if done
```
jd, line 122: why?

mergify_engine/tests/functional/actions/test_post_check.py
```diff
@@ -116,6 +117,7 @@ def test_checks_custom(self):
 class TestPostCheckActionNoSub(base.FunctionalTestBase):
     def test_checks_feature_disabled(self):
         self.subscription = subscription.Subscription(
+            asyncio.run(utils.create_aredis_for_cache(max_idle_time=0)),
```
jd, line 120: Does that make even sense?

jd (changes requested):

mergify_engine/tests/functional/actions/test_post_check.py
sileht: It's not really logic but it works because of the aredis implementation.

All tests should be async, but pytest-asyncio does not work for tests using unittest.TestCase :( 
I don't know yet how to fix that. Maybe we need to rewrite the setUp and all helpers with pytest, but that a lot of works (and a lot of tox -e record).

sileht: I just found python3.8 have introduced unittest.IsolatedAsyncioTestCase, I will make a separate PR for cleaning this.

mergify_engine/utils.py
```diff
@@ -26,61 +26,51 @@
...
-import redis
 
 from mergify_engine import config
 
 
 _PROCESS_IDENTIFIER = os.environ.get("DYNO") or socket.gethostname()
 
+# NOTE(sileht): I wonder with mypy thing aredis.StrictRedis is Any...
+RedisCache = typing.NewType("RedisCache", aredis.StrictRedis)  # type: ignore
+RedisStream = typing.NewType("RedisStream", aredis.StrictRedis)  # type: ignore
```
jd, line 37: smart!

jd (approved):

PR #1803, opened 2021-01-17
chore: use asyncio subprocess

sileht:


mergify_engine/utils.py
```diff
@@ -108,54 +110,66 @@ def compute_hmac(data):
...
-        kwargs.setdefault("stderr", subprocess.STDOUT)
+        kwargs.setdefault("stdout", asyncio.subprocess.PIPE)
+        kwargs.setdefault("stderr", asyncio.subprocess.STDOUT)
         try:
-            return subprocess.run(["git"] + list(args), **kwargs)
+            p = await asyncio.create_subprocess_exec(
+                "git", ["git"] + list(args), **kwargs
+            )
+            # Worker timeout at 5 minutes, so ensure subprocess return before
+            await asyncio.wait_for(p.wait(), 4 * 60 + 30)
```
jd, line 134: can we use a variable rather than this time?
just in case we change the 5 minutes and to make this more "readable"

mergify_engine/utils.py
```diff
@@ -108,54 +110,66 @@ def compute_hmac(data):
...
+            if p.returncode:
+                raise subprocess.CalledProcessError(
```
jd, line 136: should we stop using `subprocess` entirely here?

mergify_engine/utils.py
sileht: done.

mergify_engine/utils.py
sileht: I reviewed the API of gitter (exception and returned value), move it into a separate file, added tests.

mergify_engine/gitter.py
```diff
@@ -0,0 +1,96 @@
...
+    output: str
+
+
+@dataclasses.dataclass
+class Gitter(object):
+    logger: logging.LoggerAdapter
+    tmp: str = dataclasses.field(init=False)
+
+    # Worker timeout at 5 minutes, so ensure subprocess return before
+    GIT_COMMAND_TIMEOUT: int = dataclasses.field(init=False, default=4 * 60 + 30)
```
jd, line 39: can we use the 5 minutes your referring to in the `default`? I guess it comes from a variable.

mergify_engine/gitter.py
```diff
@@ -0,0 +1,96 @@
...
+                cwd=self.tmp,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.STDOUT,
+                stdin=None if input is None else asyncio.subprocess.PIPE,
+            )
+
+            stdout, _ = await asyncio.wait_for(
+                p.communicate(input=None if input is None else input.encode("utf8")),
+                self.GIT_COMMAND_TIMEOUT,
+            )
```
jd, line 61: What happens if that raises `TimeoutError`?

mergify_engine/gitter.py
sileht: I see

mergify_engine/gitter.py
sileht: Before it was raising `TimeoutExpired` and it was not handled, and worker just retry it later. It should have been logged in sentry, if that happen. I just keep the same behavior.

mergify_engine/gitter.py
sileht: I don't find a trace about that this comment mean, I guess it comes from the time we having celery.

jd: ok we might need to fix that then…

jd (approved):

PR #1809, opened 2021-01-18
refactor: pass Repository context to Queue()

sileht:


mergify_engine/queue.py
```diff
@@ -37,42 +37,34 @@ class QueueConfig(typing.TypedDict):
...
-    owner: str
-    repo_id: int
-    repo: str
+    repository: context.Repository
     ref: str
 
     log: logging.LoggerAdapter = dataclasses.field(init=False)
 
     def __post_init__(self):
         self.log = daiquiri.getLogger(
```
jd, line 46: It'd be great to have a method in utils that would give a logger like this:
```python
def get_logger(name: str, gh_owner: github_types.GitHubLogin, gh_repo: github_types.GitHubRepositoryName, …):
   return daiquiri.getLogger(name, gh_owner=gh_owner, gh_repo=gh_repo, …)
```

So we're sure to always have logger with the same format and the type being checked.

mergify_engine/queue.py
```diff
@@ -37,42 +37,34 @@ class QueueConfig(typing.TypedDict):
...
     def from_context(cls, ctxt: context.Context) -> "Queue":
-        return cls(
-            ctxt.redis,
-            ctxt.pull["base"]["repo"]["owner"]["id"],
-            ctxt.pull["base"]["repo"]["owner"]["login"],
-            ctxt.pull["base"]["repo"]["id"],
-            ctxt.pull["base"]["repo"]["name"],
-            ctxt.pull["base"]["ref"],
-        )
+        return cls(ctxt.repository, ctxt.pull["base"]["ref"])
```
jd, line 55: I'd use a `Repository` as arg here, it looks overkill to require an entire `Context`.

jd (changes requested):

mergify_engine/queue.py
sileht: The method is called `from_context` :p

mergify_engine/context.py
```diff
@@ -93,6 +95,19 @@ def base_url(self) -> str:
         """The URL prefix to make GitHub request."""
         return f"/repos/{self.installation.owner_login}/{self.name}"
 
+    @property
+    def id(self) -> github_types.GitHubRepositoryIdType:
+        # TODO(sileht): Ensure all events push to worker have repo id set, to be able to
+        # remove this
+        if self._id is None:
+            ctxt = first.first((c for _, c in self.pull_contexts.items()))
```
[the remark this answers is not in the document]
sileht: Fixed

mergify_engine/queue.py
sileht: Agreed, maybe another time, because I have other issues in mind with all these loggers.

mergify_engine/queue.py
jd: I guess you could rename it?

jd (approved):
I think it's fine this way but the `from_context` one is still ☹️

PR #1813, opened 2021-01-19
fix: add/remove pull and config to/from queue in a transaction

sileht:
This ensures we always have the config when a pull is in queue.
And we don't have config left behind with a pull is removed from the
queue.

Fixes MRGFY-71

### DIFF
--- a/mergify_engine/queue.py
+++ b/mergify_engine/queue.py
@@ -79,8 +79,10 @@ class Queue:
             self._config_redis_queue_key(pull_number)
         )
         if config_str is None:
-            # TODO(sileht): Everything about queue should be done in redis transaction
-            # e.g.: add/update/get/del of a pull in queue
+            self.log.error(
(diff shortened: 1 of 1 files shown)


jd (approved):

mergify_engine/queue.py
```diff
@@ -100,15 +102,17 @@ async def get_config(
     async def add_pull(self, ctxt: context.Context, config: QueueConfig) -> None:
         await self._remove_pull_from_other_queues(ctxt)
 
-        await self.repository.installation.redis.set(
-            self._config_redis_queue_key(ctxt.pull["number"]),
-            json.dumps(config),
-        )
+        with self.repository.installation.redis.pipeline() as pipeline:
```
jd, line 105: async with?

sileht: `async with wait` !

PR #1854, opened 2021-01-19
feat: cache team members

sileht:
Fixes MRGFY-269

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -17,6 +17,7 @@ import base64
 import contextlib
 import dataclasses
 import itertools
+import json
 import logging
 import typing
 from urllib import parse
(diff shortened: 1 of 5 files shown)


mergify_engine/context.py
```diff
@@ -68,6 +68,7 @@ class Installation:
     repositories: "typing.Dict[github_types.GitHubRepositoryName, Repository]" = (
         dataclasses.field(default_factory=dict)
     )
+    _cache: typing.Dict[str, typing.Any] = dataclasses.field(default_factory=dict)
```
jd, line 71: My original idea was more about using Redis than just an internal cache.
We have access to the event to invalidate the cache so it should be ok to cache for some time.

jd: Also you could used:

```python
class InstallationCacheDictT(typing.TypedDict):
    team_members: typing.Dict[typing.Tuple[github_types.GitHubLogin, str], typing.List[typing.GitHubLogin]]
```

jd (changes requested):

mergify_engine/context.py
```diff
@@ -82,6 +82,65 @@ def get_repository(self, name: github_types.GitHubRepositoryName) -> "Repository
...
+        if not members:
+            members = [
+                github_types.GitHubLogin(member["login"])
+                async for member in self.client.items(
+                    f"/orgs/{self.owner_login}/teams/{team_slug}/members"
+                )
+            ]
+            pipe = await self.redis.pipeline()
+            await pipe.rpush(key, *members)
+            await pipe.expire(key, self.TEAM_MEMBERS_EXPIRATION)
```
jd, line 140: I feel like this is going to crash if the team is empty and has 0 member.

jd (changes requested):

mergify_engine/context.py
sileht: I have added a test to covert that

jd: push it!

mergify_engine/context.py
```diff
@@ -82,6 +83,63 @@ def get_repository(self, name: github_types.GitHubRepositoryName) -> "Repository
...
+            members = [
+                github_types.GitHubLogin(member["login"])
+                async for member in self.client.items(
+                    f"/orgs/{self.owner_login}/teams/{team_slug}/members"
+                )
+            ]
+            # TODO(sileht): move to msgpack when we remove redis-cache connection
+            # from decode_responses=True (eg: MRGFY-285)
+            pipe = await self.redis.pipeline()
+            await pipe.hset(key, team_slug, json.dumps(members))
```
jd, line 134: do you have to reencode here since you could get GitHub response directly?

jd (commented):
just one comment

mergify_engine/context.py
sileht: What do you mean by reencode ?   if  you mean str.encode() the json dumps, it's not needed because the redis client is configured to do it.
also members is not a GitHub response but just list of GitHubLogin (str).

jd: I meant that GitHub replies with a JSON that we could store right away to Redis.
But I see your points, we're only storing logins.

jd (approved):

PR #1863, opened 2021-01-20
feat: add typing for YAML

GuillaumeOj:


jd (dismissed):
I'd rather use `refactor` than `feat` here 😉

jd (approved):

PR #1873, opened 2021-01-22
ci: add bandit

sileht:
Fixes MRGFY-275


mergify_engine/actions/label.py
```diff
@@ -1,4 +1,4 @@
-# -*- encoding: utf-8 -*-
+# -*- encoding: utf-8 -*:
```
jd, line 1:
```suggestion
# -*- encoding: utf-8 -*-
```

jd (dismissed):
Add Linear issue?

jd (approved):

Issue #1872, opened 2021-01-22
'update' action seems to be not consistent


Borda:
We have defined `update` action and it seems to work with some PRs but not for all, could it be like to presence of config in the target branch?
So we have updated our `master` branch with the actual mergify config which I was debugging in mergify UI

this PR is going to `master` and action was executed: https://github.com/PyTorchLightning/pytorch-lightning/pull/5607
![Screenshot 2021-01-22 at 11 24 20](https://user-images.githubusercontent.com/6035284/105479427-b9692600-5ca4-11eb-8194-9fece48cfe40.png)

this PR is going to `release/1.2-dev` and action was NOT executed: https://github.com/PyTorchLightning/pytorch-lightning/pull/5596
![Screenshot 2021-01-22 at 11 24 42](https://user-images.githubusercontent.com/6035284/105479541-ddc50280-5ca4-11eb-8422-7dae56dd8ceb.png)

When I check the conditions in UI in both cases it shall be triggered...

sileht:
On PyTorchLightning/pytorch-lightning#5596 rules said to update the PR only if the PR conflict.

You have fixed the configuration here:
https://github.com/PyTorchLightning/pytorch-lightning/commit/015cb2f33eafdf3353e934f2e1872f507f4bdd96e

But Mergify didn't re-run automatically since nothing change on this PR.

You can use `@Mergifyio refresh` to re-trigger Mergify with the updated configuration.

Borda:
OK, you say that mergify is event-driven, so why in UI is a clock which stated when the next update comes?
like this message in the UI: `Nothing planned to be merged. Next refresh at 1:51 PM`
So is there another way to propagate updates on mergify config? It can be quite painful after each cofing update if you need to check all open PRs and manually ping the bot... imaging large repos with hundreds of PRs

jd:
You're looking at `Queues` in the dashboard, which as its names states, shows merge queues. 😉 

Re-evaluating all open PR in a large repo would exhaust GitHub API rate limit anyway, so there's little chance we could do that 😢

PR #1877, opened 2021-01-23
docs: add example with update

Borda:
Adding example for update example it was not trivial to get it done, see my question #1872

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -410,7 +410,7 @@ little message.
                   We're delighted to have you onboard <3
 
 🤜 Request for Action
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~
 
 If any event that requires the author of the pull request to edit its pull
 request happen, you could write a rule that says something about it.
@@ -527,6 +527,26 @@ request.
               Conditions:
 
               {{ check_conditions }}
+
+
+🐍 Updating PR for linear history
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As GitHub supports linear history in pull request settings, it is very handy to use a rule
+to keep your pull requests up-to-date. As you do not want to trigger your CI too often by always
+re-running it on every pull request — especially when there is still work in progress — you can
+limit this action to labeled pull requests.
+
+.. code-block:: yaml
+
+    pull_request_rules:
+      - name: automatic update for PR marked as “Ready-to-Go“
+        conditions:
+          - -conflict # skip PRs with conflicts
+          - -draft # filter-out GH draft PRs
+          - label="Ready-to-Go"
+        actions:
+          update: {}
 
 
 .. include:: examples/bots.rst
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -118,6 +118,11 @@ Here's the relevant configuration file content:
   ``merge`` action with a ``method`` parameter containing the merge method to
   use.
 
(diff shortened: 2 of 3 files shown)


Borda:
@sileht @jd mind have look? ^^

docs/source/examples.rst
```diff
@@ -529,5 +529,28 @@ request.
...
+          - -draft # filter-out GH draft PRs
+          - label="Ready-to-Go"
+        actions:
+          update: {}
+
+.. note::
+
+    Mergify is event driven bot so even you add this cool rule, it wound be executed till each PR is updated
+     or the bot is pinged in a comment as `@Mergifyio refresh`.
+
```
jd, line 554: We should write that somewhere else in a more "general" place.
```suggestion

```

docs/source/examples.rst
```diff
@@ -529,5 +529,28 @@ request.
               {{ check_conditions }}
 
 
+🐍 Updating PR for linear history
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As GitHub recently introduced linear history in PR setting it become very handy to use an automatic tool
+ to keep your PRs up-to-date. On the other hand you do not want to overkill your CI with building all PRs
+ always especially when there is still work in progress, so you can limit this action just to ready labeled PRs.
```
jd, line 537:
```suggestion
As GitHub supports linear history in pull request settings, it is very handy to use a rule
to keep your pull requests up-to-date. As you do not want to trigger your CI too often by always re-running it on every pull request — especially when there is still work in progress — you can limit this action to labeled pull requests.
```

jd (changes requested):
Thanks a lot!
Suggested some rewording, LGTM otherwise.
If you can just amend your commit that'd be good.

docs/source/examples.rst
Borda: so at the beginning or where bc till I knew it caused me lost of troubles to understands what is happening

Borda:
@jd added your suggestion and moved the general note to the starter page

docs/source/examples.rst
jd: I'll probably add this to https://docs.mergify.io/getting-started.html

docs/source/getting-started.rst
```diff
@@ -118,6 +118,11 @@ Here's the relevant configuration file content:
   ``merge`` action with a ``method`` parameter containing the merge method to
   use.
 
+.. note::
+
+    Mergify is event driven bot so even you add this cool rule, it wound be executed till each PR is updated
+     or the bot is pinged in a comment as `@Mergifyio refresh`.
```
jd, line 124:
```suggestion
    Mergify is event-driven: rules are only executed when pull requests are updated
    or when the bot is manually triggered in a comment with `@Mergifyio refresh` (see :ref:`Commands`).
```

jd (changes requested):
sounds great, reworded

jd:
You'll need to squash your commit in a single one :)

Borda:
> You'll need to squash your commit in a single one :)

ok, I ll do if but why you do not make squash merge here in GH UI? :P

jd:
Haha, we usually use rebase method around here 😄 
You need to update your commit message to match the PR title (which is correct for semantic).

What we need is a `@mergifyio squash` command!

Borda:
> What we need is a `@mergifyio squash` command!

that would be great as I think that usage is better and also easier than pushing/explaining to contributors how to make squash in cmd 🐰

jd (approved):

PR #1906, opened 2021-01-25
chore: add type to Context._cache

sileht:


mergify_engine/check_api.py
```diff
@@ -61,7 +61,7 @@ async def get_checks_for_ref(
         kwargs = {}
     else:
         kwargs = {"check_name": check_name}
-    checks = [
+    checks: typing.List[github_types.GitHubCheckRun] = [
```
jd, line 64: can we rather cast the `client.items` call?

mergify_engine/context.py
```diff
@@ -594,25 +595,29 @@ async def pull_engine_check_runs(self) -> typing.List[github_types.GitHubCheckRu
...
         ]
 
+    @property
+    async def pull_statuses(self) -> typing.List[github_types.GitHubStatus]:
+        if "pull_statuses" in self._cache:
+            return self._cache["pull_statuses"]
+
+        statuses: typing.List[github_types.GitHubStatus] = [
+            s
+            async for s in self.client.items(
```
jd, line 605: can we cast this?

mergify_engine/context.py
```diff
@@ -769,11 +771,9 @@ def __str__(self) -> str:
...
         if "reviews" in self._cache:
-            return typing.cast(
-                typing.List[github_types.GitHubReview], self._cache["reviews"]
-            )
+            return self._cache["reviews"]
 
-        reviews = [
+        reviews: typing.List[github_types.GitHubReview] = [
             review
             async for review in self.client.items(
```
jd, line 778: can we cast this?

mergify_engine/context.py
```diff
@@ -785,26 +785,21 @@ async def reviews(self) -> typing.List[github_types.GitHubReview]:
     @property
     async def commits(self) -> typing.List[github_types.GitHubBranchCommit]:
         if "commits" in self._cache:
-            return typing.cast(
-                typing.List[github_types.GitHubBranchCommit], self._cache["commits"]
+            return self._cache["commits"]
+        commits: typing.List[github_types.GitHubBranchCommit] = [
+            commit
+            async for commit in self.client.items(
```
jd, line 791: can we cast this?

mergify_engine/context.py
```diff
@@ -785,26 +785,21 @@ async def reviews(self) -> typing.List[github_types.GitHubReview]:
...
 
     @property
-    async def files(self):
+    async def files(self) -> typing.List[github_types.GitHubFile]:
         if "files" in self._cache:
             return self._cache["files"]
-        files = [
+        files: typing.List[github_types.GitHubFile] = [
             file
             async for file in self.client.items(
```
jd, line 804: can we cast this?

jd (changes requested):
I'd rather cast the `items` call so the right variable type is infered from it if that's possible.

jd (approved):

PR #1908, opened 2021-01-25
feat(queue): copy check-runs/statuses from embarked pr

sileht:
Fixes MRGFY-257


mergify_engine/merge_train.py
```diff
@@ -32,6 +32,24 @@
 from mergify_engine.clients import http
 
 
+CHECK_ASSERTS = {
+    # green check mark
+    "success": "https://raw.githubusercontent.com/Mergifyio/mergify-engine/check-runs-assets/check-green-16.png",
```
jd, line 37: Would have used a `assets` top-level directory in case we use more in the future.

jd (approved):

PR #1916, opened 2021-01-26
docs: update and simplify setup process

jd:
We don't rely on the GitHub Marketplace all the time anymore.


sileht (approved):
Stripe power!

PR #1926, opened 2021-01-26
feat(queue): add a comment when we close the merge-queue pull request

sileht:


mergify_engine/merge_train.py
```diff
@@ -273,13 +277,25 @@ async def update_summaries(
...
 
+        if status in [check_api.Conclusion.SUCCESS, check_api.Conclusion.FAILURE]:
+            await tmp_pull_ctxt.client.post(
+                f"{tmp_pull_ctxt.base_url}/issues/{self.queue_pull_request_number}/comments",
+                json={"body": tmp_pull_title},
+            )
+            await tmp_pull_ctxt.client.patch(
+                f"{tmp_pull_ctxt.base_url}/pulls/{self.queue_pull_request_number}",
+                json={"state": "closed"},
+            )
```
jd, line 297: I'm a bit confused, I thought we already closed it anyway? Where is/was that code?

sileht: Not yet. Actually, we don't close it, it's GitHub that closes it because we delete the branch. I just make it explicit.

jd: Oh ok!

jd (approved):

PR #1938, opened 2021-01-27
chore: bump GH client threshold

jd:
Since the client is per-context and per-PR anymore, it makes sense to bump it.


sileht (approved):
I have an idea to make it proportional to the number of PR proceed.

PR #1947, opened 2021-01-27
feat: adaptative requests logging

sileht:
Fixes MRGFY-292

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -315,7 +315,9 @@ class StreamProcessor:
                         owner_id, owner_login, sub, client, redis_cache
                     )
                     pulls = await self._extract_pulls_from_stream(installation)
-                    await self._consume_pulls(installation, pulls)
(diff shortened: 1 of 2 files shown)


mergify_engine/worker.py
```diff
@@ -315,6 +315,7 @@ async def consume(self, stream_name: StreamNameType) -> None:
                         owner_id, owner_login, sub, client, redis_cache
                     )
                     pulls = await self._extract_pulls_from_stream(installation)
+                    client.set_requests_ratio(len(pulls))
```
jd, line 318: can this be 0 ? 🤔

sileht: no, in theory, since we can't be here without pulls, I will ensure we don't divide by 0

jd (approved):

jd:
Semantic PR did not run.

jd:
@Mergifyio update

PR #1963, opened 2021-02-01
refactor: use owner_id in GithubActionAccessTokenAuth class

CamClrt:


mergify_engine/tests/unit/test_clients.py
```diff
@@ -68,6 +70,47 @@ async def test_client_installation_token(httpserver: httpserver.HTTPServer) -> N
...
+        )
+        httpserver.expect_request(
+            "/app/installations/12345/access_tokens"
+        ).respond_with_json(
+            {"token": "<app_token>", "expires_at": "2100-12-31T23:59:59Z"}
+        )
+
+        httpserver.expect_request(
+            "/", headers={"Authorization": "token <installation-token>"}
+        ).respond_with_json({"work": True}, status=200)
```
jd, line 101: It looks like this is really identical to the previous test. Could you refactor this so we follow [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)? 😄

jd (changes requested):

sileht:
In `_set_installation()` we actually set `self.owner_id`, can we also set `self.owner` ? so once the authentication is done, we always have both sets, whatever the one we use initially.

mergify_engine/tests/unit/test_clients.py
```diff
@@ -68,6 +70,47 @@ async def test_client_installation_token(httpserver: httpserver.HTTPServer) -> N
...
+
+        async with github.AsyncGithubInstallationClient(
+            github.get_auth(owner_id=github_types.GitHubAccountIdType(12345))
+        ) as client:
+            ret = await client.get(httpserver.url_for("/"))
+            assert ret.json()["work"]
```
sileht, line 107: Can you also check that client.auth.owner and client.auth.owner_id have the expected value (and you can also add it the previous test).

mergify_engine/tests/unit/test_clients.py
```diff
@@ -28,40 +28,74 @@
...
+    "permissions": {
+        "checks": "write",
+        "contents": "write",
+        "pull_requests": "write",
+    },
+    "account": {"login": "testing", "id": 12345},
+}
+
+
+def _httpserver_connect(httpserver: httpserver.HTTPServer, endpoint: str) -> None:
```
sileht, line 43: You can go even further by also passing the HTTP client `auth=` to this method, so when the `with AsyncGithubInstallationClient..` and all assert can be moved in this method.

jd (approved):

jd (changes requested):
just one thing: can you squash your commits?

CamClrt:
> just one thing: can you squash your commits?

Ok, I will do it in one commit. Before, I will also add a functional test.

mergify_engine/tests/unit/test_clients.py
```diff
@@ -28,44 +28,78 @@
...
+async def test_client_installation_token_with_owner_id(
+    httpserver: httpserver.HTTPServer,
+) -> None:
+    with mock.patch(
+        "mergify_engine.config.GITHUB_API_URL",
+        httpserver.url_for("/")[:-1],
+    ):
+
+        async with github.AsyncGithubInstallationClient(
+            github.get_auth(owner_id=github_types.GitHubAccountIdType(12345))
```
sileht, line 98:
```
            _httpserver_connect(
                httpserver, "/user/12345/installation", github.get_auth(owner_id=github_types.GitHubAccountIdType(12345))
            )
```

mergify_engine/tests/unit/test_clients.py
```diff
@@ -28,44 +28,78 @@
...
 
-    assert len(httpserver.log) == 3
 
-    httpserver.check_assertions()
+@mock.patch.object(github.CachedToken, "STORAGE", {})
+@pytest.mark.asyncio
+async def test_client_installation_token_with_owner_id(
+    httpserver: httpserver.HTTPServer,
+) -> None:
+    with mock.patch(
```
sileht, line 92: This can goes in _httpserver_connect

mergify_engine/tests/unit/test_clients.py
```diff
@@ -28,44 +28,78 @@
...
+        "mergify_engine.config.GITHUB_API_URL",
+        httpserver.url_for("/")[:-1],
+    ):
+
+        async with github.AsyncGithubInstallationClient(
+            github.get_auth(owner_id=github_types.GitHubAccountIdType(12345))
+        ) as client:
+            _httpserver_connect(
```
sileht, line 100: `async def` function have to be awaited to be executed. (eg: `await _httpserver_connect(...)`)

mergify_engine/tests/unit/test_clients.py
```diff
@@ -28,44 +28,78 @@
 from mergify_engine.clients import http
 
 
+gh_fake_response = {
```
sileht, line 31: The response can goes in `_httpserver_connect`, it's used anywhere else.

mergify_engine/tests/unit/test_clients.py
```diff
@@ -28,44 +28,78 @@
...
+    "permissions": {
+        "checks": "write",
+        "contents": "write",
+        "pull_requests": "write",
+    },
+    "account": {"login": "testing", "id": 12345},
+}
+
+
+async def _httpserver_connect(
```
sileht, line 43: The name is too generic,  something like `_do_test_client_installation_token()`.

mergify_engine/tests/functional/test_github_async_client.py
```diff
@@ -68,3 +68,25 @@ async def test_github_async_client(self):
...
+                    "conditions": ["base=master"],
+                    "actions": {"merge": {}},
+                }
+            ]
+        }
+
+        await self.setup_repo(yaml.dump(rules))
+        p, _ = await self.create_pr()
+
+        client = github.aget_client(self.o_integration.login, self.o_integration.id)
```
jd, line 87: since both login and ids are passed, I guess only one is used?
do we want to test only `id` being passed there?

jd (commented):
for the commit you should use `refactor`, not `chore` I'd say.

jd (commented):
LGTM I'll let @sileht validates

sileht (approved):

PR #1964, opened 2021-02-01
docs: use a custom bootstrap

sileht:


package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "devDependencies": {
+    "bootstrap": "4.6.0",
```
jd, line 3: do we need to configure dependabot for this?

sileht: done

jd (commented):
I don't think we need the `.nvimlog` file right?
should we move package.json  in the docs directory to be cleaner?

sileht:
> I don't think we need the `.nvimlog` file right?

darn, it's a recent fu^W bug of my editor :(

> should we move package.json in the docs directory to be cleaner?

I can try

jd (approved):

PR #1984, opened 2021-02-03
feat: update merge train only once per stream

sileht:
This avoid creating useless pull requests.

Fixes MRGFY-305


mergify_engine/worker.py
```diff
@@ -355,6 +359,21 @@ async def consume(self, stream_name: StreamNameType) -> None:
         )
         LOG.debug("cleanup stream end", stream_name=stream_name)
 
+    async def _refresh_merge_trains(self, installation):
+        async with self._translate_exception_to_retries(
+            installation.stream_name,
+        ):
+            for train_name in await installation.redis.keys(
+                merge_train.Train.get_redis_key_for(installation.owner_id, "*", "*")
```
jd, line 367: nitpick: would be cleaner to have a `Train.iter_trains(owner_id)` class method that would return an iterator over all the train availables.

jd (changes requested):

mergify_engine/worker.py
sileht: I like it!

jd (approved):

PR #1988, opened 2021-02-04
fix: add the type Jinja2WithNone

GuillaumeOj:


sileht (approved):

sileht:
To be perfect, you should explain the Why in the commit message.

jd (changes requested):
1. I don't see what it `fix` ?
2. No commit message 👎

PR #1990, opened 2021-02-04
chore: change the ways we check that the config file have changed.

sileht:
Fixes MRGFY-304


mergify_engine/engine/__init__.py
```diff
@@ -42,39 +42,58 @@
...
 
-            return True
-    return False
+    if config_file_to_validate is None:
+        return False
+
+    try:
+        rules.get_mergify_config(config_file_to_validate)
+    except rules.InvalidRules as e:
+        print(e)
```
jd:
```suggestion
```

mergify_engine/engine/__init__.py
```diff
@@ -42,39 +42,65 @@
...
-                    )
-                )
+                prefered_filename = f["filename"]
+                break
+        else:
+            return False
+
+    async for config_file in ctxt.repository.iter_mergify_config_files(
+        ref=ctxt.pull["head"]["sha"], prefered_filename=prefered_filename
+    ):
```
jd, line 69: I don't think we should use preferred_filename here.
In this case:
1. You have `.github/mergify.yml` in the repo already
2. The PR adds `.mergify.yml`

This is going to validate `.github/mergify.yml` but the file that is going to be used in the long run is `.mergify.yml`

jd (dismissed):

mergify_engine/engine/__init__.py
sileht: That's why I check the `path` AND the `sha` changed, to be sure I catch the new file.

sileht: I even wrote a unit test to check that ;)

mergify_engine/context.py
```diff
@@ -203,58 +207,68 @@ def get_config_location_cache_key(
...
+
+            yield MergifyConfigFile(
+                type=content["type"],
+                content=content["content"],
+                path=content["path"],
+                sha=content["sha"],
+                decoded_content=base64.b64decode(
+                    bytearray(content["content"], "utf-8")
+                ),
+            )
```
jd, line 249: I guess you could cast the dict + add the new key?
Not a big deal, that just creating a new dict is fine too :)

jd (approved):

PR #2001, opened 2021-02-08
fix: review message as None is already handled

### LINKED ISSUE #2000: 'review' action isn't working
I'm using Mergify for my [public repository](https://github.com/remal/tracing-spec/pull/202) and it seems something's been broken in 'review' action lately.

My [Mergify config](https://github.com/remal/tracing-spec/blob/master/.mergify/config.yml) has this rule:
```yaml
pull_request_rules:
- name: Auto approve dependencies
  conditions:
  - author~=^(dependabot(-preview)?|renovate)\[bot\]$
  - label!=dependencies-ci
  - label!=sync-with-template
  - -closed
  - -merged
  actions:
    review: { }
```

PR that should be automatically reviewed by Mergify, but it's not: https://github.com/remal/tracing-spec/pull/202

As I can see from Mergify summary, the conditions for the rule are met, but nothing's happening:
screen

What am I doing wrong?

sileht:
Fixes #2000


mergify_engine/actions/review.py
```diff
@@ -48,7 +48,7 @@ async def run(
     ) -> check_api.Result:
         if self.config["message"] is None:
             return check_api.Result(
-                check_api.Conclusion.SUCCESS, "Message is not set", ""
+                check_api.Conclusion.FAILURE, "Message is not set", ""
```
jd, line 51:
```suggestion
                check_api.Conclusion.FAILURE, "Message is not set", "The `review` action does not have any message configured."
```

or I'd do this:
```python
message = self.config["message"] or "👀"
```
if we want to make it work by default
maybe use 👍  / 👎  / 💬 by default depending on the `type`?

jd (dismissed):

mergify_engine/actions/review.py
sileht: I just see it's already handled some lines later.

sileht:
That needs a bit more work, re-record review tests don't work anymore.

jd (approved):

PR #2006, opened 2021-02-09
refactor(queue API): remove installation_id

CamClrt:


mergify_engine/web/__init__.py
```diff
@@ -358,6 +358,31 @@ async def queues(installation_id):
     return responses.JSONResponse(status_code=200, content=queues)
 
 
+@app.get(
+    "/queuesV2/{owner_id}",  # noqa: FS003
+    dependencies=[fastapi.Depends(auth.signature)],
+)
+async def queuesV2(owner_id):
+    queues = collections.defaultdict(dict)
+    async for queue in _AREDIS_CACHE.scan_iter(match=f"merge-queue~{owner_id}~*"):
```
jd, line 367: Isn't there a security hole there where you can request the queue for any repository if you know the owner id?

(this applies to the original method too, but it's harder since the installation id is not public)

jd (dismissed):

mergify_engine/web/__init__.py
sileht: I don't think it's a big deal, since this endpoint is protected by `auth.signature`. This authentication method just trust that the dashboard has done the check first.

If one day we make this API public, we need to create a new authentication method that does the oauth flow on engine side that will validate the owner_id

mergify_engine/web/__init__.py
```diff
@@ -358,6 +358,31 @@ async def queues(installation_id):
     return responses.JSONResponse(status_code=200, content=queues)
 
 
+@app.get(
+    "/queuesV2/{owner_id}",  # noqa: FS003
+    dependencies=[fastapi.Depends(auth.signature)],
+)
+async def queuesV2(owner_id):
```
sileht, line 365: Maybe a meaningful endpoint name like: `queues-by-owner-id`

sileht (changes requested):
Looks good, just add one test to ensure the new endpoint works.

jd (approved):

PR #2012, opened 2021-02-09
feat(simulator): return parsed rule conditions

sileht:
This returns the parsed pull request rules conditions with the
configuration validation to the dashboard.

### DIFF
--- a/mergify_engine/tests/unit/test_simulator.py
+++ b/mergify_engine/tests/unit/test_simulator.py
@@ -1,0 +1,65 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2021 Mergify SAS
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


mergify_engine/tests/unit/test_simulator.py
```diff
@@ -0,0 +1,65 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2020 Mergify SAS
```
jd:
```suggestion
# Copyright © 2021 Mergify SAS
```

mergify_engine/web/simulator.py
```diff
@@ -141,6 +141,17 @@ async def simulator(request: requests.Request) -> responses.JSONResponse:
     else:
         title, summary = ("The configuration is valid", None)
 
+    pull_request_rules_conditions = []
```
jd, line 144: Any reason not to write that as a list comprehension?

jd (changes requested):

mergify_engine/web/simulator.py
sileht: you are too fast for reviewing ;)

jd (approved):

PR #2039, opened 2021-02-11
refactor: use check-success instead of status-success

GuillaumeOj:


.mergify.yml
```diff
@@ -1,22 +1,22 @@
...
     conditions:
       - base=master
-      - status-success=Semantic Pull Request
-      - "status-success=requirements"
-      - "status-success=datadog"
-      - "status-success=pep8"
-      - "status-success=py39"
-      - "status-success=docs"
+      - check-success=Semantic Pull Request
+      - check-success=requirements"
```
jd, line 15:
```suggestion
      - check-success=requirements
```

jd (changes requested):

jd (approved):

PR #2040, opened 2021-02-11
feat: add docker configuration

sileht:
Initial step to be able to deploy the engine as docker image

Fixes MRGFY-12

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,3 +75,9 @@ jobs:
       - run: |
           pip install tox
           tox -e docs
+
+  docker:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/build-push-action@v1
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,19 @@
+name: Publish Docker image
+on:
+  release:
+    types: [published]
+jobs:
+  push_to_registry:
+    name: Push Docker image to GitHub Packages
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Push to GitHub Packages
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          registry: docker.pkg.github.com
+          repository: Mergifyio/engine
+          tag_with_ref: true
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ AUTHORS
 cover
 .coverage
 dist
+.mypy_cache
 .eggs
 .pytest_cache
 .mypy_cache
@@ -16,3 +17,4 @@ prod.monitoring.env
 constraints.txt
 docs/node_modules
 docs/source/actions/action-list.html
+.git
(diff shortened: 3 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN sed -e '/^-e ./d' /requirements.txt > /constraints.txt
+RUN python3 -m pip install --no-cache-dir -r /constraints.txt
+
+# Real install that can't be cached
+ADD . /app
+WORKDIR /app
+RUN python3 -m pip install --no-cache-dir -r ./requirements.txt
+
+
+FROM python:3.9-alpine as runner
```
jd, line 18: can we use ubuntu?

.github/workflows/ci.yaml
```diff
@@ -75,3 +75,9 @@ jobs:
       - run: |
           pip install tox
           tox -e docs
+
+  docker:
+    runs-on: ubuntu-20.04
```
jd, line 80: not latest?

jd (changes requested):

.github/workflows/ci.yaml
sileht: latest is randomly 18.04 or 20.04 currently (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on)

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,19 @@
+name: Publish Docker image
+on:
+  release:
+    types: [published]
+jobs:
+  push_to_registry:
+    name: Push Docker image to GitHub Packages
+    runs-on: ubuntu-latest
```
jd, line 8: do we want latest or 20.04 then here?

jd: ping @sileht

jd (changes requested):

jd (approved):

setup.cfg
```diff
@@ -24,6 +24,7 @@ install_requires =
     datadog
     python-dotenv
     fastapi
+    aiofiles
```
jd, line 27: why?
you missed the asgi file I guess

.gitignore
```diff
@@ -5,6 +5,7 @@ AUTHORS
 cover
 .coverage
 dist
+.mypy_cache
```
jd, line 8: https://github.com/Mergifyio/mergify-engine/pull/2060

jd (changes requested):

setup.cfg
sileht: That's used by the installer, that shouldn't be there

entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+cd /app
+
+if [ "$MERGIFYENGINE_INTEGRATION_ID" ]; then
+  case $1 in
+      web) exec gunicorn -k uvicorn.workers.UvicornH11Worker --statsd-host localhost:8125 --log-level warning mergify_engine.asgi;;
```
jd, line 6: this ain't gonna work

jd (changes requested):

entrypoint.sh
sileht: why ?

jd: I misread 😢

jd (approved):

PR #2041, opened 2021-02-11
feat: allow to pass subscription via environment variables

sileht:
Fixes MRGFY-314

### DIFF
--- a/mergify_engine/config.py
+++ b/mergify_engine/config.py
@@ -23,6 +23,9 @@ import typing
 import dotenv
 import voluptuous
 
+from mergify_engine import subscription_key
+from mergify_engine import types
+
 
 GITHUB_APP = os.environ.get("MERGIFYENGINE_MODE", "github_app") == "github_app"
 if GITHUB_APP:
@@ -51,6 +54,17 @@ def CommaSeparatedStringList(value: str) -> typing.List[str]:
 
 def CommaSeparatedIntList(value: str) -> typing.List[int]:
     return [int(s) for s in value.split(",")]
+
+
+def AccountTokens(v: str) -> typing.Dict[str, str]:
+    try:
+        return dict(
+            typing.cast(typing.Tuple[str, str], tuple(map(str.strip, bot.split(":"))))
+            for bot in v.split(",")
+            if bot.strip()
+        )
+    except ValueError:
+        raise ValueError("wrong format, expect `login1:token1,login2:token2`")
 
 
 Schema = voluptuous.Schema(
@@ -83,6 +97,16 @@ Schema = voluptuous.Schema(
         voluptuous.Required("GITHUB_URL", default="https://github.com"): str,
         voluptuous.Required("GITHUB_API_URL", default="https://api.github.com"): str,
         # Mergify website for subscription
+        voluptuous.Required("SUBSCRIPTION_KEY", default=None): voluptuous.Any(
+            None, voluptuous.Coerce(subscription_key.DecryptedSubscriptionKey)
+        ),
+        voluptuous.Required("SUBSCRIPTION_PRIVATE_KEY", default=None): voluptuous.Any(
+            None,
+            str,
+        ),
+        voluptuous.Required("ACCOUNT_TOKENS", default=None): voluptuous.Any(
+            None, voluptuous.Coerce(AccountTokens)
+        ),
         voluptuous.Required(
             "SUBSCRIPTION_BASE_URL", default="http://localhost:5000"
         ): str,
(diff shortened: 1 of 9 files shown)


mergify_engine/config.py
```diff
@@ -53,6 +56,19 @@ def CommaSeparatedIntList(value: str) -> typing.List[int]:
     return [int(s) for s in value.split(",")]
 
 
+def BotAccounts(v: str) -> typing.Dict[str, str]:
+    try:
+        return dict(
+            typing.cast(
+                typing.Tuple[str, str], tuple(map(lambda v: v.strip(), bot.split(":")))
+            )
```
jd, line 64:
```suggestion
tuple(map(str.strip, bot.split(":")))
```

mergify_engine/config.py
```diff
@@ -53,6 +56,19 @@ def CommaSeparatedIntList(value: str) -> typing.List[int]:
...
+            for bot in v.split(",")
+            if bot.strip()
+        )
+    except ValueError:
+        raise ValueError("wrong format, we expect `login1:token1,login2:token2`")
```
jd, line 69:
```suggestion
        raise ValueError("wrong format, expected `login1:token1,login2:token2`")
```

mergify_engine/config.py
```diff
@@ -158,6 +184,9 @@ def CommaSeparatedIntList(value: str) -> typing.List[int]:
 NOSUB_MAX_REPO_SIZE_KB: int
 GIT_EMAIL: str
 CONTEXT: str
+BOT_ACCOUNTS: typing.Dict[str, str]
```
jd, line 187: That should not be "bot account".
This is also going to be used for other things in the future, such as rebasing as the user for commands.

mergify_engine/subscription.py
```diff
@@ -120,6 +104,22 @@ async def get_subscription(
         cls, redis: utils.RedisCache, owner_id: int
     ) -> "Subscription":
         """Get a subscription."""
+        if config.SUBSCRIPTION_KEY is not None:
```
jd, line 107: Could we have an entirely different `Subscription` class rather than a `if`?

mergify_engine/subscription_key.py
```diff
@@ -0,0 +1,182 @@
...
+        subscription_key = SubscriptionKey(**data)
+    except pydantic.ValidationError:
+        raise ValueError("incorrect subscription key data")
+
+    try:
+        SUBSCRIPTION_PUBLIC_KEY.verify(
+            subscription_key.signature, subscription_key.data
+        )  # type: ignore
+    except cryptography.exceptions.InvalidSignature:
+        raise ValueError("incorrect subscription key signature")
```
jd, line 86: I'd propably leave the `InvalidSignature` exception being raised here

mergify_engine/subscription_key.py
```diff
@@ -0,0 +1,182 @@
...
+def DecryptedSubscriptionKey(key: str) -> typing.Dict[int, types.SubscriptionDict]:
+    try:
+        data = msgpack.unpackb(base64.b64decode(key.encode()))
+    except Exception:
+        raise ValueError("incorrect subscription key format")
+
+    try:
+        subscription_key = SubscriptionKey(**data)
+    except pydantic.ValidationError:
+        raise ValueError("incorrect subscription key data")
```
jd, line 79: not sure it's worth it

mergify_engine/subscription_key.py
```diff
@@ -0,0 +1,182 @@
...
+        encoding=serialization.Encoding.Raw, format=serialization.PublicFormat.Raw
+    )
+    current_public_key = private_key.public_key().public_bytes(
+        encoding=serialization.Encoding.Raw, format=serialization.PublicFormat.Raw
+    )
+    if current_public_key != expected_public_key:
+        print("Unexpected private key")
+        return 1
+
+    parser = argparse.ArgumentParser(description="Generate subscription key")
```
jd, line 154: This needs to handle datetime range of validity.

mergify_engine/subscription_key.py
```diff
@@ -0,0 +1,182 @@
+# -*- encoding: utf-8 -*-
```
jd, line 1: I'm not sure about this as it's a pain to update the file then.
I'd rather start with a simple mode where the engine call a public API with a secret to get a JSON description of its subscription.

jd (dismissed):

tox.ini
```diff
@@ -101,7 +101,7 @@ whitelist_externals =
 commands =
   flake8
   isort -c .
-  mypy
+  mypy --show-error-codes
```
jd, line 104: https://github.com/Mergifyio/mergify-engine/pull/2061

sileht:
Just use half of this -> #2062

PR #2056, opened 2021-02-15
feat: introduce installer

sileht:


.gitignore
```diff
@@ -5,6 +5,7 @@ AUTHORS
 cover
 .coverage
 dist
+.mypy_cache
```
jd, line 8: Split in another PR

.gitignore
```diff
@@ -15,3 +16,9 @@ prod.monitoring.env
 constraints.txt
 docs/node_modules
 docs/source/actions/action-list.html
+.git
+installer/node_modules
+installer/build
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
jd, line 24: Those last 3 should be in a different PR too

installer/asgi.py
```diff
@@ -0,0 +1,22 @@
...
+# under the License.
+
+import fastapi
+from fastapi.staticfiles import StaticFiles
+
+
+application = fastapi.FastAPI()
+application.mount(
+    "/", StaticFiles(directory="/app/installer/build", html=True), name="installer"
+)
```
jd, line 22: Should this be placed in a installer directory so it's clear it's not the engine itself?

jd (changes requested):
Is there a good reason to have this installer in this repo?

jd (commented):
Do we need dependabot set up for this directory?

sileht:
> Do we need dependabot set up for this directory?

yes

jd (approved):

PR #2066, opened 2021-02-16
feat: rework subscription cache handling

sileht:
We now cache subscription for 3 days, but check the databases every 1
hours as before.

This allow to keep the last subscription when connectivity issue occurs
on dashboard side.

Fixes MRGFY-318

### DIFF
--- a/mergify_engine/subscription.py
+++ b/mergify_engine/subscription.py
@@ -22,6 +22,7 @@ import daiquiri
 
 from mergify_engine import config
 from mergify_engine import crypto
+from mergify_engine import exceptions
 from mergify_engine import utils
 from mergify_engine.clients import http
 
@@ -66,6 +67,7 @@ class Subscription:
     reason: str
     tokens: typing.Dict[str, str]
     features: typing.FrozenSet[enum.Enum]
+    ttl: int = -2
 
     @staticmethod
     def _to_features(feature_list: typing.Iterable[str]) -> typing.FrozenSet[Features]:
@@ -89,7 +91,11 @@ class Subscription:
 
     @classmethod
     def from_dict(
-        cls, redis: utils.RedisCache, owner_id: int, sub: SubscriptionDict
+        cls,
+        redis: utils.RedisCache,
+        owner_id: int,
+        sub: SubscriptionDict,
+        ttl: int = -2,
     ) -> "Subscription":
         return cls(
             redis,
@@ -98,6 +104,7 @@ class Subscription:
             sub["subscription_reason"],
             sub["tokens"],
             cls._to_features(sub.get("features", [])),
+            ttl,
         )
 
     def get_token_for(self, wanted_login: str) -> typing.Optional[str]:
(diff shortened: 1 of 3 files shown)


mergify_engine/subscription.py
```diff
@@ -115,22 +116,45 @@ def to_dict(self) -> SubscriptionDict:
             "features": [f.value for f in self.features],
         }
 
+    RETENTION_SECONDS = 60 * 60 * 24 * 3  # 3 days
+    VALIDITY_SECONDS = 3600
+
+    async def _have_expired(self) -> bool:
+        expire_at: int = await self.redis.ttl(
```
jd, line 123:
```suggestion
        ttl: int = await self.redis.ttl(
```
expire_at implies it's a timestamp and it's not

mergify_engine/subscription.py
```diff
@@ -115,22 +116,45 @@ def to_dict(self) -> SubscriptionDict:
             "features": [f.value for f in self.features],
         }
 
+    RETENTION_SECONDS = 60 * 60 * 24 * 3  # 3 days
+    VALIDITY_SECONDS = 3600
+
+    async def _have_expired(self) -> bool:
```
jd, line 122:
```suggestion
    async def _has_expired(self) -> bool:
```

mergify_engine/subscription.py
```diff
@@ -115,22 +116,45 @@ def to_dict(self) -> SubscriptionDict:
...
         cls, redis: utils.RedisCache, owner_id: int
     ) -> "Subscription":
         """Get a subscription."""
-        sub = await cls._retrieve_subscription_from_cache(redis, owner_id)
-        if sub is None:
-            sub = await cls._retrieve_subscription_from_db(redis, owner_id)
-            await sub.save_subscription_to_cache()
-        return sub
+
+        cached_sub = await cls._retrieve_subscription_from_cache(redis, owner_id)
```
jd, line 137: might be better to return the ttl at the same time using a transaction

mergify_engine/subscription.py
```diff
@@ -115,22 +116,45 @@ def to_dict(self) -> SubscriptionDict:
...
+    VALIDITY_SECONDS = 3600
+
+    async def _have_expired(self) -> bool:
+        expire_at: int = await self.redis.ttl(
+            f"subscription-cache-owner-{self.owner_id}"
+        )
+        if expire_at < 0:
+            return True
+
+        return expire_at < (self.RETENTION_SECONDS - self.VALIDITY_SECONDS)
```
jd, line 129: I'd just use `self.VALIDITY_SECONDS` here, no?

jd (changes requested):

mergify_engine/subscription.py
sileht: I try to rewrite the condition to make it clearer

mergify_engine/subscription.py
sileht: done

jd (approved):

PR #2069, opened 2021-02-16
ci: enable dependabot for gh-actions

sileht:


.github/dependabot.yml
```diff
@@ -1,16 +1,23 @@
...
-    timezone: Europe/Paris
-  open-pull-requests-limit: 10
-- package-ecosystem: npm
-  directory: "/docs"
-  schedule:
-    interval: daily
-    time: "09:00"
-    timezone: Europe/Paris
-  open-pull-requests-limit: 10
+    :w
```
jd, line 3: nop

jd (changes requested):

jd (approved):

PR #2070, opened 2021-02-16
fix: subscription decrypt typing

sileht:
We unfortunately use decode_responses=True for `RedisCache`,
redis.get will returns str instead of bytes.

I wonder how this can have worked...


jd (approved):

sileht:
@Mergifyio refresh

sileht:
@Mergifyio refresh

sileht:
Wierd,Mergify can't merge it because of: 
```
403 Client Error: Forbidden for url `https://api.github.com/repos/Mergifyio/mergify-engine/merges`
Details: Resource not accessible by integration
```

sileht:
@Mergifyio refresh

sileht:
image

Interesting...

PR #2085, opened 2021-02-18
refactor: remove need to support multiple operator for one operation

sileht:
This ensures we only have one version of each operator. This will
simplify the number of thing to handle on dashboard side.


jd (commented):
is it really worth it?
just in case somebody used those 😉

sileht:
People can still use them in the configuration, this change just converts them to only one representation upfront, so when we write code we only have one version of each operator to handle.

jd (approved):
Oh you're right. Yeah makes sense!

PR #2101, opened 2021-02-22
feat: allow to configure the ssl verification mode

sileht:
Related: #MRGFY-218


jd (approved):

jd:
ticket number has wrong syntax

PR #2112, opened 2021-02-23
feat(queue): allow to configure train length

sileht:
Fixes MRGFY-327

### DIFF
--- a/docs/source/queues.rst
+++ b/docs/source/queues.rst
@@ -160,4 +160,10 @@ A ``queue_rules`` takes the following parameter:
      -
      - The list of ``conditions`` to match to get the queued pull request merged.
 
+Speculative Queue Length
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The number of pull request to embark into the merge train can be configured
+with ``speculative_queue_length``. The default is ``5``.
+
 .. include:: global-substitutions.rst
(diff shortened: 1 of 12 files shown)


docs/source/queues.rst
```diff
@@ -160,4 +160,10 @@ A ``queue_rules`` takes the following parameter:
      -
      - The list of ``conditions`` to match to get the queued pull request merged.
 
+Speculative Queue Length
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The number of pull request to embark into the merge train can be configured
+with ``speculative_queue_length``. The default is ``5``.
```
jd, line 167: Why isn't that in the rules parameter above?

sileht: It doesn't make sense to have this per queue rule. For example we have two rules:

```
- name: high
  queue_len: 2
  priority: 10
- name default
  queue_len: 3
  priority: 1000
```

In case of the current queue is:
```
#1, #2, #3, #4 (high)
#5, #6 (default)
```

Should I batch only 2 PRS ?

But when it becomes 
```
#4 (high)
#5, #6, #7, #8 (default)
```

Here should I batch only 2 PRs, 3, 5 ?

I think it's better to just choose on global length.

The only case I see it can be useful is per base branch. Maybe you want to allow more speculative merge on dev branch than on stable branches.

sileht:
Replaced by https://github.com/Mergifyio/mergify-engine/pull/2252

PR #2115, opened 2021-02-23
feat: don't create uneeded tmp pull request

sileht:
When the head of the train doesn't have parents pull request. We can
just rebase the pull request instead of creating a new one.

Fixes MRGFY-326


mergify_engine/actions/queue.py
```diff
@@ -96,6 +96,9 @@ def validate_config(self, mergify_config: "rules.MergifyConfig") -> None:
         self.config["update_bot_account"] = None
         self.config["strict"] = merge_base.StrictMergeParameter.ordered
 
+        # Should we support other ?
```
jd, line 99: I'd say yes, especially if we want to move away from `strict` to queue only.

mergify_engine/actions/queue.py
```diff
@@ -146,6 +149,11 @@ async def _should_be_merged(self, ctxt: context.Context, q: queue.Queue) -> bool
 
     async def _should_be_synced(self, ctxt: context.Context, q: queue.Queue) -> bool:
         # NOTE(sileht): since we create dedicated branch we don't need to sync PR
+        if await ctxt.is_behind and await q.is_first_pull(ctxt):
```
jd, line 152:
```suggestion
        if await q.is_first_pull(ctxt) and await ctxt.is_behind:
```

neat, but I guess `is_first_pull` is cheaper to execute :)

jd: Approving if you think it's not worth it, remove the label

jd (approved):

mergify_engine/queue/merge_train.py
```diff
@@ -66,6 +67,53 @@ class WaitingPull(typing.NamedTuple):
...
 
 
+class TrainCarState(enum.Enum):
+    pending = "pending"
+    created = "created"
+    updated = "updated"
+    failed = "failed"
+
+
+json.register_type(TrainCarState)
```
jd, line 77: Enum a particularly cumbersome to use, especially with json.
Isn't a `typing.Literal` good enough?

mergify_engine/queue/merge_train.py
```diff
@@ -448,6 +544,25 @@ async def _save(self):
         else:
             await self.repository.installation.redis.delete(self._get_redis_key())
 
+    def should_be_update(
```
jd, line 547:
```suggestion
    def should_be_updated(
```

jd (dismissed):

jd (approved):

PR #2117, opened 2021-02-24
feat: use new dashboard subscription endpoint

sileht:
Fixes Fixes MRGFY-177

### DIFF
--- a/mergify_engine/user_tokens.py
+++ b/mergify_engine/user_tokens.py
@@ -1,0 +1,131 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2021 Mergify SAS
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+import dataclasses
+import json
+import typing
+
+import daiquiri
+
+from mergify_engine import config
+from mergify_engine import crypto
+from mergify_engine import exceptions
+from mergify_engine import utils
+from mergify_engine.clients import http
+
+
+LOG = daiquiri.getLogger(__name__)
+
+
+class UserTokensDict(typing.TypedDict):
+    tokens: typing.Dict[str, str]
+
+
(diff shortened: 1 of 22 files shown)


mergify_engine/user_tokens.py
```diff
@@ -0,0 +1,131 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2018–2020 Mergify SAS
```
jd, line 3:
```suggestion
# Copyright © 2021 Mergify SAS
```

mergify_engine/user_tokens.py
```diff
@@ -0,0 +1,131 @@
...
+        cached_tokens = await cls._retrieve_from_cache(redis, owner_id)
+        if cached_tokens is None or await cached_tokens._has_expired():
+            try:
+                db_tokens = await cls._retrieve_from_db(redis, owner_id)
+            except Exception as exc:
+                if cached_tokens is not None and (
+                    exceptions.should_be_ignored(exc) or exceptions.need_retry(exc)
+                ):
+                    # NOTE(sileht): return the cached tokens, instead of retry the stream,
+                    # just because the dashboard have connectivity issue.
```
jd, line 80:
```suggestion
                    # NOTE(sileht): return the cached tokens, instead of retrying the stream,
                    # just because the dashboard has a connectivity issue.
```

mergify_engine/web/__init__.py
```diff
@@ -274,26 +275,20 @@ async def subscription_cache_update(
...
+    await user_tokens.UserTokens.delete(_AREDIS_CACHE, owner_id)
+
+
 @app.delete(
     "/subscription-cache/{owner_id}",  # noqa: FS003
     dependencies=[fastapi.Depends(auth.signature)],
 )
 async def subscription_cache_delete(owner_id):  # pragma: no cover
-    await _AREDIS_CACHE.delete(f"subscription-cache-owner-{owner_id}")
+    await cleanup_subscription(owner_id)
```
jd, line 288: There's no need to clean up the token cache here.

Actually, we should delete the token from the engine cache when we update the account on the dashboard and it becomes invalid, but we don't do it right now. If we add a new endpoint to the engine to delete cached token, we could improve the current workflow.

mergify_engine/web/__init__.py
```diff
@@ -310,7 +305,7 @@ async def marketplace_handler(
         gh_owner=data["marketplace_purchase"]["account"]["login"],
     )
 
-    await cleanup_subscription(data)
+    await cleanup_subscription(data["marketplace_purchase"]["account"]["id"])
```
jd, line 308: ditto

jd (changes requested):

jd (approved):

jd:
Approving though you missed some suggestions.

PR #2126, opened 2021-02-26
refactor: remove dependency between Queue() and Train()

sileht:
Many informations are duplicates between Queue() and Train()

merge_base.MergeBaseAction() just need some methods to be implemented to
get user facing data.

All this informations can be retrieved in Train() directly.
Only the config is missing. So it will now be stored within the train
data.


sileht:
I didn't handle the upgrade of the current Redis database data as no customer is using this yet and it's beta.

jd (approved):

PR #2132, opened 2021-02-28
chore: remove useless logs

sileht:


mergify_engine/context.py
```diff
@@ -857,6 +857,14 @@ def have_been_synchronized(self) -> bool:
                     return True
         return False
 
+    def have_been_opened(self) -> bool:
```
jd, line 860: `has_been_opened`

jd (changes requested):

jd (approved):

PR #2142, opened 2021-03-01
refactor-MRGFY-215: Remove `installation_id` from the queue API

CamClrt:


mergify_engine/web/__init__.py
```diff
@@ -328,35 +327,6 @@ async def marketplace_handler(
...
-                )
-            queues[owner + "/" + repo["name"]][branch] = [
-                int(pull) async for pull, _ in _AREDIS_CACHE.zscan_iter(queue)
-            ]
-
-    return responses.JSONResponse(status_code=200, content=queues)
-
-
 @app.get(
     "/queues-by-owner-id/{owner_id}",  # noqa: FS003
```
jd, line 331: it'd be great to rename this endpoint to `/queues` also

jd (approved):

mergify_engine/web/__init__.py
```diff
@@ -329,39 +328,10 @@ async def marketplace_handler(
 
 
 @app.get(
-    "/queues/{installation_id}",  # noqa: FS003
+    "/queues/{owner_id}",  # noqa: FS003
```
jd, line 331: ok, but think that you're going to break the current usage so you need to fix it first… that's gonna work like that.

jd (changes requested):

mergify_engine/web/__init__.py
CamClrt: It's already done in this PR: https://github.com/Mergifyio/mergify.io/pull/1901/files#diff-f92db070948ef8a9b2a01cb58fdd7f0396ecba62fb8a3f2244413fa21b8a5c12R94. I made it a few minutes ago.

CamClrt: Is it possible to merge the 2 PR at the same time?

jd: Not really.
I mean, yes, in theory, we _could_ press the merge buttons at the same time, *but* you shouldn't work that way. It's not safe, and if anything breaks in the middle, it's going to be a pain to go back.

Ending up with half-working stuff is never a good plan. That's why you should plan to do the change in multiple steps like you did so far. It's a little bit more work, but it's way safer as it can be interrupted at any point in time without breaking anything.

jd (changes requested):
This is going to break existing usage.

mergify_engine/web/__init__.py
```diff
@@ -329,28 +328,24 @@ async def marketplace_handler(
 
 
 @app.get(
```
sileht, line 330: We should drop this method when the dashboard side have been updated to use the new one

mergify_engine/web_cli.py
```diff
@@ -76,9 +76,7 @@ def refresher():
...
 def queues():
     parser = argparse.ArgumentParser(description="Show queue of mergify_engine")
-    parser.add_argument("installation_id")
+    parser.add_argument("owner_id")
 
     args = parser.parse_args()
-    asyncio.run(
-        api_call(config.BASE_URL + f"/queues/{args.installation_id}", method="GET")
-    )
+    asyncio.run(api_call(config.BASE_URL + f"/queues/{args.owner_id}", method="GET"))
```
sileht, line 82: We should use the new endpoint here.

sileht (changes requested):
we can cleanup more stuff.

mergify_engine/web/__init__.py
```diff
@@ -329,39 +328,35 @@ async def marketplace_handler(
...
-                        "message": f"{installation['account']['type']} account `{owner}` rate limited by GitHub"
+                        "message": f"{client.auth.owner} account with {client.auth.owner_id} ID, rate limited by GitHub"
                     },
                 )
-            queues[owner + "/" + repo["name"]][branch] = [
+            queues[client.auth.owner + "/" + repo["name"]][branch] = [
                 int(pull) async for pull, _ in _AREDIS_CACHE.zscan_iter(queue)
             ]
 
     return responses.JSONResponse(status_code=200, content=queues)
```
jd, line 353: This should not be duplicated. Reuse the same function.

mergify_engine/web_cli.py
```diff
@@ -76,9 +76,12 @@ def refresher():
...
 def queues():
     parser = argparse.ArgumentParser(description="Show queue of mergify_engine")
-    parser.add_argument("installation_id")
+    parser.add_argument("owner_id")
 
     args = parser.parse_args()
     asyncio.run(
-        api_call(config.BASE_URL + f"/queues/{args.installation_id}", method="GET")
+        api_call(
+            config.BASE_URL + f"/queues-by-owner-id/{args.owner_id}",
```
jd, line 84: Use the new URL `/queues`

jd (changes requested):

jd (approved):

PR #2151, opened 2021-03-01
fix: don't mark train car as failed to early

sileht:
In case, of CIs on pull request at the tail of the train finish before
the head and fail. Pull request are marked as FAILURE and got
dis-embarked. But maybe the failure is not due to itself, but a parents
pull request.

This change ensures in this case the state stay as pending until we are
sure the failure is the pull request fault.


jd (approved):
`too early`

PR #2187, opened 2021-03-11
feat(web): allow to limit number of redis connections

sileht:
The default pool limit is 2**31, this allows to configure it.

### DIFF
--- a/mergify_engine/config.py
+++ b/mergify_engine/config.py
@@ -114,6 +114,12 @@ Schema = voluptuous.Schema(
         voluptuous.Required(
             "REDIS_SSL_VERIFY_MODE_CERT_NONE", default=False
         ): CoercedBool,
+        voluptuous.Required(
+            "REDIS_STREAM_WEB_MAX_CONNECTIONS", default=None
+        ): voluptuous.Any(None, voluptuous.Coerce(int)),
+        voluptuous.Required(
+            "REDIS_CACHE_WEB_MAX_CONNECTIONS", default=None
+        ): voluptuous.Any(None, voluptuous.Coerce(int)),
         voluptuous.Required("STORAGE_URL", default="redis://localhost:6379?db=8"): str,
         voluptuous.Required("STREAM_URL", default=None): voluptuous.Any(None, str),
         voluptuous.Required("STREAM_PROCESSES", default=1): voluptuous.Coerce(int),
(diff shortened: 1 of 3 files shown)


mergify_engine/config.py
```diff
@@ -114,6 +114,12 @@ def AccountTokens(v: str) -> typing.Dict[str, str]:
         voluptuous.Required(
             "REDIS_SSL_VERIFY_MODE_CERT_NONE", default=False
         ): CoercedBool,
+        voluptuous.Required(
+            "REDIS_STREAM_WEB_MAX_CONNETIONS", default=None
```
jd, line 118:
```suggestion
            "REDIS_STREAM_WEB_MAX_CONNECTIONS", default=None
```

mergify_engine/config.py
```diff
@@ -114,6 +114,12 @@ def AccountTokens(v: str) -> typing.Dict[str, str]:
...
+        ): voluptuous.Any(None, voluptuous.Coerce(int)),
+        voluptuous.Required(
+            "REDIS_CACHE_WEB_MAX_CONNETIONS", default=None
```
jd, line 121:
```suggestion
            "REDIS_CACHE_WEB_MAX_CONNECTIONS", default=None
```

mergify_engine/config.py
```diff
@@ -177,6 +183,8 @@ def AccountTokens(v: str) -> typing.Dict[str, str]:
 CONTEXT: str
 ACCOUNT_TOKENS: typing.Dict[str, str]
 REDIS_SSL_VERIFY_MODE_CERT_NONE: bool
+REDIS_STREAM_WEB_MAX_CONNETIONS: typing.Optional[int]
+REDIS_CACHE_WEB_MAX_CONNETIONS: typing.Optional[int]
```
jd, line 187:
```suggestion
REDIS_STREAM_WEB_MAX_CONNECTIONS: typing.Optional[int]
REDIS_CACHE_WEB_MAX_CONNECTIONS: typing.Optional[int]
```

jd (changes requested):

sileht:
Wierd mypy didn't catch them

mergify_engine/web/__init__.py
```diff
@@ -50,8 +50,12 @@
 @app.on_event("startup")
 async def startup():
```
jd, line 51: This is not typed, this is why mypy did not catch it.
Would you mind typing it?

jd (changes requested):

mergify_engine/web/__init__.py
sileht: Ah thx!

jd (approved):

PR #2190, opened 2021-03-11
feat: change default shutdown timeout to 10 seconds

sileht:


mergify_engine/worker.py
```diff
@@ -589,7 +589,7 @@ def get_process_index_from_env() -> int:
 @dataclasses.dataclass
 class Worker:
     idle_sleep_time: float = 0.42
-    shutdown_timeout: float = 25
+    shutdown_timeout: float = config.WORKER_SHUTDOWN_TIMEOUT
```
jd, line 592: you say `int` above and float here. Decide plz 😁

jd (dismissed):

jd (approved):

PR #2200, opened 2021-03-12
feat: log py.warning messages

sileht:


jd (changes requested):
https://github.com/jd/daiquiri/blob/master/daiquiri/__init__.py#L111-L112

PR #2216, opened 2021-03-16
refactor(queue API): removing GitHub API call

CamClrt:
add a second temporary queues route before updating the dashboard part

### DIFF
--- a/mergify_engine/web/__init__.py
+++ b/mergify_engine/web/__init__.py
@@ -255,6 +255,22 @@ async def queues_by_owner_id(owner_id):
     return responses.JSONResponse(status_code=200, content=queues)
 
 
+@app.get(
+    "/queues_v2/{owner_id}",  # noqa: FS003
+    dependencies=[fastapi.Depends(auth.signature)],
+)
+async def queues(owner_id):
+    global _AREDIS_CACHE
+    queues = collections.defaultdict(dict)
+    async for queue in _AREDIS_CACHE.scan_iter(match=f"merge-queue~{owner_id}~*"):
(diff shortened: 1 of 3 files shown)


mergify_engine/web/__init__.py
```diff
@@ -255,6 +255,23 @@ async def queues_by_owner_id(owner_id):
...
+)
+async def queues(owner_id):
+    global _AREDIS_CACHE
+    queues = collections.defaultdict(dict)
+    async for queue in _AREDIS_CACHE.scan_iter(match=f"merge-queue~{owner_id}~*"):
+        _, _, repo_id, branch = queue.split("~")
+        async with github.aget_client(owner_id=owner_id) as client:
+            queues[client.auth.owner + "/" + repo_id][branch] = [
+                int(pull) async for pull, _ in _AREDIS_CACHE.zscan_iter(queue)
+            ]
```
jd, line 270: I don't think we want to make a client here to return the owner, it'd be enough to return the repo id and that's it.
The caller knows the `owner_id` since it is sending it.
Thoughts @sileht?

sileht: Agreed.

CamClrt: Do you mean just `queues[repo_id][branch]` ?

jd: Yes 👍

CamClrt: ok

sileht (changes requested):
Looks good, just need a unit test to cover the new code.

jd (approved):

PR #2218, opened 2021-03-16
feat(queue): add support strict_method

sileht:
When the first PR in queue is rebased, we can now select the method to
use.

### DIFF
--- a/mergify_engine/actions/queue.py
+++ b/mergify_engine/actions/queue.py
@@ -46,9 +46,11 @@ class QueueAction(merge_base.MergeBaseAction):
         voluptuous.Required("rebase_fallback", default="merge"): voluptuous.Any(
             "merge", "squash", None
         ),
-        voluptuous.Required("merge_bot_account", default=None): voluptuous.Any(
-            None, types.GitHubLogin
+        voluptuous.Required("strict_method", default="merge"): voluptuous.Any(
+            "rebase", "merge"
         ),
+        voluptuous.Required("merge_bot_account", default=None): types.Jinja2WithNone,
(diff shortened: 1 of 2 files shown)


sileht:
Not sure we should call it strict_method here. Also this option may be confusing since we use it only for the first PR in train and not for the temporary merge-queue pull request.

sileht:
Depends-On: https://github.com/Mergifyio/mergify-engine/pull/2219

mergify_engine/actions/queue.py
```diff
@@ -46,9 +46,11 @@ class QueueAction(merge_base.MergeBaseAction):
...
             "merge", "squash", None
         ),
-        voluptuous.Required("merge_bot_account", default=None): voluptuous.Any(
-            None, types.GitHubLogin
+        voluptuous.Required("strict_method", default="merge"): voluptuous.Any(
+            "rebase", "merge"
         ),
+        voluptuous.Required("merge_bot_account", default=None): types.Jinja2WithNone,
+        voluptuous.Required("update_bot_account", default=None): types.Jinja2WithNone,
         voluptuous.Required("commit_message", default="default"): voluptuous.Any(
```
jd, line 54: doc update?

jd (changes requested):

sileht:
Do we keep `strict_method` ? Or do I rename it `update_method` ?

jd:
Sorry, I missed your comment in the first place. You're right I don't think `strict` means anything in that context.

My question would be, why does it apply only to the first PR? I imagine that if you set it to `rebase` you could build the merge cars by rebasing PR on top of each others?

PR #2224, opened 2021-03-17
fix(worker): convert more redis ConnectionError to StreamRetry

sileht:
This will avoid some useless Sentry reporting.


mergify_engine/worker.py
```diff
@@ -355,9 +361,15 @@ async def consume(self, stream_name: StreamNameType) -> None:
...
-        await self.redis_stream.eval(
-            self.ATOMIC_CLEAN_STREAM_SCRIPT, 1, stream_name.encode(), time.time()
-        )
+        try:
+            await self.redis_stream.eval(
+                self.ATOMIC_CLEAN_STREAM_SCRIPT, 1, stream_name.encode(), time.time()
+            )
+        except aredis.exceptions.ConnectionError:
+            LOG.warning(
+                "fail to cleanup stream, it maybe partially replayed",
```
jd, line 370:
```suggestion
                "fail to cleanup stream, it may be partially replayed",
```

jd (approved):

PR #2226, opened 2021-03-17
refactor(web_cli): use the new queues endpoint

CamClrt:


mergify_engine/web_cli.py
```diff
@@ -85,3 +85,16 @@ def queues():
             method="GET",
         )
     )
+
+
+def queues_v2():
```
jd, line 90: That's not being used though.
You might want to change the `queues` function directly then.

jd (changes requested):

mergify_engine/web_cli.py
```diff
@@ -81,7 +81,7 @@ def queues():
     args = parser.parse_args()
     asyncio.run(
         api_call(
-            config.BASE_URL + f"/queues/{args.owner_id}",
+            config.BASE_URL + f"/queues_v2/{args.owner_id}",
```
jd, line 84: if that is the only changes, I'm wondering if it's worth doing it at all

since you are then going to rename this `/queues_v2` as `/queues`, you'll get back to the same point…

jd (changes requested):

mergify_engine/web_cli.py
CamClrt: ok

PR #2228, opened 2021-03-17
fix: catch aredis.ConnectionError in monitoring task

sileht:


mergify_engine/worker.py
```diff
@@ -701,6 +701,8 @@ async def monitoring_task(self) -> None:
             except asyncio.CancelledError:
                 LOG.debug("monitoring task killed")
                 return
+            except aredis.ConnectionError:
+                LOG.warning("monitoring task lose the redis connection", exc_info=True)
```
jd, line 705:
```suggestion
                LOG.warning("monitoring task lost Redis connection", exc_info=True)
```

jd (approved):

PR #2268, opened 2021-03-23
refactor MRGFY 346

CamClrt:


CamClrt:
To be validated, it's necessary to check and merge this [PR](https://github.com/Mergifyio/mergify.io/pull/2044) before

sileht (dismissed):
Looks good, but you need to rebase it, The new `train` queues mechanism have been merged yesterday and it changes some stuffs there.

mergify_engine/tests/functional/test_engine.py
```diff
@@ -754,22 +754,6 @@ async def test_merge_smart_strict(self):
         await self.wait_for("pull_request", {"action": "synchronize"})
         await self.run_engine()
 
-        # with the older endpoint
```
sileht, line 757: We should keep that and just change the returned payload

sileht: You can also  complete the test in test_queue.py that query this API

mergify_engine/web/root.py
```diff
@@ -275,22 +274,9 @@ async def queues_by_owner_id(
     ] = collections.defaultdict(dict)
     async for queue in redis_cache.scan_iter(match=f"merge-queue~{owner_id}~*"):
         _, _, repo_id, branch = queue.split("~")
-        async with github.aget_client(owner_id=owner_id) as client:
```
sileht, line 278: Both endpoints must be exactly the same. And we should not duplicate the code.

sileht (changes requested):

sileht (approved):

PR #2269, opened 2021-03-23
chore(tests): drops pygithub

sileht:
Fixes MRGFY-357


jd:
I'd use `test: drop pygithub` as a title :)

mergify_engine/github_types.py
```diff
@@ -475,3 +475,15 @@ class GitHubEventTeam(GitHubEvent):
 class GitHubEventTeamAdd(GitHubEvent, total=False):
     # Repository key can be missing on Enterprise installations
     repository: GitHubRepository
+
+
+GitHubGitRefType = typing.NewType("GitHubGitRefType", str)
```
jd, line 480: This already exists AFAIK

jd (changes requested):

mergify_engine/github_types.py
sileht: `GitHubGitRefType` != `GitHubRefType`, its Git Database API vs Branches API.

And they should not be mixed as the first one has git namespace inside (eg: refs/heads/...)

jd (approved):

PR #2279, opened 2021-03-23
feat(metric): track aredis.ConnectionError

sileht:
Fixes MRGFY-363

### DIFF
--- a/mergify_engine/web/root.py
+++ b/mergify_engine/web/root.py
@@ -18,6 +18,7 @@ import typing
 
 import aredis
 import daiquiri
+from datadog import statsd
 import fastapi
 import httpx
 from starlette import requests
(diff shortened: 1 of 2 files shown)


mergify_engine/web/root.py
```diff
@@ -66,6 +67,7 @@ async def shutdown() -> None:
 async def redis_errors(
     request: requests.Request, exc: aredis.exceptions.ConnectionError
 ) -> responses.JSONResponse:
+    statsd.increment("engine.redis.connection-errors")
```
jd, line 70:
```suggestion
    statsd.increment("redis.client.connection.errors")
```
Then we can use this in every service and apply tags.

mergify_engine/worker.py
```diff
@@ -242,6 +242,9 @@ async def _translate_exception_to_retries(
         try:
             yield
         except Exception as e:
+            if isinstance(e, aredis.exceptions.ConnectionError):
+                statsd.increment("engine.redis.connection-errors")
```
jd, line 246: There's a lot of repetition, we might forget to add this somewhere. Can we wrap the `aredis` client to take care of that?

jd (dismissed):

mergify_engine/worker.py
jd: @sileht should I wait for this or you're giving up?

sileht: Oh, I miss it, but I don't see how to do that properly without involving some python black magic. ConnectionError can be raised everywhere in aredis code.

jd (approved):

PR #2290, opened 2021-03-24
docs(close): add an example

jd:


sileht (approved):
/me imagines the same rules with `files~=.*` :smiling_imp:

PR #2306, opened 2021-03-24
fix: don't use redis connection on startup

sileht:
Currently if the redis connection is not working
* The asyncio Worker._run() task exits with a never awaited
  `aredis.ConnectionError` exception, and Worker() just waits for
  `_shutdown()` to be called.
* The fastapi/starlette  web.root._startup() hook raises a
  `aredis.ConnectionError` exception and the web dyno is restarted.

Fixes MRGFY-339

### DIFF
--- a/mergify_engine/actions/refresh.py
+++ b/mergify_engine/actions/refresh.py
@@ -29,7 +29,7 @@ class RefreshAction(actions.Action):
     async def run(
         self, ctxt: context.Context, rule: rules.EvaluatedRule
     ) -> check_api.Result:
-        async with utils.aredis_for_stream() as redis_stream:
+        with utils.aredis_for_stream() as redis_stream:
             await github_events.send_refresh(
                 ctxt.redis,
                 redis_stream,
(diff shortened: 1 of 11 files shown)


jd (commented):
So we don't really care about client name?

sileht:
If we want to keep them, we need to block the worker and web dyno at startup until the connection is up.
Without the clientname set, the workers and web dyno can start even is ready is not there, they can handle requests (and returns 503 until it's up), workers can generate metrics about aredis issue until it's up.

jd (commented):
Makes sense.

Is there a way to have a hook being called or scheduled later after startup to set the client name?

Maybe a single task in asyncio?

sileht:
I just realize that using the command `client_setname` per client basic doesn't really make sense as only the first connection of the pool will get the `NAME` set. This client doesn't allow to set the name Client wide or Pool wide.

jd (approved):
Good catch then, let's forget about it.

PR #2320, opened 2021-03-26
fix(debug): check the url before processing it

CamClrt:


mergify_engine/debug.py
```diff
@@ -319,8 +320,27 @@ async def report(
     return client
 
 
+def _filter(args, func):
```
jd, line 323: My head. Ooch. It hurts.

If you want to validate and parse and URL, there are methods in Python stdlib to do that. No need to reinvent the wheel.

Use them!

jd (dismissed):

mergify_engine/debug.py
```diff
@@ -157,19 +158,19 @@ async def report(
 ) -> typing.Union[context.Context, github.AsyncGithubInstallationClient, None]:
     redis_cache = utils.create_aredis_for_cache(max_idle_time=0)
 
-    path = url.replace("https://github.com/", "")
+    path = urllib.parse.urlparse(url).path
```
jd, line 161: 👍

mergify_engine/debug.py
```diff
@@ -157,19 +158,19 @@ async def report(
...
 
     pull_number: typing.Optional[str]
     repo: typing.Optional[str]
 
     try:
-        owner, repo, _, pull_number = path.split("/")
+        _, owner, repo, _, pull_number = path.split("/")
```
jd, line 167: This code is stil very fragile IMHO:
see:
```python
>>> urllib.parse.urlparse("https://github.com//foobar/lol/pull/12").path.split("/")
['', '', 'foobar', 'lol', 'pull', '12']
```

it's already broken in that regard, but it might be good to enhance it!

mergify_engine/tests/unit/test_debug.py
```diff
@@ -0,0 +1,44 @@
...
+        ("https://github.com/mergifyio/mergify-engine/pull/123#", res),
+        ("https://github.com/mergifyio/mergify-engine/pull/123#", res),
+        ("https://github.com/mergifyio/mergify-engine/pull/123#123", res),
+        ("https://github.com/mergifyio/mergify-engine/pull/123?foo=123", res),
+        ("https://github.com/mergifyio/mergify-engine/pull/123?foo=123&bar=123", res),
+        ("https://github.com/mergifyio/mergify-engine/pull/123?foo", res),
+        ("https://github.com//mergifyio/mergify-engine/pull/123", res),
+        ("https://github.com/mergifyio//mergify-engine/pull/123", res),
+        ("https://github.com//mergifyio/mergify-engine//pull/123", res),
+        ("https://github.com//mergifyio/mergify-engine/pull//123", res),
```
jd, line 40: Add these too:
`"https://github.com//mergifyio/mergify-engine/123"`
`"https://github.com//mergifyio/mergify-engine/foobar/pull/123"`

jd (changes requested):

mergify_engine/tests/unit/test_debug.py
```diff
@@ -0,0 +1,44 @@
...
+# under the License.
+
+import typing
+
+import pytest
+
+from mergify_engine import debug
+
+
+res = ("mergifyio", "mergify-engine", "123")
```
jd, line 24: You only have one result here for all test cases, so no need to pass it as an argument to `test_parser`.

mergify_engine/debug.py
```diff
@@ -152,28 +153,37 @@ async def _get_config(
...
 
+    return owner, repo, pull_number
+
+
+async def report(
+    url: str,
+) -> typing.Union[context.Context, github.AsyncGithubInstallationClient, None]:
+    redis_cache = utils.create_aredis_for_cache(max_idle_time=0)
+
+    owner, repo, pull_number = _parser(url)
```
sileht, line 185: Since `_parser()` can raise `ValueError`, it should be caught here, to print error message and exit.

jd, line 185: You proved in your test that this can raise `ValueError` and you don't handle it here.

jd (changes requested):

mergify_engine/debug.py
sileht: Maybe _parser can have a better name

sileht (dismissed):

jd (changes requested):
Can you add the Sentry / Linear fix tags?

jd (approved):

jd:
@Mergifyio update

PR #2345, opened 2021-04-01
feat(commands): allow PR author to run commands

### LINKED ISSUE #1070: Clarify permissions required to run backport command
## Overview

We've recently started using Mergifyio to handle our backport and merge workflows.  In general things are working really well!  One issue we've noticed is that some users are unable to run the `backport` command and get an error message `[user] is not allowed to run commands`.

I wasn't able to find documentation anywhere on what permissions on the repo are required to allow a user to run the `backport` command.  Would it be possible to include this somewhere in the documentation?

Ideally for our workflow, anyone would be able to run `mergify backport` as that will just open a pull request, and we're using merge permissions to gate whether that pull request actually gets merged.

Some more details:
* We're an open source project where often end-users want bug fixes backported from `master` to a particular `release-*` branch.
* A small group of release managers have access to the `release-*` branches; we have mergify configured to only merge pull requests to these branches if approved by a release manager.
* Ideally anyone would be able to request a backport by running the `backport` command on a pull request against `master` to open a backport PR against a `release-*` branch.
  * Even though mergify blocks this, anyone with a Github account could manually do the backport and open the PR, so it's not clear to me the benefit of blocking PR creation.
* Once the backport PR is opened (whether manually or by mergify), branch protection rules and mergify config only allow it to be merged after approval by a release manager.

The only part of the above workflow that's not working is that most people can't request backports, so either need to comment on the PR tagging a release manager to ask them to run the command, or they need to manually do the backport and open the PR.  It's a much easier workflow for the small group of release managers if they can review PRs that are already open/ready and decide whether to merge, instead of being pinged in multiple channels to help with the backport process itself.

## Summary

We'd love if there was a way to allow any Github user to request a backport via the `backport` command, perhaps gated by a setting in the mergify config.

In the absence of that, it would be great to document what permissions are required to run that command, as I was unable to find that anywhere in the current documentation.

CamClrt:
Fixes MRGFY-44

Resolve #1070


sileht (changes requested):
Test ?

mergify_engine/engine/commands_runner.py
```diff
@@ -176,7 +176,10 @@ def log(comment_out: str, result: typing.Optional[check_api.Result] = None) -> N
     if (
         user
         and user["id"] != config.BOT_USER_ID
-        and not await ctxt.repository.has_write_permission(user)
+        and not (
+            await ctxt.repository.has_write_permission(user)
+            or user["id"] == ctxt.pull["user"]["id"]
+        )
```
jd, line 182: This is barely readable now, I'd reverse the `and` in `or`.
https://fr.wikipedia.org/wiki/Alg%C3%A8bre_de_Boole_(logique)

following https://fr.wikipedia.org/wiki/Lois_de_De_Morgan law
```python
       user
        and user["id"] != config.BOT_USER_ID
        and not (
            await ctxt.repository.has_write_permission(user)
            or user["id"] == ctxt.pull["user"]["id"]
        )
```
can be rewritten
```python
        not (
          user is not None
          or user["id"] == config.BOT_USER_ID
          or (await ctxt.repository.has_write_permission(user)
              or user["id"] == ctxt.pull["user"]["id"])
         )
```

since you only have `or` now you can remove the `()` which makes it simpler:
```python
        not (
          user is not None
          or user["id"] == config.BOT_USER_ID
          or await ctxt.repository.has_write_permission(user)
          or user["id"] == ctxt.pull["user"]["id"]
         )
```

now you have a list of conditions that makes it possible to run a command, with `or` around them, and just a `not` in front of those to negate the condition (and post the comment to say it's disallowed).

jd (changes requested):

mergify_engine/tests/unit/test_command.py
```diff
@@ -54,3 +60,162 @@ def test_command_loader():
...
+        },
+    )
+
+    with pytest.raises(RuntimeError) as error_msg:
+        await handle(ctxt=ctxt, comment="@Mergifyio update", user=None)
+    assert "user must be set if rerun is false" in str(error_msg.value)
+
+
+@pytest.mark.asyncio
+class Test_run_command_without_permissions(TestCase):
```
sileht, line 139: Why do you have a class here ?

CamClrt: I would like to check the logging message generated by `log(message)`. I saw I can use `assertLogs()` from TestCase.

jd: https://docs.pytest.org/en/stable/logging.html

sileht (changes requested):

mergify_engine/tests/unit/test_command.py
```diff
@@ -84,3 +92,178 @@ def test_command_loader_wuth_defaults():
...
+            "login": "wall-e",
+            "type": "Bot",
+            "avatar_url": "https://avatars.githubusercontent.com/u/583231?v=4",
+        },
+    )
+
+    http_calls = []
+
+    async def mock_post(*args, **kwargs):
+        http_calls.append("http_call")
```
sileht, line 260: You should put *args and **kwargs in the list of calls, and at the bottom assert on all arguments passed, to be sure this is the call you really want catch.

mergify_engine/tests/unit/test_command.py
```diff
@@ -84,3 +91,267 @@ def test_command_loader_wuth_defaults():
...
+            "login": "wall-e",
+            "type": "Bot",
+            "avatar_url": "https://avatars.githubusercontent.com/u/583231?v=4",
+        },
+    )
+
+    http_calls = []
+
+    async def mock_post(*args, **kwargs):
+        http_calls.append(args)
```
sileht, line 346: It's recommended that each item of a list have the same kind of object, here we have one time the args and one time the kwargs, it's a bit confusing, you can use a tuple so each list item is a post request, eg:

```
http_calls.append((args, kwargs))
```

mergify_engine/tests/unit/test_command.py
```diff
@@ -84,3 +92,602 @@ def test_command_loader_wuth_defaults():
...
+            "type": "Bot",
+            "avatar_url": "https://avatars.githubusercontent.com/u/583231?v=4",
+        },
+    )
+
+    http_calls = []
+
+    async def mock_post(*args, **kwargs):
+        http_calls.append((args, kwargs))
+        return MockResponseWithoutPermissions()
```
sileht, line 353: Here you should return something that looks like an httpx.Response()

mergify_engine/tests/unit/test_command.py
```diff
@@ -84,3 +92,602 @@ def test_command_loader_wuth_defaults():
...
+
+    monkeypatch.setattr(client, "post", mock_post)
+
+    permission_calls = []
+
+    async def mock_permission(*args, **kwargs):
+        permission_calls.append((args, kwargs))
+        return MockResponseWithoutPermissions()
```
sileht, line 361: If you change has_write_permission(), you should return the same thing, here a bool.

More generally to make this test more robust you should only monkeypatch/mock external stuff, here the HTTP client.
This will allow us to also test the logic inside has_write_permission.

unittest should only replaces thing that need to access to external resources (here GitHub API calls).

sileht (changes requested):

mergify_engine/tests/unit/test_command.py
```diff
@@ -84,3 +92,271 @@ def test_command_loader_wuth_defaults():
         "labels": [],
         "label_conflicts": "conflicts",
     }
+
+
+@pytest.mark.asyncio
+async def test_run_command_without_rerun_and_without_user(redis_cache):
```
sileht, line 98: You can remove this test, it doesn't add much value.

mergify_engine/tests/unit/test_command.py
```diff
@@ -84,3 +92,271 @@ def test_command_loader_wuth_defaults():
...
+
+    client = mock.MagicMock()
+    client.auth.installation.__getitem__.return_value = 123
+    sub = subscription.Subscription(
```
sileht, line 102: I think you can also merge the rerun test with those you have parametized. Or at least create a helper to build the context since they are always the same.

mergify_engine/tests/unit/test_command.py
```diff
@@ -84,3 +92,271 @@ def test_command_loader_wuth_defaults():
...
+        (
+            1,
+            "nothing",
+            "Sorry but I didn't understand the command",
+        ),
+        (
+            666,
+            "admin",
+            "Sorry but I didn't understand the command",
+        ),
```
sileht, line 263: Maybe test also if the author has `write` permission.

sileht (changes requested):
Almost perfect.

sileht (approved):

PR #2348, opened 2021-04-01
ci: add semantic pull request GitHub actions

sileht:


sileht:
The GitHub action is not as good as the ProBot App. It doesn't support linting commit, soo...

PR #2370, opened 2021-04-07
feat: make commands pick defaults from configuration

sileht:
This make command uses defaults from configuration.

Fixes MRGFY-146


jd (dismissed):
need rebase

jd (approved):

PR #2371, opened 2021-04-07
feat: allow to use null instead of {} in action config

sileht:
Allow to use null instead of {}, when we have no option to pass.

Fixes MRGFY-378


jd (dismissed):
We should update the doc/example at the same time.

jd (approved):

PR #2380, opened 2021-04-08
copy fix and command

sileht:
## fix(copy): delete temporary branches

copy branches starts by mergify/copy. We only clean mergify/bp branches.

This change also cleanup copy branches

Fixes MRGFY-385

## feat: allow copy command

Fixes MRGFY-384


sileht:
I have already set `MERGIFYENGINE_BOT_USER_LOGIN` in heroku

mergify_engine/data/default_pull_request_rules.yml
```diff
@@ -1,9 +1 @@
-rules:
-  - name: delete backport branch (Mergify rule)
-    hidden: true
-    conditions:
-      - author=mergify[bot]
-      - head~=^mergify/bp/
-      - closed
-    actions:
-      delete_head_branch: {}
+
```
jd, line 1: if the file is empty we could probably just remove it

jd (changes requested):

jd (approved):

sileht:
@mergifyio update

sileht:
@Mergifyio update

PR #2381, opened 2021-04-08
fix: allow to set default for queue name

sileht:
This makes defaults working with `Required` config key, like the queue
name.

### DIFF
--- a/mergify_engine/rules/__init__.py
+++ b/mergify_engine/rules/__init__.py
@@ -309,22 +309,26 @@ def YAML(v: bytes) -> typing.Any:
         raise YAMLInvalid(message="Invalid YAML", error_message=error_message)
 
 
-PullRequestRulesSchema = voluptuous.All(
-    [
-        voluptuous.All(
-            {
-                voluptuous.Required("name"): str,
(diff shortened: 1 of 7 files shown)


sileht:
Depends on https://github.com/Mergifyio/mergify-engine/pull/2380

docs/source/commands.rst
```diff
@@ -80,6 +80,30 @@ Runs the :ref:`backport action` action.
    If the repository is bigger than 512 MB, the ``backport`` action is only
    available for `Essential and Premium Plan subscribers <https://mergify.io/pricing>`_.
 
+.. _copy command:
+
+copy
```
jd, line 85: That does not sound related to the PR?

jd (changes requested):

mergify_engine/rules/__init__.py
```diff
@@ -365,6 +374,7 @@ def FullifyPullRequestRules(v):
     return v
 
 
+<<<<<<< HEAD
```
jd, line 377: lol

jd (dismissed):

jd (approved):

PR #2383, opened 2021-04-08
chore: improve the user-agent string

sileht:
Fixes MRGFY-387

### DIFF
--- a/mergify_engine/clients/http.py
+++ b/mergify_engine/clients/http.py
@@ -14,9 +14,10 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-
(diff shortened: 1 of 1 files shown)


mergify_engine/clients/http.py
```diff
@@ -26,10 +27,16 @@
 
 LOG = daiquiri.getLogger(__name__)
 
+ENGINE_VERSION = os.getenv("HEROKU_SLUG_COMMIT", "dev")[:7]
```
jd, line 30: `HEROKU_RELEASE_VERSION`?

jd (approved):

PR #2388, opened 2021-04-09
docs: move admonitions as float right

jd:


jd:
Screenshot 2021-04-09 at 13 16 05

sileht (approved):

PR #2395, opened 2021-04-10
chore: informs users about bot_account removal

sileht:
Fixes MRGFY-255

### DIFF
--- a/mergify_engine/actions/merge_base.py
+++ b/mergify_engine/actions/merge_base.py
@@ -51,6 +51,16 @@ FORBIDDEN_MERGE_COMMITS_MSG = "Merge commits are not allowed on this repository.
 FORBIDDEN_SQUASH_MERGE_MSG = "Squash merges are not allowed on this repository."
 FORBIDDEN_REBASE_MERGE_MSG = "Rebase merges are not allowed on this repository."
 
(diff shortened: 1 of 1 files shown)


mergify_engine/actions/merge_base.py
```diff
@@ -51,6 +51,16 @@
...
+BOT_ACCOUNT_DEPRECATION_NOTICE = """This pull request has been merged with an
+unsupported configuration option called `bot_account`.
+
+This configuration option will be ignored starting at may, 1 2021, and removed
+on the june 1, 2021.
+
+It can be replaced by `update_bot_account`, `merge_bot_account` or both
+depending on your use-case (https://docs.mergify.io/actions/merge/).
+"""
+
```
jd, line 63:
```suggestion
BOT_ACCOUNT_DEPRECATION_NOTICE = """This pull request has been merged with the
unsupported configuration option `bot_account`.

This option will be ignored starting May 1st, 2021, and removed
on June 1st, 2021.

This option can be replaced by `update_bot_account`, `merge_bot_account` or both
depending on your use-case (https://docs.mergify.io/actions/merge/).
"""

```

jd (changes requested):

jd (approved):

PR #2396, opened 2021-04-11
fix(conditions): add cleaner condition examples

Nytelife26:
## Description of the change

I added cleaner examples of conditions for regex matching and the all-checks validation problem, as well as correcting some typos.

### DIFF
--- a/docs/source/conditions.rst
+++ b/docs/source/conditions.rst
@@ -309,6 +309,20 @@ For example, to automatically merge a pull request if its author is ``foo`` or
           merge:
             method: merge
 
+Alternatively, if you wish to use a cleaner but less explicit solution, it
+may be prevalent to use regex matching:
+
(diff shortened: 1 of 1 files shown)


docs/source/conditions.rst
```diff
@@ -373,6 +387,16 @@ something like:
        - check-success=build: Windows
        - check-success=build: Linux
 
+Alternatively, you may use the following:
+
+.. code-block:: yaml
+
+     conditions:
+       - "#check-failure=0"
+       - "#check-neutral=0"
```
sileht, line 396: This is not an alternative, as explained above and below this kind of condition should not be used.

`- #check-XXX=0` is always true just after a pull request is open or synchronized (force-push or new commits)

All required checks must be provided.

sileht (changes requested):

PR #2410, opened 2021-04-13
fix: pull_request_rules is now optional

sileht:
Fixes MERGIFY-ENGINE-25H


jd (approved):
did you mean "is optional" in the title?

jd:
Might be worth adding a test if possible.

Issue #2414, opened 2021-04-13
Append the original description to the backported and copied PRs


v1v:
## What

Enable to append the original description in the backported and copioed PRs.

### Proposal

Extend the `backport` action with a new option, f.i: `append_original_description` (or something else).

See the below snippet as an example of how to configure mergify with it:

```
pull_request_rules:
  - name: backport patches to stable branch
    conditions:
      - base=master
      - label=backport-to-stable
    actions:
      backport:
        append_original_description: True
        branches:
          - stable
```

Thanks

AraHaan:
Should this be closed @jd?

jd:
I think so

PR #2415, opened 2021-04-13
feat(copy): append original description

v1v:
## Description of the change

Append original description to the copy and backport commands

## Related issues

Fix https://github.com/Mergifyio/mergify-engine/issues/2414

### Development

- [ ] All checks must pass (Semantic Pull Request, pep8, requirements, unit tests, functional tests, security checks, …)
- [ ] The code changed/added as part of this pull request must be covered with tests
- [ ] Hotfixes must have a link to Sentry issue and the ``hotfix`` label
- [ ] Features must have a link to a Linear task

### Code Review


mergify_engine/duplicate_pull.py
```diff
@@ -319,6 +321,9 @@ async def duplicate(
             "checking-out-pull-requests-locally/"
         )
 
+    if append_original_description and not ctxt.pull["body"] is None:
```
v1v, line 324: I quickly submitted my draft PR but it's failing when running the flake8 with

```
TypedDict "GitHubPullRequest" has no key 'body'  [misc]
```

I somehow though the `body` attribute was a correct one. I'll need to debug further

mergify_engine/actions/copy.py
```diff
@@ -55,6 +55,7 @@ def get_config_schema(
             voluptuous.Required("branches", default=[]): [str],
             voluptuous.Required("regexes", default=[]): [voluptuous.Coerce(Regex)],
             voluptuous.Required("ignore_conflicts", default=True): bool,
+            voluptuous.Required("append_original_description", default=False): bool,
```
sileht, line 58: I got the idea, but to make it more generic like the title, maybe the new attribute should just be:

```
voluptuous.Required("body_suffix", default=""): types.Jinja2)
```

And the configuration will look like:

```
backport:
    body_suffix: "{{ body }}"
```

WDYT @jd ?

jd: What about having the whole body being customizable?
Where you could reuse part of the original pull request like `{{ body }}` in it and maybe some other data such as `{{ git_cherry_pick_output }}` etc?

mergify_engine/duplicate_pull.py
jd: `body` is valid but it's just missing from the type definition in `github_types`.
Those definitions are not always complete (we're lazy). You can double-check GitHub API if you doubt.

mergify_engine/actions/copy.py
sileht: We can do that too, `rendering_template()` already support additional variables

sileht:
Replaced by https://github.com/Mergifyio/mergify-engine/pull/2479

PR #2417, opened 2021-04-14
fix: better messages with no rules setuped

sileht:
Even if no rules are setuped, we still post "Be patient, the page will be
updated soon".

Since we can't known during the event ingestion if rules are defined or
now (we don't do any GitHub API calls except this temporary message), we
must always update it.

This changes always update the summary even for people that doesn't have
rules setuped.


sileht:
I will add one functional test for people without rules defined.

jd (approved):

PR #2424, opened 2021-04-15
feat: report team resolution errors

sileht:
This introduces a new mechanism to track conditions that may raise error
during its evaluation.

This change a bit the behavior of the engine. Before team resolution was
just ignored, and the rule can never ever match. Now we will report the
issue in the Summary.

Fixes MRGFY-137


sileht:
image

mergify_engine/rules/live_resolvers.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+async def teams(
+    ctxt: context.Context,
+    values: typing.Optional[typing.Union[typing.List[str], typing.Tuple[str], str]],
+) -> typing.List[github_types.GitHubLogin]:
+    if not values:
+        return []
+    if not isinstance(values, (list, tuple)):
+        values = [values]
```
jd, line 77: out of the scope of this PR but this really should go away

mergify_engine/tests/functional/test_engine.py
```diff
@@ -935,77 +964,63 @@ async def test_short_teams(self):
...
+
+        with self.assertRaises(live_resolvers.LiveResolutionFailure):
+            await live_resolvers.teams(ctxt, ["@invalid/team/break-here"])
+
+        summary = [
+            c for c in await ctxt.pull_engine_check_runs if c["name"] == "Summary"
+        ][0]
+        assert summary["output"]["title"] == "2 faulty rules and 2 potential rules"
+        for message in (
+            "Team `@mergifyio-testing/noexists` does not exists",
```
jd, line 1019:
```suggestion
            "Team `@mergifyio-testing/noexists` does not exist",
```

mergify_engine/tests/functional/test_engine.py
```diff
@@ -935,77 +964,63 @@ async def test_short_teams(self):
...
+            "Team `@another-org/testing` is not part of the organization",
```
jd, line 1020: Maybe say which organization if you can?

jd (changes requested):
Awesome work!

jd (approved):

PR #2451, opened 2021-04-19
fix: issue when STREAM_URL is not provided

sileht:


mergify_engine/config.py
```diff
@@ -211,8 +211,8 @@ def AccountTokens(v: str) -> typing.Dict[str, str]:
 
 globals().update(Schema(CONFIG))
 
-if CONFIG["STREAM_URL"] is None:
-    STREAM_URL = CONFIG["STORAGE_URL"]
+if globals()["STREAM_URL"] is None:
```
jd, line 214: Why not using `config.get` or `in`?

jd (changes requested):
wrong title :)

mergify_engine/config.py
sileht: `STREAM_URL` is always set in `globals()` thanks to voluptuous and the L212

jd (approved):

PR #2463, opened 2021-04-22
feat: add signaling mechanism

sileht:
This introduce a signal mechanism with an example
plugin for datadog.

The new datadog metric "engine.signals.actions.count"

* "engine.actions.count" count when action is ran by rule but it count
  too much thing:
   * some actions run on each batch until they unmatch
   * some other return PENDING until they can complete and get counter
     multiple times.
   * the action may run but do nothing (labels already set, pull
     already up2date or merged)
* "engine.command.count" count when an action is ran as a command

"engine.signals.actions.count" will track action run by rules and commands
but when they really do something on user pull requests.

### DIFF
--- a/mergify_engine/actions/merge_base.py
+++ b/mergify_engine/actions/merge_base.py
@@ -30,6 +30,7 @@ from mergify_engine import context
 from mergify_engine import json as mergify_json
 from mergify_engine import queue
 from mergify_engine import rules
+from mergify_engine import signals
 from mergify_engine import subscription
 from mergify_engine import utils
 from mergify_engine.clients import http
@@ -517,6 +518,7 @@ class MergeBaseAction(actions.Action):
             else:
                 return await self._handle_merge_error(e, ctxt, rule, q)
         else:
+            await signals.send(ctxt, "action.merge")
             await ctxt.update()
             ctxt.log.info("merged")
             if self.config[
--- a/mergify_engine/actions/post_check.py
+++ b/mergify_engine/actions/post_check.py
@@ -20,6 +20,7 @@ from mergify_engine import actions
 from mergify_engine import check_api
 from mergify_engine import context
 from mergify_engine import rules
+from mergify_engine import signals
 from mergify_engine import subscription
 from mergify_engine.rules import types
 
@@ -98,6 +99,7 @@ class PostCheckAction(actions.Action):
                 str(rmf),
             )
 
+        await signals.send(ctxt, "action.post_check")
         if rule.missing_conditions:
             return check_api.Result(check_api.Conclusion.FAILURE, title, summary)
         else:
--- a/setup.py
+++ b/setup.py
@@ -18,5 +18,5 @@ import setuptools
 
 
 setuptools.setup(
-    setup_requires=["setuptools>=30.3.0"],
+    setup_requires=["setuptools>=40.1.0"],
 )
--- a/mergify_engine/actions/assign.py
+++ b/mergify_engine/actions/assign.py
@@ -22,6 +22,7 @@ from mergify_engine import actions
 from mergify_engine import check_api
 from mergify_engine import context
 from mergify_engine import rules
+from mergify_engine import signals
 from mergify_engine.clients import http
 from mergify_engine.rules import types
 
@@ -73,6 +74,7 @@ class AssignAction(actions.Action):
                     f"GitHub error: [{e.status_code}] `{e.message}`",
                 )
 
+            await signals.send(ctxt, "action.assign.added")
             return check_api.Result(
                 check_api.Conclusion.SUCCESS,
                 "Assignees added",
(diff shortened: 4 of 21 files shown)


setup.py
```diff
@@ -18,5 +18,5 @@
 
 
 setuptools.setup(
-    setup_requires=["setuptools>=30.3.0"],
+    setup_requires=["setuptools>=40.1.0"],
```
jd, line 21: Why?

mergify_engine_hooks/datadog_hook/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+from mergify_engine import github_types
+from mergify_engine import hooks
+
+
+class Hook(hooks.HookBase):
+    async def __call__(
+        self, account: github_types.GitHubAccount, event: hooks.EventName
+    ) -> None:
+        if event.startswith("action"):
+            statsd.increment("engine.hooks.action.count", tags=[f"event:{event[7:]}"])
```
jd, line 26: 1. This looks different than what we currently have. Is this better/necessary? Should we only have one metric?
2. I'd remove `hooks` from the name. The mechanism used to gather the metric is not useful in the name of the metric I'd say 🤔

mergify_engine/hooks.py
```diff
@@ -0,0 +1,81 @@
+# -*- encoding: utf-8 -*-
```
jd, line 1: I'd rather call that `signal` than `hooks` since it's mostly a one way message being sent.

mergify_engine/actions/post_check.py
```diff
@@ -98,6 +99,7 @@ async def _post(
                 str(rmf),
             )
 
+        await hooks.send(ctxt, "action.post_check")
```
jd, line 102: This tracks failure+success but some hooks track only success.
Do we want to make a difference?

jd (dismissed):

setup.py
sileht: setuptools need 40.1.0 to support native namespace.

mergify_engine/actions/post_check.py
sileht: I track `the action did something on the pull request`

mergify_engine_hooks/datadog_hook/__init__.py
sileht: I describe in the commit message what we are currently tracking and what this track, this is not the same. We need both I would say. Customer doesn't need to implementation details of the engine. While we want it for the monitoring.

docs/mergify_engine_gendoc.py
```diff
@@ -63,11 +63,13 @@ def write_redirect(app, pagename, templatename, context, doctree):
     url = f"https://docs.mergify.io/{pagename}/"
     os.makedirs(os.path.join(app.outdir, pagename), exist_ok=True)
     with open(os.path.join(app.outdir, pagename + ".html"), "w") as f:
-        f.write(f"""<head>
+        f.write(
```
jd, line 66: ?

mergify_engine/actions/merge_base.py
```diff
@@ -717,7 +719,7 @@ async def merge_report(
 <br />
 This pull request must be merged manually."""
 
-        # NOTE(sileht): remaining state "behind, clean, unstable, has_hooks
+        # NOTE(sileht): remaining state "behind, clean, unstable, has_signals
```
jd, line 722: does not look legit

mergify_engine/engine/__init__.py
```diff
@@ -264,7 +264,7 @@ async def run(
     await _ensure_summary_on_head_sha(ctxt)
 
     # NOTE(jd): that's fine for now, but I wonder if we wouldn't need a higher abstraction
-    # to have such things run properly. Like hooks based on events that you could
+    # to have such things run properly. Like signals based on events that you could
```
jd, line 267: nop

mergify_engine/github_types.py
```diff
@@ -157,7 +157,7 @@ class GitHubIssue(GitHubIssueOrPullRequest):
 # unstable: branch up2date (if strict: True) and not required status
 #           checks are failure or pending
 # clean: branch up2date (if strict: True) and all status check OK
-# has_hooks: Mergeable with passing commit status and pre-recieve hooks.
+# has_signals: Mergeable with passing commit status and pre-recieve signals.
```
jd, line 160: nop

mergify_engine/github_types.py
```diff
@@ -169,7 +169,7 @@ class GitHubIssue(GitHubIssueOrPullRequest):
     "behind",
     "unstable",
     "clean",
-    "has_hooks",
+    "has_signals",
```
jd, line 172: nop

mergify_engine/github_types.py
```diff
@@ -202,7 +202,7 @@ class GitHubPullRequest(GitHubIssueOrPullRequest):
     commits: int
 
 
-# https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads
+# https://docs.github.com/en/free-pro-team@latest/developers/websignals-and-events/webhook-events-and-payloads
```
jd, line 205: nop

jd (changes requested):
I think it misses the `mergify_engine_signal` module?

sileht:
Eheh, my sed was a bit too broad

jd (approved):

PR #2481, opened 2021-04-27
refactor(bot_account): do bot_account check the same everywhere

sileht:
This creates an helper function to validate the bot_account


mergify_engine/actions/utils.py
```diff
@@ -0,0 +1,72 @@
...
+import dataclasses
+import typing
+
+from mergify_engine import check_api
+from mergify_engine import context
+from mergify_engine import subscription
+
+
+@dataclasses.dataclass
+class MissingPermission(Exception):
```
jd, line 26: unused?

mergify_engine/actions/utils.py
```diff
@@ -0,0 +1,72 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2020 Mergify SAS
```
jd, line 3:
```suggestion
# Copyright © 2021 Mergify SAS
```

jd (changes requested):

jd (approved):

PR #2494, opened 2021-04-30
feat(rebase): pick the author token if possible

sileht:
If we have the author token pick it instead of a random one.


mergify_engine/branch_updater.py
```diff
@@ -23,6 +23,7 @@
 from mergify_engine import config
 from mergify_engine import context
 from mergify_engine import exceptions
+from mergify_engine import github_types
```
jd, line 26: ?

jd (dismissed):

jd (approved):

PR #2496, opened 2021-04-30
feat(git): leverage new user tokens info

sileht:
This change leverages new user tokens information
to configure Git during rebase/backport/copy.

Fixes MRGFY-421


mergify_engine/gitter.py
```diff
@@ -83,9 +83,18 @@ async def cleanup(self) -> None:
...
 
-    async def configure(self) -> None:
-        await self("config", "user.name", f"{config.CONTEXT}-bot")
-        await self("config", "user.email", config.GIT_EMAIL)
+    async def configure(
+        self, name: typing.Optional[str] = None, email: typing.Optional[str] = None
+    ) -> None:
+        effective_name = name
+        if effective_name is None:
+            effective_name = f"{config.CONTEXT}-bot"
```
jd, line 91:
```suggestion
        if name is None:
            name = f"{config.CONTEXT}-bot"
```
looks like it's a bit too complicated now, no? 😁 

same below

jd (changes requested):

mergify_engine/gitter.py
sileht: I should stop doing JS :) I though it was enforced by mypy, but that eslint-airbnb doesn't allow changing function parameters.

jd (approved):

PR #2500, opened 2021-05-03
refactor(merge/queue): process branch protection conditions earlier

sileht:
The current way to handle the branch protection is a bit hacky, we have
to take care of them in different places, making the code fragile, while
it's just a bunch of additional checks-success.

In QueueRule, we just add the special Branch Protection conditions to
the list of conditions to evaluate in the fly.

This change to the same thing for pull request Rules. For each rule is
split the rule into two rules, one with branch protection conditions and
the other without.

This makes also easier to have the same rendering everwhere when we
print the conditions.

Fixes MRGFY-428


sileht:
https://github.com/mergifyio-testing/functional-testing-repo/pull/10569/checks

jd (approved):

PR #2573, opened 2021-05-11
feat(merge): always run cancel

sileht:
This change will always run cancel() of merge action.
In case where we lose the summary and something is queued,
it can be automatically removed from the queue by close or refresh
events.


sileht:
replaced by https://github.com/Mergifyio/mergify-engine/pull/2600

PR #2600, opened 2021-05-12
fix(summary): reintroduce getting summary from synchronize event

sileht:
If a queued pull request didn't get events for more than 1 month, and
suddenly get a synchronize event we can lose the state of the PR.

This change restore a old mechanism that get the summary sha information
via synchronize event, but use it at fallback only.

### DIFF
--- a/mergify_engine/engine/__init__.py
+++ b/mergify_engine/engine/__init__.py
@@ -128,6 +128,38 @@ async def _get_summary_from_sha(ctxt, sha):
         return checks[0]
 
 
+async def _get_summary_from_synchronize_event(ctxt):
+    synchronize_events = {
+        s["data"]["after"]: s["data"]
+        for s in ctxt.sources
+        if s["event_type"] == "pull_request"
+        and s["data"]["action"] == "synchronize"
+        and "after" in s["data"]
+    }
+    if synchronize_events:
+        ctxt.log.warning("get summary from synchronize events")
+
+        # NOTE(sileht): We sometimes got multiple synchronize events in a row, that's not
+        # always the last one that has the Summary, so we also look in older ones if
+        # necessary.
+        after_sha = ctxt.pull["head"]["sha"]
+        while synchronize_events:
+            sync_event = synchronize_events.pop(after_sha, None)
+            if sync_event:
+                previous_summary = await _get_summary_from_sha(
+                    ctxt, sync_event["before"]
+                )
+                if previous_summary and actions_runner.load_conclusions_line(
+                    ctxt, previous_summary
+                ):
+                    return previous_summary
+
+                after_sha = sync_event["before"]
+            else:
+                ctxt.log.warning("summary from synchronize events not found")
+                break
+
+
 async def _ensure_summary_on_head_sha(ctxt: context.Context) -> None:
     for check in await ctxt.pull_engine_check_runs:
         if check["name"] == ctxt.SUMMARY_NAME and actions_runner.load_conclusions_line(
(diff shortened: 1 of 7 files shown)


mergify_engine/engine/__init__.py
```diff
@@ -136,6 +168,7 @@ async def _ensure_summary_on_head_sha(ctxt: context.Context) -> None:
             return
 
     opened = ctxt.has_been_opened()
+    synchronized = ctxt.has_been_synchronized()
```
jd, line 171: might not be worth computing this value here if it's not used later because the function returns early or because `previous_summary` is `None`.

jd (changes requested):

mergify_engine/engine/__init__.py
```diff
@@ -145,6 +177,14 @@ async def _ensure_summary_on_head_sha(ctxt: context.Context) -> None:
         return
 
     previous_summary = await _get_summary_from_sha(ctxt, sha)
+
+    if previous_summary is None:
+        # NOTE(sileht): If the cached summary sha expires and the next event we got for
+        # a pull request is "synchronize" we will loose the summary. Most of the times
+        # it's not a big deal, but if the pull request in queued for merge, it may
+        # be stuck.
```
jd, line 185:
```suggestion
        # NOTE(sileht): If the cached summary sha expires and the next event we got for
        # a pull request is "synchronize" we will lose the summary. Most of the times
        # it's not a big deal, but if the pull request is queued for merge, it may
        # be stuck.
```

mergify_engine/engine/__init__.py
```diff
@@ -128,6 +128,38 @@ async def _get_summary_from_sha(ctxt, sha):
...
+        if s["event_type"] == "pull_request"
+        and s["data"]["action"] == "synchronize"
+        and "after" in s["data"]
+    }
+    if synchronize_events:
+        ctxt.log.warning("get summary from synchronize event")
+
+        # NOTE(sileht): We sometimes got multiple synchronize events in a row, that's not
+        # always the last one that have the Summary, so we also looks in older ones if
+        # necessary.
```
jd, line 144:
```suggestion
        # NOTE(sileht): We sometimes got multiple synchronize events in a row, that's not
        # always the last one that has the Summary, so we also look in older ones if
        # necessary.
```

mergify_engine/engine/__init__.py
```diff
@@ -128,6 +128,38 @@ async def _get_summary_from_sha(ctxt, sha):
...
+                    ctxt, sync_event["before"]
+                )
+                if previous_summary and actions_runner.load_conclusions_line(
+                    ctxt, previous_summary
+                ):
+                    return previous_summary
+
+                after_sha = sync_event["before"]
+            else:
+                ctxt.log.warning("didn't find symmary from synchronize events")
```
jd, line 159:
```suggestion
                ctxt.log.warning("summary from synchronize events not found")
```

mergify_engine/engine/__init__.py
```diff
@@ -128,6 +128,38 @@ async def _get_summary_from_sha(ctxt, sha):
...
+async def _get_summary_from_synchronize_event(ctxt):
+    synchronize_events = {
+        s["data"]["after"]: s["data"]
+        for s in ctxt.sources
+        if s["event_type"] == "pull_request"
+        and s["data"]["action"] == "synchronize"
+        and "after" in s["data"]
+    }
+    if synchronize_events:
+        ctxt.log.warning("get summary from synchronize event")
```
jd, line 140:
```suggestion
        ctxt.log.warning("get summary from synchronize events")
```

jd (changes requested):

jd (approved):

PR #2627, opened 2021-05-19
feat(actions): meter engine action run time

jd:
This should allow to see which action are taking too much time and see the
impact of optimizations.


sileht (approved):

sileht:
My bet, backport/rebase/merge(when rebase occurs) will be the winners

PR #2633, opened 2021-05-19
refactor: keep all repository data

sileht:
This change keeps the repository data provided by GitHub.

This will be used by the new timed conditions to be able to send a
refresh event.


mergify_engine/context.py
```diff
@@ -208,7 +219,7 @@ class Repository(object):
     @property
     def base_url(self) -> str:
         """The URL prefix to make GitHub request."""
-        return f"/repos/{self.installation.owner_login}/{self.name}"
+        return f"/repos/{self.installation.owner_login}/{self.repo['name']}"
```
jd, line 222: Wouldn't be better to return `self.repo["url"]` at this point?

mergify_engine/tests/functional/actions/test_review.py
```diff
@@ -47,9 +47,10 @@ async def test_review(self):
 
         p, _ = await self.create_pr()
         await self.run_engine()
-
         await self.wait_for("pull_request_review", {}),
+
         await self.run_engine()
+        await self.wait_for("pull_request_review", {}),
```
jd, line 53: sound unrelated, maybe in a different PR? 😁

jd (approved):

mergify_engine/tests/functional/actions/test_review.py
sileht: record was failing sometime due to this.

jd: I understood that, which is exactly my point. This should be fixed in a different PR. 😉

PR #2643, opened 2021-05-21
fix(tests): skip workflow tests on GHE

sileht:
Currently for GHE some tests are run with a PAT instead of OAUTH token.

Until we have something to generate oauth_token without the dashboard,
disable tests that can't work with a PAT token.

Fixes MRGFY-460

### DIFF
--- a/mergify_engine/tests/functional/actions/test_merge.py
+++ b/mergify_engine/tests/functional/actions/test_merge.py
@@ -16,6 +16,7 @@
 import datetime
 import logging
 
+import pytest
 import yaml
 
 from mergify_engine import config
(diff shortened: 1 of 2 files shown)


mergify_engine/tests/functional/actions/test_merge.py
```diff
@@ -252,6 +253,12 @@ async def test_merge_rule_switch(self):
         pulls_in_queue = await q.get_pulls()
         assert pulls_in_queue == [p2["number"], p1["number"]]
 
+    # FIXME(sileht): Provide a tools to generate oauth_token without
+    # the need if the dashboard
```
jd, line 257:
```suggestion
    # the need of the dashboard
```

mergify_engine/tests/functional/actions/test_queue.py
```diff
@@ -1336,6 +1336,12 @@ async def test_queue_no_tmp_pull_request(self):
         p1 = await self.get_pull(p1["number"])
         assert p1["merged"]
 
+    # FIXME(sileht): Provide a tools to generate oauth_token without
+    # the need if the dashboard
```
jd, line 1340:
```suggestion
    # the need of the dashboard
```

jd (approved):

PR #2646, opened 2021-05-21
feat(filter): add back multi operator

sileht:


mergify_engine/rules/filter.py
```diff
@@ -115,6 +117,11 @@ class Filter:
         "~=": (lambda a, b: a is not None and b.search(a), any, re.compile),
     }
 
+    multiple_operators: typing.ClassVar[typing.Dict[str, MultipleOperatorT]] = {
+        "or": operator.or_,
+        "and": operator.and_,
```
jd, line 122: This is the `|` and `&` operators, which is not going to do a lazy evaluation of all conditions.

You want to use `any` and `all` instead.

And you need to make sure each element to `any` and `all` is evaluated with something like:

```
return multiple_op(await element(values) for element in elements)
```

Which will stop (for `any`) as soon as one is true.

mergify_engine/rules/filter.py
```diff
@@ -204,51 +211,76 @@ def build_evaluator(
...
-                raise InvalidArguments(str(e))
+                try:
+                    binary_operator = self.binary_operators[operator_name]
+                except KeyError:
+                    raise UnknownOperator(operator_name)
+                nodes = typing.cast(TreeBinaryLeafT, nodes)
+                return self._handle_binary_op(binary_operator, nodes)
+            nodes = typing.cast(TreeT, nodes)
+            return self._handle_unary_op(unary_operator, nodes)
+        nodes = typing.cast(typing.List[TreeT], nodes)
```
jd, line 230: I feel that there's a check missing here (and probably a test too).
This is casting blindly, but nodes could be something else than a `list` and then the code would fail.

jd (dismissed):

mergify_engine/rules/filter.py
sileht: I miss this one should be Iterator

mergify_engine/rules/filter.py
sileht: unfortunately, any() doesn't support async generator :(

jd (changes requested):

mergify_engine/rules/filter.py
```diff
@@ -201,51 +209,76 @@ def build_evaluator(
...
 
         return _unary_op
+
+    def _handle_multiple_op(
+        self, multiple_op: MultipleOperatorT, nodes: typing.Iterable[TreeT]
+    ) -> typing.Callable[[GetAttrObjectT], typing.Awaitable[bool]]:
+        elements = [self.build_evaluator(node) for node in nodes]
+
+        async def _multiple_op(values: GetAttrObjectT) -> bool:
+            return multiple_op([await element(values) for element in elements])
```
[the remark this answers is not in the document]
sileht: As I said before `any()` doesn't support async generator :( if we want to do that, we must implement any_async ourself.

And now I have another reason to always evaluate all branches of the tree, you can start take a look at:
https://github.com/Mergifyio/mergify-engine/pull/2654

This change evaluates all the conditions in one pass (using `{"and": <list-of-conditions>}`). I need to evaluate all nodes tree, so I can introspect the evaluation result to keep our fancy summary and display sometimes like:

- [X] base=main
- [ ] any of:
  -  [ ] author=jd
  -  [ ] author=sileht
  -  [ ] all of:
      - [X] author=cam
      - [ ] approved-review-by>=2
- [ ] check-success=GitGuardian Security Checks
- [X] check-success=Semantic Pull Request
- [X] check-success=requirements
- [X] check-success=pep8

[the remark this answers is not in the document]
jd: Ok, makes sense to me then. 👍

jd (approved):

PR #2647, opened 2021-05-22
fix(gitter): disable gc

jd:
This disables the GC auto run in repo handled by Gitter.
By default it runs on large repository and fork() in the background because
gc.autoDetach is true.

This can leads to failure when cleaning up the repository later.

Fixes MERGIFY-ENGINE-26P
Fixes MERGIFY-ENGINE-26Q


jd:
@sileht I was unable to record, if you can lend me a hand here 😉

sileht (approved):

PR #2648, opened 2021-05-22
refactor(actions_runner): remove log filter

jd:
The current filter does not seem to be useful: it expect a list of
dictionaries, and from this list it removes a bunch of keys such as "node_id"
or "tree_id".

The current code sends as argument a list of T_PayloadEventSource which does
not contains any of this.

I imagine this was needed when sources were (probably?) directly a list of
GitHubEventType, but it's not the case anymore, and a lot of filtering is
already done before we log.

### DIFF
--- a/mergify_engine/engine/actions_runner.py
+++ b/mergify_engine/engine/actions_runner.py
@@ -12,7 +12,6 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 import base64
-import copy
 import typing
 
 from datadog import statsd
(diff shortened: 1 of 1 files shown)


sileht:
I think `T_PayloadEventSource` is far from been complete.

sileht:
For a pull request event, just removing url save 3/4 of the payload size

```bash
$ curl -H "Authorization: token $GITHUB_TOKEN" https://api.github.com/repos/Mergifyio/mergify-engine-prod/pulls/26 | wc -c
19139
$ curl -H "Authorization: token $GITHUB_TOKEN" https://api.github.com/repos/Mergifyio/mergify-engine-prod/pulls/26 | grep -v url | wc -c
5004
```

sileht:
Datadog has a size limit per logline, we already miss some logging sometimes because sources is too big. I don't think this helps.

sileht (approved):
Oh, we are in the summary..., we don't have the full payload anymore here.

PR #2650, opened 2021-05-22
ci(mypy): enable disallow_untyped_calls

jd:


mergify_engine/actions/request_reviews.py
```diff
@@ -172,7 +172,9 @@ async def run(
         )
 
         if user_reviews_to_request or team_reviews_to_request:
-            requested_reviews_nb = len(await ctxt.pull_request.review_requested)
+            requested_reviews_nb = len(
+                typing.cast(typing.List[str], await ctxt.pull_request.review_requested)
```
sileht, line 176: I wonder if we can do better for  `PullRequest.__getattr__, I don't really like the casting each time we use an attribute.

sileht (approved):
Ben c'est toujours pas typé datadog ? :yum:

mergify_engine/actions/request_reviews.py
jd: Yes, I thought about it. We should add something like a `@property` for each field that is exposed on the pull request object.

PR #2654, opened 2021-05-23
refactor(rule): evaluate all conditions in one pass

sileht:


mergify_engine/rules/__init__.py
```diff
@@ -35,30 +35,100 @@
...
+        except filter.parser.pyparsing.ParseException as e:
+            raise voluptuous.Invalid(
+                message=f"Invalid condition '{condition}'. {str(e)}",
+                error_message=str(e),
+            )
+        except filter.InvalidQuery as e:
+            raise voluptuous.Invalid(
+                message=f"Invalid condition '{condition}'. {str(e)}",
+                error_message=str(e),
+            )
```
jd, line 62:
```suggestion
        except (filter.parser.pyparsing.ParseException, filter.InvalidQuery) as e:
            raise voluptuous.Invalid(
                message=f"Invalid condition '{condition}'. {str(e)}",
                error_message=str(e),
            )
```

mergify_engine/rules/__init__.py
```diff
@@ -35,30 +35,100 @@
...
+    async def __call__(self, obj: filter.GetAttrObjectT) -> bool:
+        if self.evaluated:
+            raise RuntimeError("RuleCondition can be evaluated only once")
+        self.evaluated = True
+        try:
+            self.match = await self.partial_filter(obj)
+        except live_resolvers.LiveResolutionFailure as e:
+            self.match = False
+            self.evaluation_error = e.reason
+        return self.match
```
jd, line 80:
```suggestion
        try:
            self.match = await self.partial_filter(obj)
        except live_resolvers.LiveResolutionFailure as e:
            self.match = False
            self.evaluation_error = e.reason
        finally:
            self.evaluated = True
        return self.match
```

mergify_engine/rules/__init__.py
```diff
@@ -35,30 +35,100 @@
...
+class RuleConditions:
+    conditions: typing.List[RuleCondition]
+    match: bool = dataclasses.field(init=False, default=False)
+    evaluated: bool = dataclasses.field(init=False, default=False)
+
+    async def eval(self, obj: filter.GetAttrObjectT) -> bool:
+        if self.evaluated:
+            raise RuntimeError("RuleCondition can be evaluated only once")
+        self.evaluated = True
+        self.match = await filter.Filter({"and": self.conditions})(obj)
```
jd, line 100:
```suggestion
        self.match = await filter.Filter({"and": self.conditions})(obj)
        self.evaluated = True
```
Sounds better in case the `Filter` fails.

mergify_engine/rules/filter.py
```diff
@@ -266,7 +263,9 @@ async def _op(obj: GetAttrObjectT) -> bool:
     def _handle_unary_op(
         self, unary_op: UnaryOperatorT, nodes: TreeT
     ) -> typing.Callable[[GetAttrObjectT], typing.Awaitable[bool]]:
-        element = self.build_evaluator(nodes)
+        element: typing.Callable[
+            [GetAttrObjectT], typing.Awaitable[bool]
+        ] = self.build_evaluator(nodes)
```
jd, line 268: Typing sounds useless here, no?

mergify_engine/rules/filter.py
```diff
@@ -276,7 +275,9 @@ async def _unary_op(values: GetAttrObjectT) -> bool:
     def _handle_multiple_op(
         self, multiple_op: MultipleOperatorT, nodes: typing.Iterable[TreeT]
     ) -> typing.Callable[[GetAttrObjectT], typing.Awaitable[bool]]:
-        elements = [self.build_evaluator(node) for node in nodes]
+        elements: typing.List[
+            typing.Callable[[GetAttrObjectT], typing.Awaitable[bool]]
+        ] = [self.build_evaluator(node) for node in nodes]
```
jd, line 280: ditto

jd (changes requested):
Did you also check the output is correct?
It's hard to read in the HTTP log and we don't have much assertion around here.
I'm worried something might be broken.

mergify_engine/rules/filter.py
sileht: I agree, but mypy doesn't.

sileht:
> Did you also check the output is correct?
> It's hard to read in the HTTP log and we don't have much assertion around here.
> I'm worried something might be broken.

The output didn't change, just how we generate it changed.

An example I recorded for this PR: https://github.com/mergifyio-testing/functional-testing-repo/pull/12125/checks

sileht:
I will rerecord the debugger and add more assertions:
https://github.com/Mergifyio/mergify-engine/blob/main/mergify_engine/tests/functional/test_debug.py#L125-L127

mergify_engine/rules/__init__.py
sileht: I have added this boolean to ensure conditions are copied and never reused. It's not there to know if the Filter has worked or not.
It helped me to find all occurrences of `condition(...)` to replace them by `condition.copy()(...)`

Maybe we can find a better name.

jd: Fair enough!

mergify_engine/rules/__init__.py
jd: why u don't take this into account? 😢

sileht: I missed it, fixed now.

sileht:
This one is ready.

jd (approved):

PR #2669, opened 2021-05-24
feat(context): less verbose __repr__

sileht:
Remove some attributes from __repr__ output.

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -38,7 +38,7 @@ from mergify_engine import constants
 from mergify_engine import date
 from mergify_engine import exceptions
 from mergify_engine import github_types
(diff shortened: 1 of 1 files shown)


jd (approved):
Why not passing repr=False to the attributes you don't want?

jd (changes requested):
No I didn't want to approve 😁

sileht:
> Why not passing repr=False to the attributes you don't want?

Because I didn't `field()` has repr :yum:

sileht:
We can't implement __repr__ for TypedDict(). This is maybe still a bit too verbse.

Let's see how it goes.

jd (approved):

PR #2672, opened 2021-05-24
fix(docs): regex101 certificate has expired

sileht:


docs/source/configuration.rst
```diff
@@ -132,14 +132,14 @@ Regular Expressions
...
 
-  You can use `regex101 <https://regex101.com/>`_ or `PyRegex
-  <http://www.pyregex.com>`_ to test your regular expressions.
+   You can use `PyRegex <http://www.pyregex.com>`_ or `Pythex
+   <https://pythex.org/>`_ to test your regular expressions.
 
 ..
-   # FIXME(sileht): pyregex is currently down, ssl expired
+   # FIXME(sileht): regex101 is currently down, ssl expired
    You can use `regex101 <https://regex101.com/>`_, `PyRegex
```
jd, line 140: This is not removed and it works?

sileht: linkchecker doesn't looks at comments

sileht:
The site has been fixed.

PR #2675, opened 2021-05-24
feat(rules): allow and/or in conditions

sileht:
This allows to use and and or in conditions except for some complicated
attributes.

Fixes MRGFY-117
Fixes MRGFY-470
Related MRGFY-16


sileht:
https://github.com/mergifyio-testing/functional-testing-repo/pull/12142/checks

PR #2680, opened 2021-05-25
feat(rules): allow and/or in conditions

sileht:
## feat(rules): allow and/or in conditions

This allows to use and and or in conditions except for some complicated
attributes.

Fixes MRGFY-117
Fixes MRGFY-470
Related MRGFY-16

## feat: allow merged/closed/author/base/head in and/or conditions

Fixes MRGFY-117
Fixes MRGFY-470


docs/source/conditions.rst
```diff
@@ -311,35 +311,37 @@ modified files:
...
+      - name: automatic merge if author is foo or bar
         conditions:
-          - author=foo
+          - or:
+            - author=foo
+            - author=bar
           - check-success=Travis CI - Pull Request
         actions:
           merge:
             method: merge
```
jd, line 333: This needs more example, at least one with `and` and one imbricated.

mergify_engine/rules/__init__.py
```diff
@@ -35,33 +37,44 @@
 LOG = daiquiri.getLogger(__name__)
 
 
+# This help mypy for breaking the recurcive definition
```
jd, line 40:
```suggestion
# This helps mypy breaking the recursive definition
```

mergify_engine/rules/__init__.py
```diff
@@ -35,33 +37,44 @@
...
+FakeTreeT = typing.Dict[str, typing.Any]
+
+
 @dataclasses.dataclass
 class RuleCondition:
```
jd, line 45: It'd be great to add some docstring at some point, this is getting hard to follow.

mergify_engine/rules/__init__.py
```diff
@@ -75,16 +88,68 @@ async def __call__(self, obj: filter.GetAttrObjectT) -> bool:
...
+        tree = typing.cast(filter.TreeT, self.partial_filter.tree)
+        tree = tree.get("-", tree)
         name = list(tree.values())[0][0]
         if name.startswith(filter.Filter.LENGTH_OPERATOR):
             return str(name[1:])
         return str(name)
 
 
+@dataclasses.dataclass
+class RuleConditionGroup:
```
jd, line 100: ditto would be great to have some docstrings.

mergify_engine/rules/__init__.py
```diff
@@ -75,16 +88,68 @@ async def __call__(self, obj: filter.GetAttrObjectT) -> bool:
...
+            if isinstance(cond, RuleCondition):
+                attribute = cond.get_attribute_name()
+                # TODO(sileht): remove this heuristics to allow this attribute everywhere
+                if attribute.partition("-")[0] in [
+                    "check",  # heuristic in merge/queue
+                    "status",  # heuristic in merge/queue
+                ]:
+                    raise voluptuous.Invalid(
+                        f"{attribute} cannot be used inside and/or"
+                    )
```
jd, line 135: It'd be great to validate this earlier in the schema too but I imagine it's temporary?

mergify_engine/rules/__init__.py
```diff
@@ -75,16 +88,68 @@ async def __call__(self, obj: filter.GetAttrObjectT) -> bool:
...
+        return self.match
+
+    def copy(self) -> "RuleConditionGroup":
+        return RuleConditionGroup({self.operator: [c.copy() for c in self.conditions]})
+
+
 @dataclasses.dataclass
 class RuleConditions:
-    conditions: typing.List[RuleCondition]
+    conditions: typing.List[typing.Union[RuleConditionGroup, RuleCondition]]
```
jd, line 152: Do we really need to have 3 different classes?

I could easily imagine the `RuleConditions` would be a `RuleConditionGroup` where we set it to `and` manually.

i.e:
```
conditions:
  - foo
  - bar
```
is transformed to
```
conditions:
  - and:
    - foo
    - bar
```

And you don't need 3 different classes.

mergify_engine/rules/__init__.py
```diff
@@ -95,35 +160,79 @@ async def eval(self, obj: filter.GetAttrObjectT) -> bool:
         self.match = await filter.Filter({"and": self.conditions})(obj)
         return self.match
 
-    def __getitem__(self, idx: int) -> RuleCondition:
-        return self.conditions[idx]
+    def iter_root_rule_conditions(
```
jd, line 163: IIRC this is temporary until we support `check-` everywhere?

mergify_engine/rules/__init__.py
```diff
@@ -95,35 +160,79 @@ async def eval(self, obj: filter.GetAttrObjectT) -> bool:
...
+    ) -> typing.Iterator[RuleCondition]:
+        for condition in self.conditions:
+            if isinstance(condition, RuleCondition):
+                yield condition
 
-    def __iter__(self) -> typing.Iterator[RuleCondition]:
-        # TODO(sileht): browse the complete tree later
-        return iter(self.conditions)
+    @classmethod
+    def _traverse_rule_condition(
```
jd, line 171:
```suggestion
    def walk(
```
`walk` is more often used in this case like `os.walk`

mergify_engine/rules/__init__.py
```diff
@@ -95,35 +160,79 @@ async def eval(self, obj: filter.GetAttrObjectT) -> bool:
...
-        return len(self.conditions)
+    def traverse_rule_condition(self) -> typing.Iterator[RuleCondition]:
+        for condition in self._traverse_rule_condition(self.conditions):
+            yield condition
+
+    def is_faulty(self) -> bool:
+        return (
+            first.first(c for c in self.traverse_rule_condition() if c.evaluation_error)
+            is not None
+        )
```
jd, line 191:
```suggestion
        return (
            any(c.evaluation_error for c in self.traverse_rule_condition())
        )
```

mergify_engine/rules/__init__.py
```diff
@@ -95,35 +160,79 @@ async def eval(self, obj: filter.GetAttrObjectT) -> bool:
...
+        summary += f"- [{checked}] `{cond}`"
+        if cond.description:
+            summary += f" [{cond.description}]"
+        if cond.evaluation_error:
+            summary += f" ⚠️ {cond.evaluation_error}"
+        summary += "\n"
         return summary
 
+    @classmethod
+    def _traverse_for_summary(
```
jd, line 209: -> `walk`

mergify_engine/rules/__init__.py
```diff
@@ -289,18 +397,28 @@ async def create(
...
-            for condition in rule.conditions:
-                if hide_rule and condition.evaluation_error:
-                    self.faulty_rules.append(typing.cast(T_EvaluatedRule, rule))
-                    break
-                elif (
-                    hide_rule
-                    and condition.get_attribute_name() in self.BASE_ATTRIBUTES
-                    and not condition.match
-                ):
+            categorized = False
```
jd, line 400: can you explain?

mergify_engine/rules/__init__.py
```diff
@@ -289,18 +397,28 @@ async def create(
...
+
+            if hide_rule and not rule.conditions.match:
+                # NOTE(sileht): Replace non base attributes by true, if it
+                # still match it's a potential rule otherwise hide it.
```
jd, line 404:
```suggestion
                # NOTE(sileht): Replace non-base attribute by true, if it
                # still matches it's a potential rule otherwise hide it.
```

mergify_engine/rules/__init__.py
```diff
@@ -459,7 +594,7 @@ def get_pull_request_rules_schema(partial_validation: bool = False) -> voluptuou
                     voluptuous.Required("name"): str,
                     voluptuous.Required("hidden", default=False): bool,
                     voluptuous.Required("conditions"): voluptuous.All(
-                        [voluptuous.All(str, voluptuous.Coerce(RuleCondition))],
+                        [voluptuous.Coerce(RuleConditionSchema)],
```
jd, line 597: This means that this is valid:
```
conditions:
  or:
    - foobar
    - foobaz
```

I think it's neat, but we might want to add a test?

jd (changes requested):
Probably need more tests too 👍

jd:
Also title is wrong

mergify_engine/rules/__init__.py
sileht: It's done during the validation because it's done in `__post_init__()` and within a `voluptuous.Coerce()`.

And yes it should temporary, but I didn't find yet how to workaround the check-XXX

mergify_engine/rules/__init__.py
sileht: yes

mergify_engine/rules/__init__.py
sileht: In the summary, we display rules in four groups:
* rules where all attributes match
* rules where only `BASE_ATTRIBUTES` match (filter out rule written for other branches)
* rules that won't work due to a configuration issue detected at runtime (eg team doesn't exists)
* rules where only `BASE_ATTRIBUTES` don't match (mainly to hide rule written for other branches)

jd: Ok, can you add this as a comment in the code so it's easier for understand to everyone passing by? 😉

mergify_engine/rules/__init__.py
sileht: That looks good, but I will focus on the current structure and add this neat syntax later.

jd:
Is this splittable in multiple PRs?

jd (changes requested):

sileht:
> Is this splittable in multiple PRs?

I can send the 3 commits separately, but no more.

jd:
> I can send the 3 commits separately, but no more.

That'd be a good start.

sileht:
I close this one, I will maintain the 3 other with `git-push-stack`

PR #2683, opened 2021-05-26
fix(post_check): fix the type of check_succeed

sileht:
check_succeed must be a boolean not a string.


jd (approved):

mergify_engine/context.py
```diff
@@ -1237,7 +1237,9 @@ async def items(self) -> typing.Dict[str, ContextAttributeType]:
     async def render_template(
         self,
         template: str,
-        extra_variables: typing.Optional[typing.Dict[str, str]] = None,
+        extra_variables: typing.Optional[
+            typing.Dict[str, typing.Union[str, bool]]
```
jd, line 1241: This should probably be `typing.Any` since it's jus passing down objects, no?

sileht: I explicitly put str, bool, to control what we pass to Jinja, I don't want to send list or dict or a real object.

jd: Well that could happen at some point for good reasons, no?

sileht: It's Jinja, if we send a `context.Context()` in a template, I don't want to imagine what people can do with that.

PR #2684, opened 2021-05-26
fix(bot_account): allow admin to be used

sileht:
Except for merge_bot_account.

### DIFF
--- a/mergify_engine/actions/utils.py
+++ b/mergify_engine/actions/utils.py
@@ -18,6 +18,7 @@ import typing
 
 from mergify_engine import check_api
 from mergify_engine import context
+from mergify_engine import github_types
 from mergify_engine import subscription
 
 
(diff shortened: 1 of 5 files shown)


mergify_engine/actions/utils.py
```diff
@@ -45,19 +47,30 @@ async def validate_bot_account(
             ),
         )
 
-    if need_write_permission:
+    if required_permissions is None:
+        required_permissions = ["admin", "write", "maintain"]
+
+    if required_permissions:
```
jd, line 53: This is suboptimal, as if `required_permissions` is `None` (e.g., `comment`), then there's no need to check for permissions at all.

mergify_engine/actions/utils.py
```diff
@@ -29,7 +29,9 @@ async def validate_bot_account(
     # TODO(sileht): make it mandatory when all bot_account need subscription
     required_feature: typing.Optional[subscription.Features],
     missing_feature_message: str = "This action with `bot_account` set is unavailable",
-    need_write_permission: bool = True,
+    required_permissions: typing.Optional[
```
jd, line 32: Is there a GitHub type we could use rather than our own literal?

jd (dismissed):

mergify_engine/actions/utils.py
sileht: comment has `[]` not `None` to avoid the permission lookup. The code is confusing, just because mypy complains I was using mutable in function arguments, so I put `None`. It may be more readable with an `Unset` object marker.

jd: Right. Indeed it's a bit confusing and I understand mypy. 😆

jd (approved):

PR #2694, opened 2021-05-27
refactor(filter): make it generic

sileht:
This makes the Filter class Generic.

The binary implementation is now called FilterBinary

A future implementation will be done to return the more recent
timestamp.

Change-Id: Ice711d1f106d445badf6dc5539fc1b309887b35a

PR #2693, opened 2021-05-27
feat(conditions): allow checks condition in and/or

sileht:
Change-Id: I9f2df8ae19a41352a39446e2e114a60c1ce64292

PR #2692, opened 2021-05-27
feat: allow merged/closed/author/base/head in and/or conditions

sileht:
Fixes MRGFY-117
Fixes MRGFY-470

Change-Id: Ia0458dc9eb5f1fbe88529bc2c62bcbfb0488f948

PR #2691, opened 2021-05-27
feat(rules): allow and/or in conditions

sileht:
This allows to use and and or in conditions except for some complicated
attributes.

Fixes MRGFY-117
Fixes MRGFY-470
Related MRGFY-16

Change-Id: I135062de71ed647e95fef9c9153e385bacc53119

PR #2695, opened 2021-05-27
feat(current-time): add time conditions

sileht:
This adds current-time conditions. To make it works it introduces two new
concept:

* delayed refresh, it schedules the next time we must refresh the pull
  request
* NearDatetimeFilter, it returns the NearDatetime that match the
  conditions

Change-Id: I63f5bc6750ab00484c54242c1fc79ad486824a3b


sileht:
* feat(rules): allow and/or in conditions ([#2691](https://github.com/Mergifyio/mergify-engine/pull/2691))
* feat: allow merged/closed/author/base/head in and/or conditions ([#2692](https://github.com/Mergifyio/mergify-engine/pull/2692))
* feat(conditions): allow checks condition in and/or ([#2693](https://github.com/Mergifyio/mergify-engine/pull/2693))
* refactor(filter): make it generic ([#2694](https://github.com/Mergifyio/mergify-engine/pull/2694))
* feat(time) add time conditions ([#2695](https://github.com/Mergifyio/mergify-engine/pull/2695))

docs/source/conditions.rst
```diff
@@ -252,6 +252,15 @@ Here's the list of pull request attribute that can be used in conditions:
...
      - string
      - The title of the pull request.
+   * - ``time``
+     - string
+     - A time in format ``HH:MM``.
+
+
+.. important::
+
+   ``time`` does not support ``~=``, ``=`` and ``!=`` operators.
```
jd, line 262: It makes a bit sad that we don't support `=` TBH. :)

docs/source/conditions.rst
```diff
@@ -252,6 +252,15 @@ Here's the list of pull request attribute that can be used in conditions:
    * - ``title``
      - string
      - The title of the pull request.
+   * - ``time``
+     - string
+     - A time in format ``HH:MM``.
```
jd, line 257: This does not make any sense. What time it is?
Time of creation? Time of update?

Reading the code I think you want this to be named `now`, no?

mergify_engine/delayed_refresh.py
```diff
@@ -0,0 +1,86 @@
...
+
+async def plan_next_refresh(match: rules.RulesEvaluator, ctxt: context.Context) -> None:
+    best_bet: typing.Optional[datetime.datetime] = None
+    for rule in match.matching_rules:
+        f = filter.NearDatetimeFilter(rule.conditions.extract_raw_filter_tree())
+        bet = await f(ctxt.pull_request)
+        if best_bet is None or best_bet > bet:
+            best_bet = bet
+
+    zset_subkey = f"{ctxt.repository.installation.owner_id}~{ctxt.repository.installation.owner_login}~{ctxt.repository.repo['id']}~{ctxt.repository.repo['name']}~{ctxt.pull['number']}"
```
jd, line 43: why do we include the login here?
I'd think sticking to `id` was safer ane enough here.

mergify_engine/rules/filter.py
```diff
@@ -147,11 +163,12 @@ async def __call__(self, obj: GetAttrObjectT) -> FilterResultT:
 
     LENGTH_OPERATOR = "#"
 
-    def _to_list(self, item: typing.Any) -> typing.List[typing.Any]:
-        if isinstance(item, list):
-            return item
+    @staticmethod
+    def _to_list(item: typing.Any) -> typing.List[typing.Any]:
```
jd, line 167:
```suggestion
   
    @typing.overload
    def _to_list(item: typing.Iterable[_T]) -> typing.List[_T]: ...

    @typing.overload
    def _to_list(item: _T) -> typing.List[_T]: ...

    @typing.overload
    def _to_list(item: str) -> typing.List[str]: ...

    def _to_list(item) -> typing.List[_T]:
```
?

mergify_engine/rules/filter.py
```diff
@@ -287,3 +304,97 @@ def BinaryFilter(
...
+    """
+    The principe:
+    * the attribute can't be mapped to a datetime -> datetime.datetime.max
+    * the time/datetime attribute can't change in the future -> datetime.datetime.max
+    * the time/datetime attribute can change in the future -> return when
+    * we have a list of time/datetime, we pick the more recent
+    """
+    return Filter[datetime.datetime](
+        tree,
+        {"-": _same_datetime},
```
jd, line 385: That does not sound right. I was trying to look for a test to understand how that worked and I didn't find one. 👎  😁 

Also can't you use `_identity`? If it's a typing issue it should work with a `TypeVar`.

mergify_engine/rules/filter.py
```diff
@@ -287,3 +304,97 @@ def BinaryFilter(
...
+        {
+            "=": (_dt_op(operator.eq), _minimal_datetime, _identity),
+            "<": (_dt_op(operator.lt), _minimal_datetime, _identity),
+            ">": (_dt_op(operator.gt), _minimal_datetime, _identity),
+            "<=": (_dt_op(operator.le), _minimal_datetime, _identity),
+            ">=": (_dt_op(operator.ge), _minimal_datetime, _identity),
+            "!=": (_dt_op(operator.ne), _minimal_datetime, _identity),
+            # NOTE(sileht): This is not allowed in parser, so we are same to ignore datetime here
```
jd, line 393: I didn't understand the comment 😢

mergify_engine/rules/filter.py
```diff
@@ -287,3 +304,97 @@ def BinaryFilter(
             "and": all,
         },
     )
+
+
+DT_INFINITY = datetime.datetime.max.replace(tzinfo=datetime.timezone.utc)
```
jd, line 309:
```suggestion
DT_MAX = datetime.datetime.max.replace(tzinfo=datetime.timezone.utc)
```
Infinity has 2 directions, it's unclear which one it is.

mergify_engine/rules/filter.py
```diff
@@ -287,3 +304,97 @@ def BinaryFilter(
...
+    op: typing.Callable[[typing.Any, typing.Any], bool],
+) -> typing.Callable[[typing.Any, typing.Any], datetime.datetime]:
+    def _operator(value: typing.Any, ref: typing.Any) -> datetime.datetime:
+        dt_value = _as_datetime(value)
+        dt_ref = _as_datetime(ref)
+        handle_equality = op in (operator.eq, operator.ne, operator.le, operator.ge)
+        if handle_equality and dt_value == dt_ref:
+            # It will change very soon!
+            try:
+                return dt_ref + datetime.timedelta(minutes=5)
```
jd, line 352: Can you add a comment on what is this 5 minutes here?

mergify_engine/rules/filter.py
```diff
@@ -287,3 +304,97 @@ def BinaryFilter(
...
+
+def NearDatetimeFilter(
+    tree: typing.Union[TreeT, CompiledTreeT[GetAttrObject, datetime.datetime]],
+) -> "Filter[datetime.datetime]":
+    """
+    The principe:
+    * the attribute can't be mapped to a datetime -> datetime.datetime.max
+    * the time/datetime attribute can't change in the future -> datetime.datetime.max
+    * the time/datetime attribute can change in the future -> return when
+    * we have a list of time/datetime, we pick the more recent
```
jd, line 381:
```suggestion
    The principles:
    * the attribute can't be mapped to a datetime -> datetime.datetime.max
    * the time/datetime attribute can't change in the future -> datetime.datetime.max
    * the time/datetime attribute can change in the future -> return when
    * we have a list of time/datetime, we pick the more recent
```

jd (dismissed):
I've a question: what

I'm wondering something with this design is that I feel like there's a chance it misses.

E.g. you schedule a refresh at 23:50 because you have `time>=23:50`.

The refresh event is pushed
The engine is overloaded and starts processing the event at 00:01.
The rules does not match anymore. 😢 

You missed the rule.

No?

docs/source/conditions.rst
sileht: If we go with `now`, then it should be `now-time`, to also have `now-day`, `now-month`, `now-day-of-week`.

jd: Ah. We're back at "naming things is hard". 😆 

what about `current-time` and `current-day`, `current-month`… ? I'm stealing this naming from PostgreSQL.

sileht (commented):
> I've a question: what
> 
> I'm wondering something with this design is that I feel like there's a chance it misses.
> 
> E.g. you schedule a refresh at 23:50 because you have `time>=23:50`.
> 
> The refresh event is pushed
> The engine is overloaded and starts processing the event at 00:01.
> The rules does not match anymore. cry
> 
> You missed the rule.
> 
> No?

Yes, that the same issue as why I skip the = implementation, we may need to add some minutes to the threshold.

mergify_engine/delayed_refresh.py
jd: Ok, I'm fine with it. I just wanted to be sure it was an explicit choice.
I imagine it'd be ok to remove some name-based logging in the worker since this is not useful in general on a day-to-day basis.

Then the org-renaming issue would be "solved". But that can also be done later.

docs/source/conditions.rst
jd: Yep, which circle back to the issue I was mentioning here https://github.com/Mergifyio/mergify-engine/pull/2695#pullrequestreview-677099915

mergify_engine/rules/filter.py
```diff
@@ -147,11 +165,12 @@ async def __call__(self, obj: GetAttrObjectT) -> FilterResultT:
 
     LENGTH_OPERATOR = "#"
 
-    def _to_list(self, item: typing.Any) -> typing.List[typing.Any]:
-        if isinstance(item, list):
-            return item
+    @staticmethod
+    def _to_list(item: typing.Union[_T, typing.Iterable[_T]]) -> typing.List[_T]:
```
jd, line 169: Why not picking my suggested `overload` approach here?

mergify_engine/rules/filter.py
sileht: I added a test

mergify_engine/tests/unit/rules/test_filter.py
```diff
@@ -154,6 +157,242 @@ async def test_str() -> None:
...
+
+    f = filter.NearDatetimeFilter({"!=": ("foo", today.timetz())})
+    assert await f(FakePR({"foo": time(5, 8)})) == soon
+    assert await f(FakePR({"foo": time(2, 1)})) == today
+    assert await f(FakePR({"foo": time(5, 1)})) == today
+    assert await f(FakePR({"foo": time(6, 2)})) == nextday
+    assert await f(FakePR({"foo": time(8, 9)})) == nextday
+
+    f = filter.NearDatetimeFilter({"-": {"<=": ("foo", today.timetz())}})
+    assert await f(FakePR({"foo": time(5, 8)})) == soon
```
jd, line 268: Not sure I understand here. If you say `-current-time<=05:08` then it means `current-time>05:08` which means you should schedule at midnight. The result here should be exactly like the L246 `f = filter.NearDatetimeFilter({">": ("foo", today.timetz())})`.

jd (changes requested):

docs/source/conditions.rst
```diff
@@ -252,6 +252,18 @@ Here's the list of pull request attribute that can be used in conditions:
...
+     - string
+     - The current time in format ``HH:MM``.
+
+
+.. important::
+
+   Time based conditions have a 5 minutes precision. If you wrote a rule like ``current-time>=10:00``, Mergify will refresh
+   the pull request state to detect the condition change between 10:00 and 10:05 if no event has been received in the meantime.
+
+   ``current-time`` does not support ``~=``, ``=`` and ``!=`` operators.
```
jd, line 265:
```suggestion
   Time-based conditions have a 5 minutes precision at best. Do not write conditions based on time ranges that are too narrow or the condition might never be true.

   ``current-time`` does not support ``~=``, ``=`` and ``!=`` operators.
```

mergify_engine/rules/filter.py
```diff
@@ -55,7 +58,10 @@ def __init__(self, arguments: typing.Any) -> None:
         self.arguments = arguments
 
 
-def _identity(value):
+_IdentityT = typing.TypeVar("_IdentityT")
```
jd, line 61: you can reuse the same `_T` as below

jd (changes requested):
still test issues :)

mergify_engine/rules/filter.py
```diff
@@ -283,3 +304,101 @@ def BinaryFilter(
...
+    * the attribute can't be mapped to a datetime -> datetime.datetime.max
+    * the time/datetime attribute can't change in the future -> datetime.datetime.max
+    * the time/datetime attribute can change in the future -> return when
+    * we have a list of time/datetime, we pick the more recent
+    """
+    return Filter[datetime.datetime](
+        tree,
+        # NOTE(sileht): This is not allowed in parser on all time based attributes
+        # so we can just return DT_MAX for all other attributes
+        {"-": _dt_identiy_max},
```
jd, line 388:
```suggestion
        {"-": _dt_identity_max},
```

jd (changes requested):
small typo

PR #2707, opened 2021-06-01
chore: finish the bot_account deprecation period

sileht:
We have tell reported on all public repository during three months we
will break their merge and rebase action today.

This change does it.

Fixes MRGFY-395
Fixes MRGFY-420

Change-Id: I8f698cb01763069ce27f0d32139a52df2baaea58


mergify_engine/actions/utils.py
```diff
@@ -53,12 +54,20 @@ async def validate_bot_account(
...
-            ),
-        )["permission"]
+        try:
+            permission = typing.cast(
+                github_types.GitHubRepositoryCollaboratorPermission,
+                await ctxt.client.item(
+                    f"{ctxt.base_url}/collaborators/{bot_account}/permission"
+                ),
+            )["permission"]
+        except http.HTTPNotFound:
```
jd, line 64: This looks good though unrelated no?

jd (changes requested):

mergify_engine/actions/utils.py
sileht: Yes and no.
No because before the rebase action didn't check for the permission and just fail on git push, now I check the permission upfront, so I had to handle the error here, since a test was covering this case.
And yes because, the merge action already uses this code and may have a bug if the user doesn't here.

jd (approved):

PR #2709, opened 2021-06-01
feat(conditions): add schedule condition

sileht:
Change-Id: Ia58ae9db00a50997954346435314b7231363f158

PR #2708, opened 2021-06-01
feat(conditions): add support for named current-day-of-week

sileht:
Change-Id: Iac71bd17e89a93c2172dd9f93c62d4f1944678e7

PR #2705, opened 2021-05-31
feat(conditions): add current-day/month/year/day-of-week conditions

sileht:
Change-Id: Id6407aaff6e4ad76f55f7a12770a07dbd7cf6769

PR #2710, opened 2021-06-01
feat(conditions): add updated-at condition

sileht:
Change-Id: I8bdf4821605d3b9efcbf473197294095c3066d02

### DIFF
--- a/mergify_engine/date.py
+++ b/mergify_engine/date.py
@@ -14,6 +14,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 import dataclasses
+import datetime
 
 
 @dataclasses.dataclass(order=True)
@@ -66,3 +67,13 @@ class DayOfWeek(PartialDatetime):
 
     def __str__(self) -> str:
         return self._SHORT_DAY[self.value - 1].capitalize()
+
+
+def fromisoformat(s: str) -> datetime.datetime:
+    if s[-1] == "Z":
+        s = s[:-1]
+    dt = datetime.datetime.fromisoformat(s)
+    if dt.tzinfo is None:
+        return dt.replace(tzinfo=datetime.timezone.utc)
+    else:
+        return dt.astimezone(datetime.timezone.utc)
--- a/mergify_engine/tests/unit/rules/test_parser.py
+++ b/mergify_engine/tests/unit/rules/test_parser.py
@@ -15,11 +15,15 @@
 # under the License.
 import datetime
 
+from freezegun import freeze_time
 import pyparsing
 import pytest
 
 from mergify_engine import date
 from mergify_engine.rules import parser
+
+
+now = datetime.datetime.fromisoformat("2012-01-14T20:32:00+00:00")
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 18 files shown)


sileht:
1. feat(rules): allow and/or in conditions ([#2691](https://github.com/Mergifyio/mergify-engine/pull/2691))
1. feat: allow merged/closed/author/base/head in and/or conditions ([#2692](https://github.com/Mergifyio/mergify-engine/pull/2692))
1. feat(conditions): allow checks condition in and/or ([#2693](https://github.com/Mergifyio/mergify-engine/pull/2693))
1. refactor(filter): make it generic ([#2694](https://github.com/Mergifyio/mergify-engine/pull/2694))
1. feat(time): add time conditions ([#2695](https://github.com/Mergifyio/mergify-engine/pull/2695))
1. feat(conditions): add day/month/year/day-of-week conditions ([#2705](https://github.com/Mergifyio/mergify-engine/pull/2705))
1. feat(conditions): add support for named day-of-week ([#2708](https://github.com/Mergifyio/mergify-engine/pull/2708))
1. feat(conditions): add calendar condition ([#2709](https://github.com/Mergifyio/mergify-engine/pull/2709))
1. feat(conditions): add update-at condition ([#2710](https://github.com/Mergifyio/mergify-engine/pull/2710))

sileht:
test

mergify_engine/tests/unit/rules/test_parser.py
```diff
@@ -128,6 +132,27 @@
...
+            "updated-at>=18:02",
+            {">=": ("updated-at", now + datetime.timedelta(hours=18, minutes=2))},
+        ),
+        (
+            f"updated-at<={now.isoformat()}",
+            {"<=": ("updated-at", now)},
+        ),
+        (
+            "updated-at<=7 days",
+            {"<=": ("updated-at", now + datetime.timedelta(days=7))},
```
jd, line 145:
```suggestion
            "updated-at<=7 days",
            {"<=": ("updated-at", now + datetime.timedelta(days=7))},
```

I don't think this is right. It means it gets translated to `update-at <= now + 7 days`.

If you write `updated-at<=7 days` you mean: `updated-at` is between now and 7 days ago which means:
`now-7days<=update-at<now()`. The `<now()` is obviously true, so we can skip it so it boils down to `now -  7 days <= updated-at` which is equivalent to `now <= updated_at + 7 days`.

jd (changes requested):

mergify_engine/date.py
```diff
@@ -66,3 +67,13 @@ def from_string(cls, string: str) -> "DayOfWeek":
 
     def __str__(self) -> str:
         return self._SHORT_DAY[self.value - 1].capitalize()
+
+
+def fromisoformat(s: str) -> datetime.datetime:
```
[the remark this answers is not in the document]
sileht: For 1. I have plan to move `utcnow()` to the  module `date`. I will do that separatly.

mergify_engine/tests/unit/rules/test_parser.py
sileht: :facepalm: 
```patch
- +
+ - 
```

jd: Actually, it's not just changing the sign. We're not talking about the same things.

1. When I write `updated-at <= 7 days` I mean "updated-at is less than 7 days old" which means: `now - 7 days <= updated-at`

2. When I wrote `updated-at <= 7 days ago` I mean "updated-at date is before the day it was 7 days ago" which translates to `updated-at <= now - 7 days`.

The problem here is that the language says 1 but the code runs 2.

We might need 2 different syntaxes here to make things very clear. It's not clear to me.

jd (changes requested):
unclear

PR #2717, opened 2021-06-04
feat: temporary publish docs from pull build on gcp

sileht:
Because pull_request workflow run on forked repository doesn't have
access to secret this is done in two steps:
1. docs is uploaded as GitHub artefact
2. a priviledged workflow copy the artefact to gcp

Change-Id: I5dca961bb515876da004631a8890c3549b951a04

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -89,6 +89,14 @@ jobs:
           sudo apt-get install graphviz
           pip install tox
           tox -e docs
+      # NOTE(sileht): workflow run on/pull_request doesn't have access to
+      # secrets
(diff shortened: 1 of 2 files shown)


jd:
This seems to copy the zip file, does that extract it so it's readable after?

sileht:
> This seems to copy the zip file, does that extract it so it's readable after?

Yeah, I miss the unzip. I can't really test it without merging it. I just copy/paste/adapt this GitHub blog post:

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

So the docs-pr.yaml may have some other issue.

jd (approved):
ok let's try then

PR #2722, opened 2021-06-07
fix(docs): finish upload to gcp

sileht:
This:
* splits out Documentation job from ci.yaml. This allows to upload to
documentation even if other part of the CI fail
* post a check run with the url of the documentation

Cloudfare Worker is used to redirect docs-preview.mergify.io to our
public gcp bucket.

Change-Id: I3890e3c2699e79cb50ab946e5ba1000fc975fb30

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -1,0 +1,34 @@
+name: Documentation
+permissions: read-all
+
+on:
+  pull_request:
+    branches:
+      - main
+      - git_push_stack/**
+
+jobs:
+  docs:
+    timeout-minutes: 5
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - uses: actions/setup-python@v2.2.2
+        with:
+          python-version: 3.9
+      - run: |
+          sudo apt-get install graphviz
+          pip install tox
+          tox -e docs
+      # NOTE(sileht): workflow run on/pull_request doesn't have access to
+      # secrets
+      # So, we upload it as artifact, then another workflow
+      # docs-upload-gcp.yaml will upload it on gcp.
+      - uses: actions/upload-artifact@v2
+        with:
+          name: docs-pr-${{ github.event.pull_request.number }}
+          path: docs/build
+
+      # yamllint disable-line rule:line-length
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-docs.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/setup-python@v2.2.2
+        with:
+          python-version: 3.9.5
+      - run: |
+          sudo apt-get install graphviz
+          pip install tox
+          tox -e docs
+      # NOTE(sileht): workflow run on/pull_request doesn't have access to
+      # secrets
+      # So, we upload it as artefact, then another workflow docs-pr.yaml will
```
jd, line 25:
```suggestion
      # So, we upload it as artifact, then another workflow docs-upload-gcp.yaml will
```

.github/workflows/docs-upload-gcp.yaml
```diff
@@ -1,18 +1,20 @@
-name: Upload temporary docs of pull request to GCP
+name: Upload temporary docs of to GCP
```
jd, line 1:
```suggestion
name: Upload temporary docs to GCP
```

.github/workflows/docs-upload-gcp.yaml
```diff
@@ -23,6 +25,7 @@ jobs:
                repo: context.repo.repo,
                run_id: ${{github.event.workflow_run.id }},
             });
+            console.log(artifacts.data.artifacts);
```
jd, line 28: want?

jd (changes requested):
typo

.github/workflows/docs-upload-gcp.yaml
sileht: yes !

.github/workflows/ci-docs.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  docs:
+    timeout-minutes: 5
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - uses: actions/setup-python@v2.2.2
+        with:
+          python-version: 3.9.5
```
jd, line 18:
```suggestion
          python-version: 3.9
```
I don't think we need to me that precise

jd (dismissed):

jd (approved):

PR #2734, opened 2021-06-08
feat(refactor): move branch protection checks earlier in the process

sileht:
Change-Id: Ibeae85d3e4ecfb299ce8cd0c5466d13e2fe9aea9

PR #2735, opened 2021-06-08
fix(queue): get_rule_checks_status return value

sileht:
When some CIs are pending and some success, get_rule_checks_status may
return FAILURE instead of PENDING

This change fixes that.

Change-Id: I9acb38f149b1174196d13bb5d5fe3c18f9ed5048

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -778,6 +778,13 @@ class Context(object):
             comments, _ = await self.consolidated_reviews()
             return [r["user"]["login"] for r in comments if r["state"] == "COMMENTED"]
 
+        elif name == "check-success-or-neutral-or-pending":
+            return [
+                ctxt
+                for ctxt, state in (await self.checks).items()
+                if state in ("success", "neutral", "pending", None)
+            ]
+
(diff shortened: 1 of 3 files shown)


sileht:
1. feat(refactor): move branch protection checks earlier in the process ([#2734](https://github.com/Mergifyio/mergify-engine/pull/2734))
1. fix(queue): get_rule_checks_status return value ([#2735](https://github.com/Mergifyio/mergify-engine/pull/2735))

mergify_engine/context.py
```diff
@@ -778,6 +778,13 @@ async def _get_consolidated_data(self, name: str) -> ContextAttributeType:
             comments, _ = await self.consolidated_reviews()
             return [r["user"]["login"] for r in comments if r["state"] == "COMMENTED"]
 
+        elif name == "check-success-or-neutral-or-pending":
```
jd, line 781: Do we want to add this in the documentation?

sileht: No it's internal only.

jd (approved):

sileht:
@Mergifyio rebase

PR #2736, opened 2021-06-08
fix(ci): use status instead of check-runs to report the doc preview link

sileht:
Change-Id: I553ea5d3dc50f5c80cf048e78bf5951954c3c448


jd (approved):

sileht:
> Not seeing a difference?
> I'm also dubious about why this check is reported under CodeQL?

I didn't set a check_suite,, I guess GitHub pick the last one.

PR #2750, opened 2021-06-11
chore(time): never refresh pull request bellow 10 minutes

sileht:
No need to refresh the pull request under our documented precision.

Change-Id: I41e848cbc5efc7d9616d87e0e2b46696c3b2c6b9

### DIFF
--- a/mergify_engine/delayed_refresh.py
+++ b/mergify_engine/delayed_refresh.py
@@ -30,6 +30,7 @@ from mergify_engine.rules import filter
 LOG = daiquiri.getLogger(__name__)
 
 DELAYED_REFRESH_KEY = "delayed-refresh"
+DELAYED_REFRESH_MIN_MINUTES = 10
 
 
 async def plan_next_refresh(match: rules.RulesEvaluator, ctxt: context.Context) -> None:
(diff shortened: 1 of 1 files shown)


mergify_engine/delayed_refresh.py
```diff
@@ -30,6 +30,7 @@
 LOG = daiquiri.getLogger(__name__)
 
 DELAYED_REFRESH_KEY = "delayed-refresh"
+DELAYED_REFRESH_MIN_MINUTES = 10
```
jd, line 33: doc says 5?

mergify_engine/delayed_refresh.py
```diff
@@ -44,11 +45,18 @@ async def plan_next_refresh(match: rules.RulesEvaluator, ctxt: context.Context)
...
     if best_bet is None or best_bet >= filter.DT_MAX:
         await ctxt.redis.zrem(DELAYED_REFRESH_KEY, zset_subkey)
-    else:
-        await ctxt.redis.zadd(
-            DELAYED_REFRESH_KEY,
-            **{zset_subkey: best_bet.timestamp()},
-        )
+        return
+
+    in_ten_minutes = utils.utcnow() + datetime.timedelta(
```
jd, line 50: if you change `DELAYED_REFRESH_MIN_MINUTES` then this variable name is wrong.

mergify_engine/delayed_refresh.py
```diff
@@ -44,11 +45,18 @@ async def plan_next_refresh(match: rules.RulesEvaluator, ctxt: context.Context)
...
+        minutes=DELAYED_REFRESH_MIN_MINUTES
+    )
+    if best_bet < in_ten_minutes:
+        best_bet = in_ten_minutes
```
jd, line 54: I'm not sure I get the optimization here.
It's 12:03.
If the next best bet is today at 12:10 because the condition is `>=12:10`.

Now you're saying that since 12:10 is < than (12:03+10min), we should reschedule at 12:13.

Alright, but what's the point? What does that optimize in the end?

You're still refreshing the PR

Worst than that, if the PR has the condition: `>=12:10` and `<=12:20` and it's 12:09, then you're going to schedule the refresh as 12:09+10min = 12:19, which gives a very narrow window of 1 minute to evaluate the rule to get it true.

I feel like I miss something?

jd (changes requested):
Do you think it's worth doing this, in the sense that

mergify_engine/delayed_refresh.py
sileht: You feel I miss something! 

I just reread and this clearly didn't do what I expected. I wanted to ensure we didn't refresh a pull request every seconds/minutes but every 5 minutes minimum.

PR #2771, opened 2021-06-16
feat(conditions): add datetime fields condition

sileht:
## refactor(tests): correctly manipule the time

Since we freeze the time, we should move it during tests. Otherwise if
code use utcnow(), tests may not work as expected.

Change-Id: I7a622b7d768a7d8f382bce628b6128c1950ba36c

## feat(conditions): add datetime fields condition

* current-datetime
* created-at
* updated-at
* merged-at
* closed-at

Change-Id: I8bdf4821605d3b9efcbf473197294095c3066d02


docs/source/configuration.rst
```diff
@@ -169,6 +169,35 @@ Examples
...
+
+      - name: end of life version 10.0
+        conditions:
+          - base=stable/10.0
+          - -closed
+          - current-datetime>=2021-04-05
+        actions:
+          close:
+            message: |
+              The pull request base branch is end of life.
```
jd, line 199:
```suggestion
              The pull request base branch has reached end-of-life.
```

mergify_engine/tests/unit/rules/test_parser.py
```diff
@@ -206,6 +231,10 @@ def test_search(line, result):
         "current-month=100",
         "current-year=0",
         "current-day=100",
+        "current-day>100",
+        "update-at=7 days 18:00",
+        "update-at>=100",
```
jd, line 236: did you mean:
```suggestion
        "updated-at=7 days 18:00",
        "updated-at>=100",
```
?

jd (changes requested):

jd:
in body of the commit:
> Otherwize
should be
> Otherwise

jd (approved):

PR #2772, opened 2021-06-16
chore(logs): log the tmp queue when removing pull from train

sileht:
Change-Id: I3847b160d283b167c51172dd761e0142f4f373d7


jd (approved):
`chore(queue)` would have been better

PR #2793, opened 2021-06-21
feat(queue): wait for check that are not check-XXX

sileht:
To unqueue a pull request it must unmatch the pull_request_rules or
checks conditions in queue_rules. Otherwise we just wait the conditions
to become true.

Fixes MRGFY-496
Fixes MRGFY-491


mergify_engine/actions/merge_base.py
```diff
@@ -102,6 +102,8 @@ def strict_merge_parameter(v):
 async def get_rule_checks_status(
     ctxt: context.Context,
     rule: typing.Union["rules.EvaluatedRule", "rules.EvaluatedQueueRule"],
+    *,
+    unrelated_as_failure: bool = True,
```
jd, line 106: This name is not obvious to understand. Is there a better name or some docstring to help?

sileht: I changed it

mergify_engine/actions/merge_base.py
```diff
@@ -102,6 +102,8 @@ def strict_merge_parameter(v):
 async def get_rule_checks_status(
     ctxt: context.Context,
     rule: typing.Union["rules.EvaluatedRule", "rules.EvaluatedQueueRule"],
+    *,
+    other_unmatch_conditions_return_failure: bool = True,
```
jd, line 106:
```suggestion
    unmatched_conditions_return_failure: bool = True,
```
"other" does not mean much in this context 😢

mergify_engine/actions/merge_base.py
```diff
@@ -102,6 +102,8 @@ def strict_merge_parameter(v):
 async def get_rule_checks_status(
     ctxt: context.Context,
     rule: typing.Union["rules.EvaluatedRule", "rules.EvaluatedQueueRule"],
+    *,
+    unmatch_conditions_return_failure: bool = True,
```
jd, line 106:
```suggestion
    unmatched_conditions_return_failure: bool = True,
```

jd (changes requested):
You'll get this!!

mergify_engine/actions/merge_base.py
sileht: omg!!!!!

sileht:
```
$ history -10000 | grep sed | tail -3
 9974  sed -i -e 's/unrelated_as_failure/other_unmatch_conditions_return_failure/g' mergify_engine/**/*.py
 9985  sed -i -e 's/other_unmatch_conditions_return_failure/unmatch_conditions_return_failure/g' mergify_engine/**/*.py
 9996  sed -i -e 's/unmatch_conditions_return_failure/unmatched_conditions_return_failure/g' mergify_engine/**/*.py
```

jd (approved):

PR #2794, opened 2021-06-21
feat(queue): allow to use attributes from original pull request

sileht:
It doesn't make sense to check label, reviewer, author, ... of the
temporary pull request.

This change looks at the original pull request attributes when Rule
evaluator lookup for them.

This make more natural the configuration of queue rules more natural.

Fixes MRGFY-496

Change-Id: I37007d05e1053594ed2bb725cdb6e065397fb83c


mergify_engine/actions/queue.py
```diff
@@ -140,16 +143,31 @@ async def run(
...
                 )
             await car.update_summaries(
                 status, status, queue_rule_evaluated, will_be_reset=need_reset
             )
+        elif car and car.state == "created":
+            if not ctxt.has_been_only_refreshed():
+                # NOTE(sileht): It's not a only refreshed, so we need to
+                # updated the associated transient pull request.
+                # This is mandatory to filter out refresh to avoid unless loop
+                # of refresh between this PR and the transient one
```
jd, line 158:
```suggestion
                # NOTE(sileht): It's not only refreshed, so we need to
                # update the associated transient pull request.
                # This is mandatory to filter out a refresh to avoid useless loop
                # of refreshes between this PR and the transient one.
```

jd (changes requested):

jd (approved):

PR #2819, opened 2021-06-30
fix(exceptions): ignore per_page arg

jd:
Fixes MERGIFY-ENGINE-27P


mergify_engine/exceptions.py
```diff
@@ -67,7 +67,7 @@ def should_be_ignored(exception: Exception) -> bool:
 
         # NOTE(sileht): a repository return 404 for /pulls..., so can't do much
         if exception.status_code == 404 and str(exception.request.url).endswith(
-            "/pulls"
+            "/pulls?per_page=100"
```
sileht, line 70: We should use `exception.request.url.path.endswith("/pulls")`

sileht (approved):

PR #2831, opened 2021-07-02
fix: ensure Context.sources is empty on each engine run

sileht:
Since we cache for the duration of a batch, the sources may not be empty
when we start the engine.

This fixes that, so no event will be seen twice.

Change-Id: I39566487afd974407866bef0977137518ebf050b


sileht:
1. fix: ensure Context.sources is empty on each engine run ([#2831](https://github.com/Mergifyio/mergify-engine/pull/2831))
1. feat: new event streaming system ([#2811](https://github.com/Mergifyio/mergify-engine/pull/2811))

jd (approved):

PR #2834, opened 2021-07-02
refactor: move UserTokensUser selection to UserTokens class

sileht:
Change-Id: I106d71188c783385690d6f8238f85f383cd64ee2

PR #2835, opened 2021-07-02
feat(squash): add squash command and action

sileht:
Fixes MRGFY-284

Change-Id: I26d3c9bb5be07768cff182707309176880a18b6c


sileht:
1. refactor: move some Git error logic to Gitter() ([#2833](https://github.com/Mergifyio/mergify-engine/pull/2833))
1. refactor: move UserTokensUser selection to UserTokens class ([#2834](https://github.com/Mergifyio/mergify-engine/pull/2834))
1. feat(squash): add squash command and action ([#2835](https://github.com/Mergifyio/mergify-engine/pull/2835))

mergify_engine/actions/squash.py
```diff
@@ -0,0 +1,99 @@
...
+            )
+
+        if commit_title_and_message is not None:
+            title, message = commit_title_and_message
+            message = f"{title}\n\n{message}"
+        else:
+            message = f"{(await ctxt.pull_request.title)} (#{(await ctxt.pull_request.number)})"
+            message += "\n\n".join(
+                [commit["commit"]["message"] for commit in await ctxt.commits]
+            )
```
jd, line 83: Pretty sure this is not how GitHub formats the squash message by default.

mergify_engine/squash_pull.py
```diff
@@ -0,0 +1,160 @@
...
+
+    git = gitter.Gitter(ctxt.log)
+
+    try:
+        await git.init()
+        # NOTE(sileht): Bump the repository format. This ensures required
+        # extensions (promisor, partialclonefilter) are present in git cli and
+        # raise an error if not. Avoiding git cli to fallback to full clone
+        # behavior for us.
+        await git("config", "core.repositoryformatversion", "1")
```
jd, line 63: wondering if we shouldn't do this in `git.init` at this point?

mergify_engine/squash_pull.py
```diff
@@ -0,0 +1,160 @@
...
+    if ctxt.pull["commits"] <= 1:
+        return
+
+    try:
+        users = await user_tokens.UserTokens.select_users_for(ctxt, bot_account)
+    except user_tokens.UserTokensUserNotFound:
+        raise SquashFailure(
+            f"Unable to squash: user `{bot_account}` is unknown. "
+            f"Please make sure `{bot_account}` has logged in Mergify dashboard."
+        )
```
jd, line 139: We have this elsewhere too, might be great to refactor at some point.

mergify_engine/squash_pull.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+        else:
+            return
+
+    ctxt.log.warning("unable to squash pull request: no tokens are valid")
+
+    if ctxt.pull_from_fork and ctxt.pull["base"]["repo"]["private"]:
+        raise SquashFailure(
+            "Squashing a branch for a forked private repository is not supported by GitHub"
+        )
```
jd, line 158: Why is that so late in the process? We should do this first thing to bypass the retry loop?

jd (changes requested):

mergify_engine/squash_pull.py
sileht: When we wrote this for branch_updater, we was not 100% confident about the conditions for this to work or not.
That why we wrote the message only after we try all tokens.

jd: Can we check the log in prod to see if that happen with the branch updater?

jd (approved):

jd:
@mergifyio update

mergify_engine/squash_pull.py
sileht: I don't think we can catch that in logs.

jd:
@mergifyio update

PR #2836, opened 2021-07-02
ci: trigger the production pipeline workflow

sileht:


jd (approved):
also put in the signal repo?

PR #2839, opened 2021-07-02
ci: only check codeQL on push event

sileht:
The job takes 10 minutes and we don't require it to merge PR.

So just tests all commits of the main branch.

Change-Id: I72a8ada8f1d2521355332ea212f3b16767bf3eb3


jd:
Are we going to be notified if anything is "found" but it?

sileht:
> Are we going to be notified if anything is "found" but it?

Yes by email

PR #2844, opened 2021-07-05
chore(deps): remove unused aiofiles

jd:


sileht:
This is an optional deps of fastapi for its staticfiles modules used by the installer.

jd:
I see, it's actually needed by starlettte.
The horror. The good news is that it's going to go away in the next starlette release.

PR #2847, opened 2021-07-07
fix: workaround to detect closed PR

sileht:
GitHub automerge doesn't always close pull request when it merge them...

This is a known GitHub bug, but GitHub is not in an hurry to fix it.

This is a workaround that consider merged pull request as closed.

Fixes MERGIFY-ENGINE-26Z
Change-Id: Ie2c44ce9c049a6c786290834e6da7614104ab9a2

Change-Id: I82989fb9562ff085c2089ed03a55e3c93f2db610

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -732,7 +732,7 @@ class Context(object):
             return self.pull["merged"]
 
         elif name == "closed":
-            return self.pull["state"] == "closed"
+            return self.closed
 
         elif name == "milestone":
             return (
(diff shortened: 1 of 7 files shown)


mergify_engine/context.py
```diff
@@ -1154,6 +1151,10 @@ async def files(self) -> typing.List[github_types.GitHubFile]:
         self._cache["files"] = files
         return files
 
+    @property
+    def closed(self) -> bool:
+        return self.pull["state"] == "closed" or self.pull["merged"]
```
jd, line 1156: Do we want to have a comment on why this exists in the first place?

jd (approved):

PR #2848, opened 2021-07-07
chore: add note about github automerge

sileht:
Change-Id: I2f438230441a10b1ed25ede6568a71593855de89

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -1153,6 +1153,8 @@ class Context(object):
 
     @property
     def closed(self) -> bool:
+        # NOTE(sileht): GitHub automerge doesn't always close pull requests
+        # when it merges them.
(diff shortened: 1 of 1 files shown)


mergify_engine/context.py
```diff
@@ -1153,6 +1153,8 @@ async def files(self) -> typing.List[github_types.GitHubFile]:
 
     @property
     def closed(self) -> bool:
+        # NOTE(sileht): GitHub automerge doesn't always close pull request when
+        # it merge them.
```
jd, line 1157:
```suggestion
        # NOTE(sileht): GitHub automerge doesn't always close pull requests when
        # it merges them.
```

jd (changes requested):

jd (approved):

Issue #456, opened 2019-07-18
Don't try to merge draft pull requests


alecmocatta:
GitHub introduced the [draft pull requests feature](https://github.blog/2019-02-14-introducing-draft-pull-requests/) earlier this year.

Mergify currently tries to merge draft pull requests, resulting in the failure:

```
Repository settings are blocking automatic merging

GitHub error message: Pull Request is still a draft
```

beatngu13:
It would be great if one could specify this as a condition. In our case, we also don't want to perform e.g. rebase when the given PR is a draft.

PR #2867, opened 2021-07-12
fix: don't use elapsed time in summary

sileht:
This is confusing, for example the table said thing like:

Number | Queued
\#123   | 1 seconds ago
\#456   | 30 minutes ago
...

But this is updated only when we got a event for the pull request.
So it's always out of date.

This change just puts the timestamp, it's less fancy but at least it's
right.


mergify_engine/date.py
```diff
@@ -99,35 +98,10 @@ def fromtimestamp(timestamp: float) -> datetime.datetime:
...
-        parts.append(_number_with_unit(seconds, "second"))
-
-    return _pretty_join(parts)
+def pretty_datetime(dt: datetime.datetime) -> str:
+    if dt.tzinfo != datetime.timezone.utc:
+        raise RuntimeError(
+            f"Wrong timezone, got {dt.tzinfo} instead of {datetime.timezone.utc}"
+        )
+    dt = dt.replace(tzinfo=None, second=0, microsecond=0)
+    return f"{dt} UTC"
```
jd, line 107: I'm not sure error-ing out is great.
Just convert to UTC and return the result?

jd (changes requested):

jd (approved):

PR #2878, opened 2021-07-13
chore(queue): avoid useless pull body edit

sileht:
Change-Id: I7ab4c5895fb3f4b92a616b60722a6d98dd3d81bf


jd (approved):
I'm surprised this does not change any existing test record.

PR #2889, opened 2021-07-15
fix(woker-status): display owner login instead of id

sileht:
Change-Id: I70e32d82e34639c0f93b2463da3401980bb66db5


jd (approved):

sileht:
@mergifyio refresh

PR #2899, opened 2021-07-15
fix(worker): add metric on the number of msg read per bucket

jd:


jd:
Actually I think `engine.streams.batch-size` is the same, no?

mergify_engine/worker.py
```diff
@@ -557,6 +557,7 @@ async def _consume_buckets(
             )
 
             messages = await self.redis_stream.xrange(bucket_sources_key)
+            statsd.histogram("engine.buckets.read_size", len(messages))  # type: ignore[no-untyped-call]
```
sileht, line 560: it's the same as engine.streams.batch-size

sileht: Maybe we should drop engine.streams.batch-size and keep your's !

sileht: > Actually I think `engine.streams.batch-size` is the same, no?

The only diff is `engine.streams.batch-size` is not computed for pull number 0

jd: Good point, I've changed it that way.

sileht (approved):

PR #2917, opened 2021-07-15
feat(worker): count the number of times we got preempted processing buckets

sileht:
## feat(worker): count the number of times we got preempted processing buckets

Change-Id: I2a0d3a0769863dcfeeeca291a5b26700fc4d4266

## fix(worker): use tags for engine.buckets.backlog backlog

Change-Id: Id782e6d8e545514505c8f302557d002383fbff33

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -878,20 +878,27 @@ class Worker:
 
                 # TODO(sileht): maybe we can do something with the bucket scores to
                 # build a latency metric
-                bucket_backlog = 0
+                bucket_backlog_low = 0
                 bucket_backlog_high = 0
                 for org_bucket, _ in org_buckets:
                     bucket_contents = await self._redis_stream.zrangebyscore(
                         org_bucket, min=0, max="+inf", withscores=True
                     )
-                    bucket_backlog += len(bucket_contents)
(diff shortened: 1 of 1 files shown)


mergify_engine/worker.py
```diff
@@ -489,6 +494,16 @@ async def _consume_buckets(
             else:
                 break
 
+            if (time.monotonic() - started_at) >= config.BUCKET_PROCESSING_MAX_SECONDS:
+                low_prio_threshold = get_low_priority_minimal_score()
+                if _bucket_score < low_prio_threshold:
+                    statsd.increment("engine.buckets.preempted.high-prio")
+                else:
+                    statsd.increment("engine.buckets.preempted.low-prio")
+                break
```
jd, line 503:
```suggestion
            if (time.monotonic() - started_at) >= config.BUCKET_PROCESSING_MAX_SECONDS:
                low_prio_threshold = get_low_priority_minimal_score()
                prio = "high" if _bucket_score < low_prio_threshold else "low"
                statsd.increment("engine.buckets.preempted", tags=[f"priority:{prio}"])
                break
```

I feel like we should also change the other metrics to set the priority as a tag.

jd (changes requested):

jd (approved):

PR #2934, opened 2021-07-20
fix(worker): reset Context() cache on each run_engine()

sileht:
run_engine() can be ran multiple times for the same pull requests with
the new worker code. So we must be sure we the Context cache is
cleared each time.

Change-Id: I23799c68a6d46e6184493fd0eb51c0e3bb22d4dc


mergify_engine/worker.py
```diff
@@ -233,6 +233,10 @@ async def run_engine(
         # NOTE(sileht): Reset sources as a pull request may be evaluated multiple
         # times during worker batch.
         ctxt.sources = []
+        # NOTE(sileht): A pull request may be reevaluated during one call of
+        # consume_buckets(), so we need to clear the cache to ensure we get the
+        # last photo of the pull request
```
jd, line 238:
```suggestion
        # last snapshot of the pull request
```

jd (changes requested):

jd (approved):

PR #2947, opened 2021-07-23
fix(queue): improve the type of merge_bot_account

sileht:
This sync the type with the merge action.

Change-Id: I3ca8ee880451b0052cb97a090382d7b99ed36e50


jd (dismissed):
doc change?

jd (approved):

PR #2949, opened 2021-07-23
fix(bot-account): render bot_account template

sileht:
All bot account are describe as template, but there are not all
templates.

This change fixes that.

Change-Id: I30887768624e8c4a3bc2691b7dff9621d0db1ef4


mergify_engine/actions/utils.py
```diff
@@ -52,6 +66,26 @@ async def validate_bot_account(
...
+        )
+
+    try:
+        bot_account = typing.cast(
+            github_types.GitHubLogin, GitHubLoginSchema(bot_account)
+        )
+    except voluptuous.Invalid as e:
+        raise RenderBotAccountFailure(
+            check_api.Conclusion.FAILURE,
+            f"Invalid {option_name} template",
```
jd, line 85:
```suggestion
            f"Invalid {option_name} value",
```
The template is ok but the value rendered is wrong.

jd (changes requested):

jd (approved):

PR #2961, opened 2021-07-30
chore(rules): add type to QueueRules

sileht:
Change-Id: Ida624bfab115548cf2c0b5dc8fa31aedda62cb78

PR #2960, opened 2021-07-30
fix(queue): preserve order when train is shrunk

sileht:
Change-Id: I9ca0063b0794320f502e3e1e92db51ec2bdf38b0

PR #2962, opened 2021-07-30
feat(queue): sync configuration of queued PR

sileht:
Related MRGFY-522

Change-Id: I59cae5899bfa9922774f997b54eddf9d961a5b81

### DIFF
--- a/mergify_engine/queue/merge_train.py
+++ b/mergify_engine/queue/merge_train.py
@@ -789,11 +789,14 @@ class Train(queue.QueueBase):
             )
             return
 
+        queue_rules = mergify_config["queue_rules"]
+
         # NOTE(sileht): workaround for cleaning unwanted PRs queued by this bug:
         # https://github.com/Mergifyio/mergify-engine/pull/2958
         await self._remove_duplicate_pulls()
 
-        await self._populate_cars(mergify_config["queue_rules"])
+        await self._sync_configuration_change(queue_rules)
+        await self._populate_cars(queue_rules)
         await self._save()
         self.log.info("train cars refreshed")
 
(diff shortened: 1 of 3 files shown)


sileht:
1. fix(queue): preserve order when train is skrink ([#2960](https://github.com/Mergifyio/mergify-engine/pull/2960))
1. chore(rules): add type to QueueRules ([#2961](https://github.com/Mergifyio/mergify-engine/pull/2961))
1. feat(queue): sync configuration of queued PR ([#2962](https://github.com/Mergifyio/mergify-engine/pull/2962))

mergify_engine/queue/merge_train.py
```diff
@@ -812,6 +815,20 @@ async def _remove_duplicate_pulls(self) -> None:
...
+            if queue_rule is None:
+                # NOTE(sileht): We just slice the cars list here, so when the
+                # car will be recreated if the rule doesn't exists anymore, the
+                # failure will be reported properly
+                await self._slice_cars_at(i)
+            else:
+                # NOTE(sileht): Update the queue config, so if
+                # speculative_checks change the train will grow or shrink
+                # instantly
+                pseudo_car.config["queue_config"] = queue_rule.config
```
jd, line 830: Do we actually want this, rather than reading the config from the config file?

jd (changes requested):
Does this solve MRGFY-522?

mergify_engine/queue/merge_train.py
sileht: A long time ago this was not possible as we didn't have the configuration everywhere we need it. This may be possible now.
But that not a easy refactoring, `current` values has to be stored for the comparaison. So the config will move from per Car to per Train. I keep MRGFY-522 open to do it.

jd (approved):

PR #2990, opened 2021-08-09
fix(duplicate): report git error even if not categorized

sileht:
User didn't like the current message "Backport failed." even
we quickly categorize the error when this occurs.

This change reports the reason to users, this make them happy and help
to known if we already correct/categorize that one or not.

Change-Id: I99d163634a717b6e191523500edb2e8caed751b2


jd (approved):

sileht:
semantic!!!

sileht:
@mergifyio rebase

PR #2997, opened 2021-08-10
refactor(rules): allow to parse multiple pull requests

sileht:
This updates the rule conditions objects to able multiple
context.PullRequest object at once.

This will be used for check the queue rules with future batch features.

Change-Id: Id33ce6830d92aa49f3f630b3714085bbde47cf19

### DIFF
--- a/mergify_engine/rules/__init__.py
+++ b/mergify_engine/rules/__init__.py
@@ -99,12 +99,24 @@ class RuleCondition:
         rc.partial_filter.value_expanders = self.partial_filter.value_expanders
         return rc
 
-    async def __call__(self, obj: filter.GetAttrObjectT) -> bool:
+    async def __call__(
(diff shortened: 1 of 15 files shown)


mergify_engine/rules/__init__.py
```diff
@@ -99,12 +99,24 @@ def copy(self) -> "RuleCondition":
         rc.partial_filter.value_expanders = self.partial_filter.value_expanders
         return rc
 
-    async def __call__(self, obj: filter.GetAttrObjectT) -> bool:
+    async def __call__(
+        self,
+        obj_or_pull_requests: typing.Union[
+            filter.GetAttrObjectT, typing.List[context.BasePullRequest]
+        ],
```
jd, line 106: Is there any way we can always pass a list?

PR #3014, opened 2021-08-13
fix(conditions): fix evaluation of multiple pull requests

sileht:
Change-Id: I76990df1f6e50eb1a30c89a992d85f5f2c577a6e


sileht:
This is more complicated to fix finally.

PR #3025, opened 2021-08-16
ci: add releaser GitHub action

sileht:
Tools to release `main` branch from a GitHub Workflow

To release a new version use:

* gh workflow run Releaser

Change-Id: Ib37c2df8917e5f711bb8405c89dc5372daa20530

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -1,0 +1,42 @@
+# Release command:
+#
+#  gh workflow run Releaser
+#
+name: Releaser
+permissions: write-all
+on:
+  workflow_dispatch:
+
+jobs:
+  releaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v2.3.4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/releaser.yml
```diff
@@ -0,0 +1,46 @@
...
+          python-version: 3.9.6
+
+      - name: Build release notes 📝
+        id: relnote
+        run: |
+          pip install reno
+
+          pwd
+          ls -la
+          git version
```
jd, line 31: Do you want to keep this?

.github/workflows/releaser.yml
```diff
@@ -0,0 +1,46 @@
...
+
+          git fetch --tag
+          VERSION=$(reno -q semver-next)
+
+          git tag $VERSION
```
jd, line 36: do we want to sign the tag?
I mean do we even need to push a tag if we use the gh release action?

sileht: Yes, `reno` needs the tag.

jd (approved):

PR #3030, opened 2021-08-16
fixes(context): handle body=null

sileht:
Fixes MERGIFY-ENGINE-29S

Change-Id: I10fb088d5979bc73958ddebed87c4c963d479002


jd (dismissed):
1. Do we have a correct typing for this?
2. Can we have a test?

PR #3033, opened 2021-08-19
fix(update): fix detection of synchronized event done by Mergify

sileht:
Change-Id: I3f35a881f47b1e130f31cddf8a3b4e4a8f9855f7


sileht:
This bug is exercised by a test in a future PR

jd (approved):

PR #3045, opened 2021-08-20
fix(queue): display the speculative checks pr number

sileht:
Change-Id: Id59d82c5d312fa170c306e8a0e50cb11ea3c7434


jd (approved):
Maybe use a named tuple for better readibility?

PR #3050, opened 2021-08-23
feat(redis): store active users in redis

ErwanSimonetti:
## Description of the change

Users that push code to their repositories, make comments, pull requests etc should be stored in redis to track which users are active thus using Mergify.

## Related issues

Fixes MRGFY-511

### Development

- [X] All checks must pass (Semantic Pull Request, pep8, requirements, unit tests, functional tests, security checks, …)
- [X] The code changed/added as part of this pull request must be covered with tests
- [X] Hotfixes must have a link to Sentry issue and the ``hotfix`` label
- [X] Features must have a link to a Linear task

### Code Review


mergify_engine/github_events.py
```diff
@@ -35,6 +36,15 @@
 LOG = daiquiri.getLogger(__name__)
 
 
+async def store_active_users(
+    redis_cache: utils.RedisCache, repo: github_types.GitHubRepository
+) -> None:
+    id_owner = repo["owner"]["id"]
+    id_repo = repo["id"]
+    await redis_cache.hset(id_owner, "timestamp", time.time())
```
sileht, line 44: We have many other things stored un Redis, so we should prefix it with something like `f"active-users~{id_owner}"`

mergify_engine/github_events.py
```diff
@@ -35,6 +36,15 @@
...
+    await redis_cache.hset(id_owner, "repo_id", id_repo)
```
jd, line 45: There's a couple of issues here:
- this is going to set a key named something like `12345` into Redis, which is not really great: there's plenty of keys here, a key named `12345` does not really help to understand what it stores. You should names your keys with something like `active_users~{owner_id}`
- Storage works. Now, you also have to think about query time. So what's the typical query you'll get? I can tell you, it's "give me the list of users we got active over the last month for this organization".
Here are a few hints: first problem, you don't store the organization. Second problem, your data structure here does not help you answering the query efficiently.

mergify_engine/github_events.py
```diff
@@ -144,6 +154,7 @@ async def filter_and_dispatch(
         repo_id = event["repository"]["id"]
         repo_name = event["repository"]["name"]
         pull_number = event["pull_request"]["number"]
+        store_active_users(redis_cache, event["repository"])
```
jd, line 157: You shouldn't call this function at different places. Ultimately it should be only called from only one place.

jd (changes requested):
Tests look good (I didn't review them in details but overall I think that's the good direction).

mergify_engine/github_events.py
sileht: We also need to store who has done the action we received as an event. For example in this event: https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pullrequestevent we want to track who has open/edit/close/... the pull request and when. (e.g. the `["pull_request"]["author"]["login"]` attribute of the event)

sileht (dismissed):

jd:
Closing for #3071 instead.

PR #3056, opened 2021-08-24
feat(tests): save where a test is recorded

sileht:
This change saves where a test is recorded and use a new endpoint to get
events from GitHub.

Change-Id: I5ca7833c7b3b1711caab423a696368e0412e71a6


sileht:
@Mergifyio rebase

jd (approved):

PR #3066, opened 2021-08-25
fix(tests): rename event files

sileht:
Rename some events fixture file to set event_type correctly in tests

Change-Id: I99f0df873243bd7f8e7fbaf2b0b5158d5a16c5f7


jd (approved):
I'm worried we didn't have to change the code?

PR #3069, opened 2021-08-25
fix(command): only react to @mergifyio

sileht:
950753e26 has introduced a regression making mergify anwser to some
comments it shouldn't.

Change-Id: I3d86274118e9017e42384973e53d4c375bc2e567

### DIFF
--- a/mergify_engine/engine/commands_runner.py
+++ b/mergify_engine/engine/commands_runner.py
@@ -192,6 +192,9 @@ async def handle(
     user: typing.Optional[github_types.GitHubAccount],
     rerun: bool = False,
 ) -> None:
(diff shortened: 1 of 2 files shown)


jd (approved):

mergify_engine/engine/commands_runner.py
```diff
@@ -192,6 +192,9 @@ async def handle(
     user: typing.Optional[github_types.GitHubAccount],
     rerun: bool = False,
 ) -> None:
+    if "@mergify " not in comment.lower() and "@mergifyio " not in comment.lower():
```
jd, line 195: This should use variable, right?

PR #3083, opened 2021-09-01
ci(releaser): add dry_run flag

sileht:
This allows to run the changelog generation without pushing a tag

Change-Id: I0e0ee917eca12f47ba98fdcf160deaef77559f81

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -6,7 +6,11 @@ name: Releaser
 permissions: write-all
 on:
   workflow_dispatch:
-
+    inputs:
+      dry_run:
+        description: "don't push thing"
+        required: false
+        default: false
 jobs:
   releaser:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/releaser.yml
```diff
@@ -53,8 +54,15 @@ jobs:
               sed -e '/^\.\./,+1d' -e '1,4d' > ${{ github.workspace }}-CHANGELOG.txt
           echo "::set-output name=VERSION::$VERSION"
 
-      - name: Release 🏎️
+      - name: Push git tag 🏎️
+        if: ${{ !github.event.inputs.dry_run }}
+        run: |
+          git tag ${{ steps.relnote.outputs.VERSION }}
```
jd, line 60: should we sign tag here?

sileht: Well, this means creates and manage a gpg key. If you want the green tick on GitHub UI, you also need to associate this gpg key to a GitHub user.

jd (approved):

PR #3087, opened 2021-09-02
ci(releaser): tag version before running reno

sileht:
Change-Id: I4e9bf6c75532b88e38507e8df9afa51f4efad00b


jd (commented):
why?
Not using reno sounds like you might think you're going to release X when reno is going to say Y.

sileht:
> why?
> Not using reno sounds like you might think you're going to release X when reno is going to say Y.

reno need the tag to exists, but I can do something else.

jd (approved):

PR #3095, opened 2021-09-02
fix(debug) don't raise error when pull request does not exists

sileht:
Change-Id: I46ae626013030d3087eeaed6faaef9be35d1a1f2


mergify_engine/debug.py
```diff
@@ -306,9 +306,13 @@ async def report(
...
             )
-            ctxt = await repository.get_pull_request_context(
-                github_types.GitHubPullRequestNumber(int(pull_number))
-            )
+            try:
+                ctxt = await repository.get_pull_request_context(
+                    github_types.GitHubPullRequestNumber(int(pull_number))
+                )
+            except http.HTTPNotFound:
+                print(f"pull request `{url}` does not exists")
```
jd, line 314:
```suggestion
                print(f"Pull request `{url}` does not exist")
```

jd (changes requested):

jd:
@mergifyio squash first-commit

jd (approved):

PR #3106, opened 2021-09-06
Pull request for track-other-than-sender-id

sileht:
## feat(seats): add more field where we track user

Change-Id: I3b0c91e436736bd6edf569557dee02b6b767fa13

## feat(seats): don't count mergify and GitHub web-flow user

Change-Id: I5ea1ae9f9f82ab96010db02f1ecfb7e935887e5a

### DIFF
--- a/mergify_engine/count_seats.py
+++ b/mergify_engine/count_seats.py
@@ -135,6 +135,11 @@ async def store_active_users(
     users = {}
 
     def _add_user(user: github_types.GitHubAccount) -> None:
+        if user["id"] == config.BOT_USER_ID:
+            return
+        elif user["login"] == "web-flow":
+            return
+
         users[user["id"]] = user["login"]
 
     if event_type == "push":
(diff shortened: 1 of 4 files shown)


mergify_engine/count_seats.py
```diff
@@ -132,27 +132,45 @@ async def store_active_users(
...
+        _add_user(typed_event["sender"])
+
+        repo_key = _get_active_users_key(typed_event["repository"])
+        for user_id, user_login in users.items():
+            if user_id == config.BOT_USER_ID:
+                continue
+            elif user_login == "web-flow":
+                continue
+            user_key = f"{user_id}~{user_login}"
+            await redis_cache.zadd(repo_key, **{user_key: time.time()})
```
jd, line 173: Using a pipe?

jd (dismissed):

mergify_engine/count_seats.py
sileht: done

jd (approved):

PR #3112, opened 2021-09-07
fix(refresh): don't strip the pull request number

sileht:
We removed the pull request number from the refresh event.
Making Mergify refreshing more pull request that needed.

Change-Id: I884b83140d0dc09a320b52769de2a24c9bef6349


mergify_engine/utils.py
```diff
@@ -185,23 +185,30 @@ def __init__(self, key: str, value: typing.Any):
...
-    pull_request_number: typing.Optional[github_types.GitHubPullRequestNumber] = None,
-    ref: typing.Optional[github_types.GitHubRefType] = None,
-    action: github_types.GitHubEventRefreshActionType = "user",
+    refresh_options: PartialGitHubEventRefresh,
 ) -> None:
     # Break circular import
     from mergify_engine import github_events
 
+    if not refresh_options:
+        raise ValueError("refresh_options must be set")
```
jd, line 204: I'd rather remove the `total=False` and just pass `None` as value.
Or just go ahead and use keyword arguments, no?

sileht: Hum, since I have one method of each kind of refresh I can keep keyword arguments I guess.

jd (approved):

PR #3113, opened 2021-09-07
ci(releaser): install pandoc correctly

sileht:
We currently install the python lib called pandoc instead
of the pandoc binary.

Change-Id: Iff4943a752156af32b25aa821ddba7085e06aabf


jd (changes requested):
why?

sileht:
> why?

the pandoc of pypi is not the official pandoc

jd (approved):

PR #3117, opened 2021-09-07
fix(seats): filter returned seats via API correctly by owner_id

sileht:
Change-Id: I16eb539ee9da0ceec5a4cf9ae335df78d09ae3de

### DIFF
--- a/mergify_engine/count_seats.py
+++ b/mergify_engine/count_seats.py
@@ -224,12 +224,15 @@ class Seats:
         redis_cache: utils.RedisCache,
         write_users: bool = True,
         active_users: bool = True,
+        owner_id: typing.Optional[github_types.GitHubAccountIdType] = None,
     ) -> "Seats":
         seats = cls()
         if write_users:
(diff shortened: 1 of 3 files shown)


mergify_engine/count_seats.py
```diff
@@ -224,12 +224,15 @@ async def get(
         redis_cache: utils.RedisCache,
         write_users: bool = True,
         active_users: bool = True,
+        owner_id: typing.Optional[github_types.GitHubAccountIdType] = None,
     ) -> "Seats":
         seats = cls()
         if write_users:
+            if owner_id is not None:
+                raise RuntimeError("Can't get `write_users` if `owner_id` is set ")
```
jd, line 232:
```suggestion
                raise RuntimeError("Can't get `write_users` if `owner_id` is set")
```

jd (changes requested):

jd:
@mergifyio squash first-commit

jd (approved):

PR #3133, opened 2021-09-10
fix(label): make sure ctxt.repository.ensure_label_exists() is run

sileht:
Change-Id: I1641f9dfb7295be9e1e034bceeb2a0e38306868f


jd (dismissed):
fix title plz

jd (approved):

PR #3141, opened 2021-09-10
feat(subscription): add PUBLIC_REPOSITORY function to desactivate acc…

Zerdanes:
adding new function on engine side with tests to desactivate account
without feature as the same way as PRIVATE_REPOSITORY

## Description of the change

> Description here

## Related issues

> Fix #xyz

### Development

- [x] All checks must pass (Semantic Pull Request, pep8, requirements, unit tests, functional tests, security checks, …)
- [x] The code changed/added as part of this pull request must be covered with tests
- [x] Hotfixes must have a link to Sentry issue and the ``hotfix`` label
- [x] Features must have a link to a Linear task

### Code Review


mergify_engine/engine/__init__.py
```diff
@@ -239,15 +239,21 @@ async def run(
...
-        )
+    if ctxt.pull["base"]["repo"]["private"]:
+        if not ctxt.subscription.has_feature(subscription.Features.PRIVATE_REPOSITORY):
+            ctxt.log.info("mergify disabled: private repository")
+            return check_api.Result(
+                check_api.Conclusion.FAILURE,
+                title="Mergify is disabled",
+                summary=ctxt.subscription.reason,
+            )
+        elif not ctxt.subscription.has_feature(subscription.Features.PUBLIC_REPOSITORY):
```
jd, line 250: This is only going to be checked if the repo is private, not public :)

We want this check to be done if the repo is public.

mergify_engine/tests/unit/test_subscription.py
```diff
@@ -174,11 +174,54 @@ async def test_active_feature(redis_cache):
...
         },
     )
+
     assert sub.has_feature(subscription.Features.PRIVATE_REPOSITORY) is True
     assert sub.has_feature(subscription.Features.LARGE_REPOSITORY) is True
     assert sub.has_feature(subscription.Features.PRIORITY_QUEUES) is False
 
 
+@pytest.mark.asyncio
+async def test_active_feature(redis_cache):
```
jd, line 184: This is testing that `has_feature` works so it's a duplicate of the test above.

For this feature, you need a functional test.

jd (dismissed):

mergify_engine/tests/unit/test_subscription.py
sileht: I will prepare the repository for generating the fixtures

mergify_engine/tests/unit/test_subscription.py
```diff
@@ -155,7 +155,7 @@ async def test_active_feature(redis_cache):
         "friend",
         frozenset([subscription.Features.PRIORITY_QUEUES]),
     )
-    assert sub.has_feature(subscription.Features.PRIORITY_QUEUES) is False
+    assert sub.has_feature(subscription.Features.PRIORITY_QUEUES) is True
```
sileht, line 158: Not sure I get why this one change.

sileht: The value was obviously wrong before.

jd: It changed because we don't check `self.active` in `has_feature` anymore.
I agree the test before was a bit wrong but it was the logic we implemented. I think it's better now.

jd (commented):
This is good, we just need a functional test.

mergify_engine/tests/unit/test_subscription.py
sileht: glad we drop `self.active` then :)

mergify_engine/tests/functional/test_subscription.py
```diff
@@ -0,0 +1,47 @@
...
+        await self.subscription._save_subscription_to_cache()
+        await self.setup_repo()
+        pr, _ = await self.create_pr()
+        await context.Context.create(self.repository_ctxt, pr)
+        await self.run_engine()
+
+        assert (await self.get_pull(pr["number"]))["state"] == "open"
+
+        expected = "Abuse !"
+        assert expected == self.subscription.reason
```
sileht, line 47: This only check the reason on the subscription object.
What we want is to check this reason has been posted in a  check_runs on GitHub.

sileht (changes requested):

mergify_engine/tests/functional/test_subscription.py
sileht: You can take a look at this test: https://github.com/Mergifyio/mergify-engine/blob/main/mergify_engine/tests/functional/test_engine.py#L41-L54

mergify_engine/tests/functional/test_subscription.py
```diff
@@ -0,0 +1,47 @@
...
+            )
+            if self.SUBSCRIPTION_ACTIVE
+            else frozenset([]),
+        )
+
+        await self.setup_repo()
+        p, _ = await self.create_pr()
+        await self.run_engine()
+
+        p = await self.get_pull(p["number"])
```
sileht, line 42: I think you can keep the previous `p`, the pull request dict has not changed

mergify_engine/tests/functional/test_subscription.py
```diff
@@ -0,0 +1,47 @@
...
+        ctxt = await context.Context.create(self.repository_ctxt, p, [])
+        checks_abuse = await ctxt.pull_engine_check_runs
+        assert len(checks_abuse) == 1
+        assert checks_abuse[0]["name"] == "Summary"
+        assert "Abuse" == checks_abuse[0]["output"]["summary"]
```
sileht, line 47: I think we should also assert on the `title`

sileht (changes requested):

sileht (approved):

PR #3144, opened 2021-09-13
chore(bot-account): cache bot account permission

sileht:
Change-Id: I7a7b1539ec05bba6fed1ed660688df751dda763b


mergify_engine/context.py
```diff
@@ -100,6 +100,24 @@ async def get_user_tokens(self) -> user_tokens.UserTokens:
...
+            typing.Optional[github_types.GitHubAccountIdType],
+            await self.redis.hget(self.USER_ID_MAPPING_CACHE_KEY, login),
+        )
+        if user_id is not None:
+            return user_id
+
+        user = typing.cast(
+            github_types.GitHubAccount, await self.client.item(f"/users/{login}")
+        )
+        await self.redis.hset(self.USER_ID_MAPPING_CACHE_KEY, login, user["id"])
```
jd, line 118: What about caching the whole `user` dataset?
There's a chance we need another field at some point 🤷

sileht: hum, why not

jd (approved):

PR #3147, opened 2021-09-13
chore(action): use new method to declare schemas

sileht:
This allows all schema to use partial_validation and avoid any
bug/mistake with schema that use mutable object as default (eg: []).

Change-Id: I68e9ce851b01ee3a38cc367a481064bb45e37c82

### DIFF
--- a/mergify_engine/actions/__init__.py
+++ b/mergify_engine/actions/__init__.py
@@ -13,7 +13,7 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-
+import abc
 import dataclasses
 import enum
 import typing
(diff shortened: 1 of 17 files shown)


mergify_engine/actions/assign.py
```diff
@@ -28,19 +28,23 @@
...
         | actions.ActionFlag.ALWAYS_RUN
     )
 
+    @classmethod
+    def get_config_schema(
+        cls, partial_validation: bool
+    ) -> typing.Dict[typing.Any, typing.Any]:
+        return {
+            # NOTE: "users" is deprecated, but kept as legacy code for old config
+            voluptuous.Required("users", default=[]): [types.Jinja2],
```
jd, line 43: It's not obvious, but you can actually use `default=list` and voluptuous will see it's a callable and call it.

That might be a simpler change?

mergify_engine/actions/__init__.py
```diff
@@ -109,22 +115,16 @@ def get_schema(cls, partial_validation: bool = False) -> ActionSchema:
...
     def command_to_config(string: str) -> typing.Dict[str, typing.Any]:
         """Convert string to dict config"""
         return {}
 
+    @abc.abstractmethod
     async def run(
         self, ctxt: context.Context, rule: "rules.EvaluatedRule"
     ) -> check_api.Result:  # pragma: no cover
-        pass
+        ...
```
jd, line 127: Looks unrelated?

mergify_engine/actions/assign.py
sileht: https://github.com/Mergifyio/mergify-engine/pull/3150

PR #3183, opened 2021-09-15
fix(depends-on): make the order predictable

sileht:
This was creating useless update of the summary just because the set was
not always have the same order.

Change-Id: I5a80a40d0396468975c999430154b4d4512b992a

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -972,18 +972,20 @@ class Context(object):
         re.MULTILINE | re.IGNORECASE,
     )
 
-    def get_depends_on(self) -> typing.Set[github_types.GitHubPullRequestNumber]:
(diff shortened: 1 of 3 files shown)


mergify_engine/context.py
```diff
@@ -972,18 +972,20 @@ async def _get_consolidated_data(self, name: str) -> ContextAttributeType:
         re.MULTILINE | re.IGNORECASE,
     )
 
-    def get_depends_on(self) -> typing.Set[github_types.GitHubPullRequestNumber]:
+    def get_depends_on(self) -> typing.List[github_types.GitHubPullRequestNumber]:
         if self.pull["body"] is None:
             return set()
```
jd, line 977:
```suggestion
            return []
```

jd (changes requested):

jd (approved):

PR #3186, opened 2021-09-15
feat(backport/delete_head_branch): leverage conditions

sileht:
use conditions instead of hardcoded if

Change-Id: Icb55904cde46e686b4dd6977b6d624a68c08f74c


jd (approved):

Psoffie009:
> use conditions instead of hardcoded if
> 
> Change-Id: Icb55904cde46e686b4dd6977b6d624a68c08f74c

PR #3188, opened 2021-09-16
fix(queue): reduce the number of refresh that goes between draft and main pull

sileht:
## fix(tests): don't preempted bucket consumption

Otherwise preemption doesn't occur at the same moment during record and
replay. Making some tests working during record and failing during
replay.

Change-Id: Ife97bc0a4a14d189c9b5393ce677e4453fe7c515

## fix(queue): refresh draft pull if main pull is still queued

We don't need to send a refresh if the pull request has been unqueue.
The draft pull request is closed before the refresh is handled.

Change-Id: Ie5f7ad19be3d590f7a4e147bc235a5ac28978c89

## feat(queue): don't forward all events from main pr to draft pr

If the queue rules doesn't use attributes linked to the original pull
request we don't need to refresh the draft pull request.

Change-Id: If7f3f8eea3979b2d35f4330b37403cb081e446eb

## feat(queue): don't refresh main pull request until we have a final state

Change-Id: I5152930b6ebdeeb94d04464c5cff2398d636e506


mergify_engine/actions/queue.py
```diff
@@ -381,3 +393,13 @@ async def send_signal(self, ctxt: context.Context) -> None:
                 ],
             },
         )
+
+    async def need_draft_pull_request_refresh(self) -> bool:
+        # NOTE(sileht): if the queue rules doesn't use attributes linked to the
+        # original pull request we don't need to refresh the draft pull request
```
jd, line 399:
```suggestion
        # NOTE(sileht): if the queue rules don't use attributes linked to the
        # original pull request we don't need to refresh the draft pull request
```

jd:
@mergifyio squash

jd:
@sileht argh my squash is not going to be friendly with your 2 commits here sorry

jd (dismissed):
fix the comment manually i guess

jd (approved):

PR #3190, opened 2021-09-16
feat(datadog): add tracer

sileht:
Change-Id: Id3de09ed78020e143126f72f09e9c7e020593348


jd (changes requested):
requirements?

jd (commented):
also what about onprem?

mergify_engine/logs.py
```diff
@@ -70,6 +71,17 @@ def add_fields(self, log_record, record, message_dict):
         if worker_id is not None:
             log_record.update({"worker_id": worker_id})
 
+        span = ddtrace.tracer.current_span()
```
jd, line 74: https://github.com/DataDog/dd-trace-py/blob/master/ddtrace/tracer.py#L261

Dockerfile
```diff
@@ -36,5 +36,6 @@ WORKDIR /app
 ENV VIRTUAL_ENV=/venv
 ENV PYTHONUNBUFFERED=1
 ENV DD_DOGSTATSD_DISABLE=1
+ENV DD_APM_ENABLED=0
```
jd, line 39: This controls the Datadog agent, not the tracer library.

mergify_engine/web/asgi.py
```diff
@@ -22,4 +22,5 @@
 if config.SENTRY_URL:
     application = SentryAsgiMiddleware(application)  # type:ignore[assignment]
 
+tracer.configure(service="web")
```
jd: I fear this is going to be conflicting soon with the dashboard.
Rather use `engine-web`.

mergify_engine/worker.py
```diff
@@ -1036,6 +1041,7 @@ async def run_forever() -> None:
 
 
 def main() -> None:
+    tracer.configure(service="worker")
```
jd:
```suggestion
    tracer.configure(service="engine-worker")
```

jd (dismissed):
Change Procfile too or not?

mergify_engine/web/asgi.py
```diff
@@ -22,4 +22,5 @@
 if config.SENTRY_URL:
     application = SentryAsgiMiddleware(application)  # type:ignore[assignment]
 
+ddtrace.config.service = "web"
```
jd, line 25:
```suggestion
ddtrace.config.service = "engine-web"
```

jd (changes requested):

mergify_engine/worker.py
```diff
@@ -1036,6 +1050,7 @@ async def run_forever() -> None:
 
 
 def main() -> None:
+    ddtrace.config.service = "worker"
```
jd, line 1053:
```suggestion
    ddtrace.config.service = "engine-worker"
```

jd (changes requested):

mergify_engine/web/asgi.py
sileht: This starts to be messy.
For agent and process plugin we use:
  - appname:$HEROKU_APP_NAME
  - service:web/worker

And for logging just :
      service: mergify-engine

jd: Well `appname` is Heroku specific and not use across Datadog's product, so we can ignore it basically.

Our current service naming in Datadog is kinda wrong and is not going to give a good experience in APM.

sileht: gotcha

jd (approved):

PR #3202, opened 2021-09-17
fix(worker): do not push back PR with a score

jd:
If a PR receive an event with a low priority, do not push it back because of
this and keep it around.

Only PRs with no events and a low priority event will be scheduled very far.

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -150,14 +150,13 @@ def compute_priority_score(
 ) -> str:
     now = date.utcnow()
     # NOTE(sileht): lower timestamps are processed first
-    # TODO(sileht): instead of * 10 looks at what we
-    # already have in the pipe and maybe process it earlier
-    if event_type == "push" and pull_number is not None:
-        return str(now.timestamp() * 10)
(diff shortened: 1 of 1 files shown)


jd:
@sileht I was reading the code and that seemed like a logical fix but I'm not sure, I don't know all the details. Let me know what you think.

mergify_engine/worker.py
```diff
@@ -150,14 +150,13 @@ def compute_priority_score(
...
     # NOTE(sileht): lower timestamps are processed first
-    # TODO(sileht): instead of * 10 looks at what we
-    # already have in the pipe and maybe process it earlier
-    if event_type == "push" and pull_number is not None:
-        return str(now.timestamp() * 10)
-    elif original_score is None:
-        return str(now.timestamp())
-    else:
-        return original_score
+    if original_score is None:
```
sileht, line 153: This is going the break the push back of push event.
I have another proposal to remove the ambiguty.

sileht: here the idea: https://github.com/Mergifyio/mergify-engine/pull/3204

PR #3203, opened 2021-09-17
feat(datadog): add service tag to statsd

sileht:
Change-Id: I154744aa37920655648c579744ad98cbebb04384


mergify_engine/web/asgi.py
```diff
@@ -22,5 +23,6 @@
 if config.SENTRY_URL:
     application = SentryAsgiMiddleware(application)  # type:ignore[assignment]
 
+statsd.constant_tags.append("service:engine-worker")
```
jd, line 26:
```suggestion
statsd.constant_tags.append("service:engine-web")
```

jd (dismissed):

jd (approved):
typo in title

PR #3204, opened 2021-09-17
chore(worker): move where we compute the score of events

sileht:
Change-Id: I5625d882ad81dc894ab784ab36b153f7df3fb794


jd:
Looks like a good idea.

jd (approved):

PR #3205, opened 2021-09-17
fix(label): don't try to remove a label already deleted

sileht:
If multiple rules delete/add labels, the action doesn't work as expected
as the pull cache is not updated.

Change-Id: I50a5d39661cbbd3beaf6b294fcd3aa81c5e12bb9

### DIFF
--- a/mergify_engine/actions/label.py
+++ b/mergify_engine/actions/label.py
@@ -57,12 +57,21 @@ class LabelAction(actions.Action):
                     json={"labels": list(missing_labels)},
                 )
                 labels_changed = True
+                labels_by_name = {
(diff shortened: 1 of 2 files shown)


mergify_engine/actions/label.py
```diff
@@ -57,12 +57,21 @@ async def run(
                     json={"labels": list(missing_labels)},
                 )
                 labels_changed = True
+                labels_by_name = {
+                    _l["name"]: _l for _l in await ctxt.repository.get_labels()
+                }
+                ctxt.pull["labels"].extend(
+                    [labels_by_name[label_name] for label_name in missing_labels]
+                )
```
jd, line 65: looks a bit hackish

jd (approved):

PR #3219, opened 2021-09-20
refactor: use a service setup function

jd:
This simplifies Datadog service setup among other things.


mergify_engine/web/asgi.py
```diff
@@ -11,18 +11,11 @@
...
-import ddtrace
 from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 
 from mergify_engine import config
-from mergify_engine import logs
 from mergify_engine.web.root import app as application
 
 
 if config.SENTRY_URL:
     application = SentryAsgiMiddleware(application)  # type:ignore[assignment]
```
sileht, line 21: I guess you forget `service.setup()` here ?

jd: yes

sileht (approved):

PR #3234, opened 2021-09-21
chore(github-actions): remove github action mode

sileht:
Change-Id: I35d48e61520f557c2d1bb78e4e06ef9bc4759c6b


jd (changes requested):
I think you forgot more `ACTION_ID` and `GITHUB_TOKEN`

jd (approved):

PR #3243, opened 2021-09-22
feat(queue/merge): unqueue pull from deleted rules

sileht:
We already cleanup redis if a queue doesn't not exists anymore.

This change unqueue pull request that has been queued by a rule that
doesn't exists anymore.

As side effect renamed rule will unqueue/requeue pull requests.

Fixes MRGFY-519

Change-Id: I6c34bcaf1298f19888d7f7c0f25161c1604c08cb


sileht:
tests incoming

mergify_engine/engine/actions_runner.py
```diff
@@ -483,6 +485,46 @@ async def run_actions(
...
+async def cleanup_pending_actions_with_no_associated_rules(
+    ctxt: context.Context,
+    match: rules.RulesEvaluator,
+    checks: typing.Dict[str, github_types.GitHubCheckRun],
+    previous_conclusions: typing.Dict[str, check_api.Conclusion],
+) -> None:
+    check_names = {
+        f"Rule: {rule.name} ({action})"
+        for rule in match.matching_rules
+        for action, action_obj in rule.actions.items()
```
jd, line 497: This looks quite fragile.

jd (approved):

PR #3252, opened 2021-09-23
fix(delayed-refresh): be always late

sileht:
Just in case the clock goes backward add a 1 minute delay.

Change-Id: I227f4179ff26b5a96fe1b47e22f224ac527bfa5e

### DIFF
--- a/mergify_engine/delayed_refresh.py
+++ b/mergify_engine/delayed_refresh.py
@@ -68,7 +68,7 @@ async def send(
     redis_stream: utils.RedisStream,
     redis_cache: utils.RedisCache,
 ) -> None:
-    score = date.utcnow().timestamp()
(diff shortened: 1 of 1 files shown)


jd (commented):
I'm a bit dubious about this. 🤔 

Why would any clock goes backward and how bad it is and how this solves the whole thing? 🤔

sileht:
> I'm a bit dubious about this. 
> 
> Why would any clock goes backward and how bad it is and how this solves the whole thing?

I didn't find why a PR refreshed at 8:00:35 doesn't get the schedule checked, while with 8:01:00 it works...
(And within unit tests, it works well at 8:00:35)

About the backward thing datetime.utcnow() rely on a non-monotonic clock so it can go backward or do jump in the future.


I agree, this workaround is ugly, I still looking for the root cause.

PR #3254, opened 2021-09-23
fix(branch-protection): handle case of private repository with no GitHub subscription

sileht:
Change-Id: I6153a48a7db9d2e0d48abf6a9ea52128205b4876


mergify_engine/context.py
```diff
@@ -552,28 +571,19 @@ async def get_branch_protection(
             except http.HTTPNotFound:
                 branch_protections[branch_name] = None
             except http.HTTPForbidden as e:
+                if e.status_code != 403:
```
jd, line 574: Can you have `HTTPForbidden` with a status_code not 403?

jd (approved):

PR #3271, opened 2021-09-27
fix(branch-protection): remove support for code-owners

sileht:
`#review-requested=0` is not a perfect match for CODEOWNERS.

Users are more annoyed by the strict meaning of our version (pull
request not queued, because non CODEOWNERS review are requested) than
seeing the benefice of the issue we are trying to solve.

Change-Id: Ib9e9a61586d4d7683f7a7fe4f23722a274ae1fc9

### DIFF
--- a/mergify_engine/rules/conditions.py
+++ b/mergify_engine/rules/conditions.py
@@ -447,15 +447,6 @@ async def get_branch_protection_conditions(
                     ),
                 ]
             )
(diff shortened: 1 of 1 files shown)


jd:
Commit message?

sileht:
> Commit message?

I was waiting for the number of complaints we have about this stricter version CODEOWNERS.

jd (approved):

PR #3285, opened 2021-09-29
fix(branch-protection): rework linear-history

### LINKED ISSUE #3284: PRs with `linear-history` status not checked do not get rebased anymore and do not show up in merge queue
We enforce a linear history in our repositories. When all checks were green for a PR, but the branch needed a rebase, then Mergify would merge the main branch into the PR and eventually merge the PR. This doesn't seem to happen anymore. 

Multiple of our PRs show `linear-history` as unchecked (which seems to be a new feature?) and mergify does not update the PR anymore. The PR also does not show up in the merge queue and is essentially stuck in that state - until someone rebases it manually (which, without the merge queue, is potentially racing against other PRs).

## Specifications

  - Pull Request URL: https://github.com/mozilla-mobile/fenix/pull/21446
  - Mergify Config URL: https://github.com/mozilla-mobile/fenix/blob/main/.mergify.yml

sileht:
The linear-history of Mergify has not the same meaning as the
linear-history of branch protection.

For branch protection, the condition is enforced on the commit tree
entering in the base branch (after the merge button got pressed or when
people `git push` something directly in the base branch), meaning for us
only the rebase and squash merge methods are allowed.

But in Mergify, we talk about the commit tree before it get merged.

This change reworks the merge/queue action to report configuration
mismatch between Mergify and GitHub branch protection instead of
injecting the `linear-history` condition.

Fixes #3284
Fixes MRGFY-600

Change-Id: If94138383f13f09d0b25849a3081948535d4b387


mergify_engine/actions/merge_base.py
```diff
@@ -698,9 +698,17 @@ async def merge_report(
...
+            title = "Branch protection setting 'strict' conflicts with Mergify configuration."
             summary = ""
 
+        elif (
+            await self._is_branch_protection_linear_history_enabled(ctxt)
+            and self.config["method"] == "merge"
+        ):
+            conclusion = check_api.Conclusion.FAILURE
+            title = "Branch protection setting 'linear history' conflicts with Mergify configuration."
+            summary = "Branch protection setting 'linear history' works only if `method=squash` or `method=rebase`."
```
jd, line 710:
```suggestion
            summary = "Branch protection setting 'linear history' works only if `method: squash` or `method: rebase`."
```
At least use the same syntax than the YAML.

Do we need a test for this?

jd (dismissed):

jd (approved):

PR #3298, opened 2021-09-30
docs(conflict): write down it doesn't work on draft

sileht:
Related MRGFY-605

Change-Id: Ibe4de612c18e4f652e9b43bc2070bac4d0618959


jd (changes requested):
Fixes MRGFY-605

jd (approved):

PR #3299, opened 2021-09-30
feat: add VERSION to the built docker image

sileht:
This change adds engine version to the docker image and report it to the
dashboard.

Fixes #MRGFY-557

Change-Id: I4af8fde046cd9b8003d10a6dd200cd3a196e4f34


jd (approved):

jd:
Might be great to use this config everywhere where we used Heroku vars?

PR #3330, opened 2021-10-06
feat(commits-behind): handle pull request with base branch merged in

sileht:
This also looks are merge commit from the base branch like is_behind
does.

Fixes MRGFY-596

Change-Id: I0e0bea9c1ac611e957230476c97ed6db86c44804


jd:
fix body:
> This also looks at merge commit from the base branch, like is_behind does.

jd (approved):

PR #3447, opened 2021-10-21
chore: Deprecation commit_message option

Zerdanes:
We should inform users about the option removal by adding;
* a deprecation warning in the documentation
* a release note entry about the deprecation
* a deprecation warning in the summary if the configuration option is set

Fixes MRGFY-642


sileht (dismissed):

docs/source/actions/merge.rst
```diff
@@ -171,6 +172,19 @@ request:
...
+
+   `commit_message` will be deprecated in April 25th, 2022. The :ref:`data type template`
+   will provide a more powerful solution
+   and offer total control of the merge commit.
+
+
+   To migrate from `"commit_message: title+body"` to the ``commit_message_template`` mode,
+   you just need to change in the :ref:`configuration file format` ``.mergify.yml`` in the action part,
+   and use ``commit_message_template``.
+
```
sileht, line 187:
```suggestion
   ``commit_message`` is deprecated and will be removed on April 25th, 2022. It's replaced by :ref:`data type commit_message_template`.


   To migrate from ``commit_message: title+body`` to the :ref:`data type commit_message_template` attribute, set it to:
   
   \```
   {{ title }} (#{{ number }})
   
   {{ body }}
    \```

```

sileht (changes requested):
I rephrase the doc, but @jd may have a better version.

docs/source/actions/merge.rst
```diff
@@ -171,6 +172,19 @@ request:
 Defining the Commit Message
 ---------------------------
 
+|deprecated tag|
+
+.. warning::
+
+   `commit_message` will be deprecated in April 25th, 2022. The :ref:`data type template`
+   will provide a more powerful solution
+   and offer total control of the merge commit.
```
jd, line 181: Well if you mark it deprecated today, it's not going to be deprecated on 25th April.
I guess you mean "removed" here, not deprecated.

mergify_engine/engine/actions_runner.py
```diff
@@ -57,6 +57,44 @@
 
 """
 
+COMMIT_MESSAGE_MODE_MERGE_DEPRECATION_GHES = """
+:bangbang: **Action Required** :bangbang:
+
+> **The configuration uses the deprecated `commit_message` mode of the merge action.**
+> This option will be removed on version 3.0.0.
```
jd, line 64: Not sure we want to set a version number here. Between now and April we might be at version 4.0.0…

Just write "on a future version".

releasenotes/notes/commit_message-5ffab4c289471654.yaml
```diff
@@ -0,0 +1,5 @@
+---
+deprecations:
+  - |
+    The `commit_message` configuration option of the merge action and queue action is deprecated and
+    planned for removal for 3.0.0.
```
jd, line 5: ditto

jd (changes requested):

docs/source/actions/merge.rst
```diff
@@ -171,6 +173,20 @@ request:
...
+    ``commit_message`` is deprecated and will be removed on April 25th, 2022. It's replaced by :ref:`commit_message_template <commit_message_template>`.
+    To migrate from ``commit_message: title+body`` to the :ref:`commit_message_template <commit_message_template>` attribute, set it to:
+
+
+   \```
+   {{ title }} (#{{ number }})
+   
+   {{ body }}
+    \```
+
```
jd, line 189: quite ugly https://docs-preview.mergify.com/3447/docs/docs/actions/merge/#defining-the-commit-message 😢

Zerdanes: will it be better if i put it in a code-block ?

jd: try it?

mergify_engine/engine/actions_runner.py
```diff
@@ -57,6 +57,44 @@
 
 """
 
+COMMIT_MESSAGE_MODE_MERGE_DEPRECATION_GHES = """
+:bangbang: **Action Required** :bangbang:
+
+> **The configuration uses the deprecated `commit_message` mode of the merge action.**
+> This option will be removed on a future version.
+> For more information: :ref: `doc.mergify.io/actions/merge/`
```
jd, line 65: This is markdown text posted in a PR, not RST documentation. You need a proper link formatted in Markdown.

docs/source/actions/merge.rst
```diff
@@ -171,6 +173,23 @@ request:
...
+   .. code-block:: md
+   
+    ## Commit Message Template
+   
+    ```
+    {{ title }} (#{{ number }})
+   
+    {{ body }}
+    ```
```
jd, line 191: That sounds wrong?
Should be

```suggestion
    {{ title }} (#{{ number }})
   
    {{ body }}
```

IIUC

jd (changes requested):

docs/source/actions/merge.rst
```diff
@@ -171,6 +173,21 @@ request:
...
+|deprecated tag|
+
+.. warning::
+
+    ``commit_message`` is deprecated and will be removed on April 25th, 2022. It's replaced by :ref:`commit_message_template <commit_message_template>`.
+    To migrate from ``commit_message: title+body`` to the :ref:`commit_message_template <commit_message_template>` attribute, set it to:
+
+   .. code-block:: md
+   
+    ## Commit Message Template
```
jd, line 185: why this here? 🤔

Zerdanes: To put the code block in the warning i indent it, i think it's prettier like that, no ?

jd: I'm not talking about the indentation, but the title here. Why would you need a title here?
Why would you set a markdown title in `commit_message_template`?

Zerdanes: I wanted to do like the ## Commit Message, I thought it was like a comment with the #, so that the user understands which part it corresponds to. It is not used for that ?

jd: `#` is used in some language for commenting, but I don't know any language using `##` as a comment delimiter. 

As you might know, `commit_message_template` is a Jinja2 template that generates text for a commit message.
Here's how to put comments in Jinja2: https://jinja.palletsprojects.com/en/3.0.x/templates/#comments

So in the end, `## Commit Message Template` is going to be the **title** of the commit with your example. I doubt anyone would want that 😅

Zerdanes: Okay, i thought it was a specific syntax for documentation, i'll put that in {%%}

Zerdanes: in fact i'm just going to take it off, as the user will not see it

jd (dismissed):

docs/source/actions/merge.rst
```diff
@@ -181,8 +196,6 @@ the pull request body that starts with ``Commit Message``.
 
 .. code-block:: md
 
-    ## Commit Message
```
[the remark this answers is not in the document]
Zerdanes: mb I confused with the other one that I removed, I thought it was the right one when I reread it

mergify_engine/engine/actions_runner.py
```diff
@@ -151,6 +170,18 @@ def _has_merge_action_with_strict_mode(
...
 
+def _has_merge_action_with_commit_message_mode(
+    pull_request_rules: rules.PullRequestRules,
+) -> bool:
+
+    return any(
+        action
+        for rule in pull_request_rules
+        for name, action in rule.actions.items()
+        if name == "merge" and ("commit_message" in action.raw_config)
```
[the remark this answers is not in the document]
jd: This seems like you could do simpler with just:
```suggestion
        if name in ("merge", "queue") and ("commit_message" in action.raw_config)
```
without duplicating thew hole function.

jd (changes requested):

mergify_engine/engine/actions_runner.py
```diff
@@ -151,6 +189,18 @@ def _has_merge_action_with_strict_mode(
...
+def _has_merge_action_with_commit_message_mode(
+    pull_request_rules: rules.PullRequestRules,
+) -> str:
+    nameValue = ""
+    for rule in pull_request_rules:
+        for name, action in rule.actions.items():
+            if name in ("merge", "queue") and ("commit_message" in action.raw_config):
+                nameValue = name
+                return nameValue
+    return ""
```
jd, line 201: Why are you returning an `str` here?
You could just return a `bool` that would be used in the code below.

jd (changes requested):

mergify_engine/engine/actions_runner.py
Zerdanes: Because i wanted to do a verification when i call this function, to know if the action is merge or queue and add to the summary the right message. Or maybe i can use the same message of deprecation for both, but i think it can be confusing :/

mergify_engine/engine/actions_runner.py
```diff
@@ -151,6 +189,18 @@ def _has_merge_action_with_strict_mode(
     )
 
 
+def _has_merge_action_with_commit_message_mode(
+    pull_request_rules: rules.PullRequestRules,
+) -> str:
+    nameValue = ""
```
jd, line 195: This variable is actually useless.

mergify_engine/engine/actions_runner.py
jd: Alright, that makes sense. If you want to include the name of the action though, it's only in one place in your message so it'd be easy to use something like `{action_name}` in your deprecation message and replace it on the fly.

In the case of this function, you should not remove an empty string — it's always a bad design / idea.  Just return `None` to signal that you didn't find a result.

jd (changes requested):

mergify_engine/engine/actions_runner.py
```diff
@@ -151,6 +170,19 @@ def _has_merge_action_with_strict_mode(
...
 
 
+def _has_merge_action_with_commit_message_mode(
+    pull_request_rules: rules.PullRequestRules,
+) -> bool:
+
+    for rule in pull_request_rules:
+        for name, action in rule.actions.items():
+            if name in ("merge", "queue") and ("commit_message" in action.raw_config):
+                COMMIT_MESSAGE_MODE_DEPRECATION_GHES.format(action_name=name)
```
jd, line 180: This does nothing since you're not using the returned value. 🤷

jd (changes requested):

mergify_engine/engine/actions_runner.py
```diff
@@ -164,6 +197,8 @@ async def gen_summary(
             summary += STRICT_MODE_DEPRECATION_SASS
         else:
             summary += STRICT_MODE_DEPRECATION_GHES
+    if _has_merge_action_with_commit_message_mode(pull_request_rules) is not None:
+        summary += _has_merge_action_with_commit_message_mode(pull_request_rules)
```
jd, line 201: You're calling `_has_merge_action_with_commit_message_mode`.

Imagine the function takes 2 hours… now it takes 4 hours to run this code 😢

jd (changes requested):

jd (commented):
@sileht I realize there is no test here. Do we want one?

mergify_engine/engine/actions_runner.py
```diff
@@ -164,7 +197,9 @@ async def gen_summary(
             summary += STRICT_MODE_DEPRECATION_SASS
         else:
             summary += STRICT_MODE_DEPRECATION_GHES
-
+    summaryValue = _has_merge_action_with_commit_message_mode(pull_request_rules)
```
sileht, line 200: We use snake cases everywhere
```suggestion
    summary_value = _has_merge_action_with_commit_message_mode(pull_request_rules)
```

sileht:
> @sileht I realize there is no test here. Do we want one?

We have tests that use commit_message, we should just add in one of them an assertion to check if the message appears.

sileht (commented):
Looks good for me.

mergify_engine/engine/actions_runner.py
```diff
@@ -164,7 +197,9 @@ async def gen_summary(
             summary += STRICT_MODE_DEPRECATION_SASS
         else:
             summary += STRICT_MODE_DEPRECATION_GHES
-
+    summary_Value = _has_merge_action_with_commit_message_mode(pull_request_rules)
+    if summary_Value is not None:
+        summary += summary_Value
```
jd, line 202: 1. the var name should be `summary_value` see https://en.wikipedia.org/wiki/Snake_case
2. the variable name is not really useful here. I'd rather name it `summary_commit_message_deprecation` for example

jd (approved):
Just a comment on the var name but not a blocker.

mergify_engine/engine/actions_runner.py
Zerdanes: Camel case don't want to let me go :smiling_face_with_tear:

PR #3498, opened 2021-10-29
feat(template): add new jinja filters

sileht:
* markdownify
* get_section("### Commit message")

Fixes MRGFY-635

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -318,11 +318,36 @@ will render to:
 when used in your configuration file — considering the pull request author
 login is ``jd``.
 
+`Jinja2 filters <https://jinja.palletsprojects.com/en/3.0.x/templates/#builtin-filters>` are supported, you can build string from list for example with:
+
+.. code-block:: jinja
+
+    Approved by: @{{ approved-reviews-by | join(', @') }}
+
+We also provide custom Jinja2 filters:
+
+* ``markdownify``: to convert HTML to Markdown:
+
+.. code-block:: jinja
+
+    {{ body | markdownify }}
+
+* ``get_section(<section>)``: to extract one Markdown section
+
+.. code-block:: jinja
+
+    {{ body | get_section("## Description") }}
+
 .. note::
 
    You need to replace the ``-`` character by ``_`` from the :ref:`pull request
    attribute <attributes>` names when using templates. The ``-`` is not a valid
    character for variable names in Jinja2 template.
+
+.. note::
+
+   By default, the HTML comments are stripped from ``body``. To get the
+   full body, you can use the ``body_raw`` attribute.
 
 Validation
 ----------
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -33,6 +33,7 @@ import jinja2.meta
 import jinja2.runtime
 import jinja2.sandbox
 import jinja2.utils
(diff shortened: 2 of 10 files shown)


docs/source/configuration.rst
```diff
@@ -318,12 +318,37 @@ will render to:
...
 .. note::
 
    You need to replace the ``-`` character by ``_`` from the :ref:`pull request
    attribute <attributes>` names when using templates. The ``-`` is not a valid
    character for variable names in Jinja2 template.
 
+.. note::
+
+   By default, the html comment is stripped from ``body``, to get the
+   full body, you should use ``body-raw``.
```
jd, line 350:
```suggestion
   By default, the HTML comments are stripped from ``body``. To get the
   full body, you can use the ``body_raw`` attribute.
```

mergify_engine/context.py
```diff
@@ -1601,7 +1603,13 @@ class PullRequest(BasePullRequest):
     }
 
     async def __getattr__(self, name: str) -> ContextAttributeType:
-        return await self.context._get_consolidated_data(name.replace("_", "-"))
+        data_name = name
+        if name == "body-raw":
+            data_name = "body"
+        data = await self.context._get_consolidated_data(data_name.replace("_", "-"))
+        if name == "body":
+            return re.sub("(<!--.*?-->)", "", typing.cast(str, data))
```
jd, line 1611: I'd do it from `_get_consolidate_data`, no? sounds cleaner

mergify_engine/context.py
```diff
@@ -1633,6 +1642,33 @@ async def render_template(
...
+        env.filters["get_section"] = self._filter_get_section
+
+    @staticmethod
+    def _filter_get_section(v: str, section: str) -> str:
+        if not isinstance(section, str):
+            raise TypeError("level must be a string")
+
+        section_escaped = re.escape(section)
+        level = MARKDOWN_TITLE_RE.match(section)
+        print(section_escaped)
```
jd, line 1656:
```suggestion
```

mergify_engine/context.py
```diff
@@ -1633,6 +1642,33 @@ async def render_template(
...
+
+        level_str = level[0].strip()
+
+        level_re = re.compile(fr"^{level_str} +", re.I)
+        section_re = re.compile(fr"^{section_escaped}\s*$", re.I)
+
+        lines = v.split("\n")
+        lines = list(itertools.dropwhile(lambda l: not section_re.match(l), lines))
+        if not lines:
+            raise TypeError("section not found")
```
jd, line 1668:
```suggestion
        lines = itertools.dropwhile(lambda l: not section_re.match(l), lines)
        if not any(lines):
            raise TypeError("section not found")
```

jd (changes requested):

mergify_engine/context.py
sileht: `_get_consolidate_data()` is also used with the `rules.filter` code. If we change the body there, rules trying to match something in a comment will fail. But maybe we don't care.

jd: I don't think we care indeed.
We might want to update the doc to include both `body` and `body-raw` though.

jd (changes requested):
release note?

jd (approved):

PR #3506, opened 2021-11-02
feat(api-docs): publish api reference in /api

sileht:
Fixes MRGFY-676


tox.ini
```diff
@@ -115,6 +116,7 @@ commands =
   bash -c "cd docs && yarn && yarn build-bootstrap"
   sphinx-build -W -b spelling docs/source docs/build
   sphinx-build -W -b dirhtml docs/source docs/build
+  cp -a docs/api docs/build/api
```
jd, line 119: nit but can we just generate in the right directory directly?

PR #3516, opened 2021-11-03
chore(edit): move edit_draft_state() in the EditAction class

sileht:


jd (approved):
`refactor` would be a better title :)

PR #3519, opened 2021-11-03
chore: update URL to mergify.com

jd:
Fixes MRGFY-674


jd:
@sileht recording works for this test but replay does not

```
mergify_engine/tests/functional/actions/test_queue.py::TestQueueAction::test_queue_priority PASSED [100%]
```

sileht:
> @sileht recording works for this test but replay does not
> 
> ```
> mergify_engine/tests/functional/actions/test_queue.py::TestQueueAction::test_queue_priority PASSED [100%]
> ```

https://github.com/Mergifyio/mergify-engine/pull/3538 should fix it.

sileht (approved):

PR #3539, opened 2021-11-04
fix(docs): rename api.rst in api-intro.rst

sileht:
This removes conflict between api.rst and the openapi website.

Fixes MRGFY-687


jd (changes requested):
rename failed

jd (approved):

PR #3540, opened 2021-11-04
feat: Add a global configuration option to simulate the removal of the option commit_message

Zerdanes:
Add a new configuration option `ALLOW_COMMIT_MESSAGE` to disable the commit_message option.

Fixes MRGFY-644


sileht (changes requested):
Looks good, just need to add a release note like https://github.com/Mergifyio/mergify-engine/blob/main/releasenotes/notes/strict-mode-149a1ae514e83edd.yaml but without mentioning the version.

sileht (changes requested):
The same should be done in queue action.

sileht (changes requested):
You missed my last comment.

mergify_engine/actions/queue.py
```diff
@@ -57,6 +58,37 @@ def get_config_schema(
...
+                    "default", "title+body"
+                ),
+                voluptuous.Required(
+                    "commit_message_template", default=None
+                ): types.Jinja2WithNone,
+                voluptuous.Required(
+                    "priority", default=queue.PriorityAliases.medium.value
+                ): queue.PrioritySchema,
+                voluptuous.Required("require_branch_protection", default=True): bool,
+            }
```
Zerdanes, line 91: I think i can optimize this part to avoid having duplicate voluptuous, and create validator object as the same as merge file.

sileht: You should! 😉

mergify_engine/config.py
```diff
@@ -280,6 +281,7 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
 TESTING_MERGIFY_TEST_1_ID: int
 TESTING_MERGIFY_TEST_2_ID: int
 ALLOW_MERGE_STRICT_MODE: bool
+ALLOW_COMMIT_MESSAGE_MODE: bool
```
sileht, line 284:
```suggestion
ALLOW_COMMIT_MESSAGE_OPTION: bool
```

sileht (changes requested):

mergify_engine/actions/queue.py
```diff
@@ -88,6 +87,17 @@ def get_config_schema(
             voluptuous.Required("require_branch_protection", default=True): bool,
         }
 
+        validator_commit_message = {
+            voluptuous.Required("commit_message", default="default"): voluptuous.Any(
+                "default", "title+body"
+            ),
+        }
+
+        schema = validator_default.copy()
```
sileht, line 96: Unlike the merge action, validator_default is not a class variable here, it's local to the method, so why do you copy it?

sileht (changes requested):

sileht (approved):

PR #3574, opened 2021-11-09
fix(config): fix detection of configuration change

sileht:
Fixes MRGFY-704


jd (approved):
I wish you wrote a nice commit message that explains the bug 😢

PR #3584, opened 2021-11-10
fix(conflict): change way to detect them

sileht:
There is a tiny drawback to the new method, we need to wait that
mergeable attribute becomes not None, GitHub seems slower to compute it
compared to mergeable_state.

Fixes MRGFY-703


mergify_engine/tests/functional/test_engine.py
```diff
@@ -1220,16 +1220,41 @@ async def test_refresh_on_conflict(self):
         p1, _ = await self.create_pr(files={"TESTING": "p1"})
         p2, _ = await self.create_pr(files={"TESTING": "p2"})
         await self.merge_pull(p1["number"])
+        await self.wait_for("push", {"ref": f"refs/heads/{self.main_branch_name}"})
+
+        # Wait a bit than Github refresh the mergeable_state before running the
+        # engine
```
jd, line 1226:
```suggestion
        # Wait a bit that Github refreshes the mergeable_state before running the
        # engine
```

mergify_engine/tests/functional/test_engine.py
```diff
@@ -1220,16 +1220,41 @@ async def test_refresh_on_conflict(self):
...
+            ]
+        }
+        await self.setup_repo(yaml.dump(rules), files={"TESTING": "foobar"})
+        p1, _ = await self.create_pr(files={"TESTING": "p1"})
+        p2, _ = await self.create_pr(files={"TESTING": "p2"}, draft=True)
+        await self.merge_pull(p1["number"])
+        await self.wait_for("push", {"ref": f"refs/heads/{self.main_branch_name}"})
 
         # Wait a bit than Github refresh the mergeable_state before running the
         # engine
```
jd, line 1253: ditto

jd (approved):

PR #3601, opened 2021-11-15
fix(post_check): filter useless signaling

sileht:


jd (commented):
more details on the why?

jd (approved):

sileht:
> more details on the why?

you're too fast!

PR #3605, opened 2021-11-16
fix(queue): use hash to store trains

sileht:
Instead of having one key per trains, this change uses a hash per
organization and then one key per repo_id/ref.

Since we use iter_trains() more often than before this should load train
faster that before.

Also compared to SCAN, HSCAN also load values, so we also remove one
round-trip to retrieve the data.

Fixes MRGFY-714

### DIFF
--- a/mergify_engine/queue/merge_train.py
+++ b/mergify_engine/queue/merge_train.py
@@ -916,20 +916,11 @@ class Train(queue.QueueBase):
         await q.load()
         return q
 
-    @staticmethod
-    def get_redis_key_for(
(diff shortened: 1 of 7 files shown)


sileht:
TODO:
[ ] write basic unit test for the migration script. 
[X] get a redis dump to measure the downtime migration
[ ] test the migration on the dump

jd (changes requested):
Release note? We need to handle upgrades and probably us a major version bump when releasing.

sileht:
I test the migration on a database dump and it was instant locally.

jd (approved):

PR #3610, opened 2021-11-17
feat(worker): add dedicated worker

sileht:
This changes move organization to a dedicated worker task if they have
the feature "dedicated-workers".

Both path are tested during fonctionnal tests as all tests that require
subscription will automatically spawn a dedicated worker.

Fixes MRGFY-581

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -99,6 +99,8 @@ MAX_RETRIES: int = 3
 WORKER_PROCESSING_DELAY: float = 30
 STREAM_ATTEMPTS_LOGGING_THRESHOLD: int = 20
 
+DEDICATED_WORKERS_KEY = "dedicated-workers"
+
 OrgBucketNameType = typing.NewType("OrgBucketNameType", str)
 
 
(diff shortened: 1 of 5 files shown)


mergify_engine/worker.py
```diff
@@ -235,18 +237,53 @@ async def run_engine(
...
+        return owner_id == self.owner_id
+
+    async def next_org_bucket(self) -> typing.Optional[OrgBucketNameType]:
+        # NOTE(sileht): Should we drop the 30s latency ? That may impact the rate limit...
+        now = time.time()
+        for org_bucket in await self.redis_stream.zrangebyscore(
+            "streams",
+            min=0,
+            max=now,
+        ):
```
jd, line 261: That looks suboptimal.
What about using directly `ZSCORE streams org_bucket` and checking the returned value?

jd (dismissed):
I'm wondering if we should have a little more abstraction, like a class taking care of the set of dedicated worker and starting/shutting them down. The methods start to be a bit long and it makes it harder to do unit test around them.

sileht:
> I'm wondering if we should have a little more abstraction, like a class taking care of the set of dedicated worker and starting/shutting them down. The methods start to be a bit long and it makes it harder to do unit test around them.

The management is only located in `dedicated_workers_spawner_task()`

mergify_engine/worker.py
jd: Ok so maybe just add a note at least here 😢

mergify_engine/worker.py
```diff
@@ -235,18 +237,55 @@ async def run_engine(
...
+            if self._is_org_bucket_for_me(org_bucket):
+                statsd.increment(
+                    "engine.streams.selected",
+                    tags=[f"worker_id:dedicated-{self.owner_id}"],
```
jd, line 267:
```suggestion
                    tags=[f"worker_id:dedicated-{self.owner_id}", "dedicated"],
```
otherwise it's going to be hard to select them in DD.

jd (dismissed):

mergify_engine/worker.py
```diff
@@ -986,9 +1078,63 @@ async def start(self) -> None:
...
+                    self._dedicated_worker_tasks[owner_id] = asyncio.create_task(
+                        self.dedicated_stream_worker_task(owner_id)
+                    )
+
+            except yaaredis.ConnectionError:
+                statsd.increment("redis.client.connection.errors")
+                LOG.warning(
+                    "dedicated_workers_spawner task lost Redis connection",
+                    exc_info=True,
+                )
```
jd, line 1125: This block seems quite large for what it can catch :/

jd (approved):

PR #3626, opened 2021-11-19
feat(admin): add tool to suspend installation

sileht:
Fixes MRGFY-727

### DIFF
--- a/mergify_engine/tests/unit/test_admin.py
+++ b/mergify_engine/tests/unit/test_admin.py
@@ -1,0 +1,45 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2021 Mergify SAS
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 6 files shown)


jd (changes requested):
We need at least a simple unit test.
Also, it'd be good to have unsuspend.

(Maybe with that it could even be functionally tested maybe?)

jd (dismissed):
still lacking a unit test launching the CLI

mergify_engine/tests/unit/test_admin.py
```diff
@@ -0,0 +1,46 @@
...
+import shutil
+import subprocess
+import sys
+from unittest import mock
+
+from mergify_engine import admin
+
+
+def test_admin_tool_installation() -> None:
+    shutil.which("mergify-admin")
```
jd, line 25: is this useful?

jd (changes requested):
useless command I think

jd (approved):

PR #3631, opened 2021-11-22
fix(queue): update head sha after pull request has been updated

sileht:
When the `update_method` is `rebase`, the first queued pull request (P1)
is rebased and the second one (P2) creates a temporary pull request to
test P1+P2 together.

Because we rebase P1, the old head sha of P1 may has vanished on GitHub
before we create the temporary pull request.

We don't really care to use the old or the new head sha. Even our functional
tests was accepting both.

This changes the way we got the next head sha for the first pull
request. This ensures we already use the new one when we create
subsequent temporary pull request.

Fixes MRGFY-723


mergify_engine/queue/merge_train.py
```diff
@@ -340,6 +340,10 @@ async def update_user_pull(self, queue_rule: rules.QueueRule) -> None:
             await self._set_creation_failure(f"{exc.title}\n\n{exc.message}", "update")
             raise TrainCarPullRequestCreationFailure(self) from exc
 
+        # NOTE(sileht): We must update head_sha of the pull request otherwise next temporary pull request
+        # may be created on a vanish reference.
```
jd, line 344:
```suggestion
        # NOTE(sileht): We must update head_sha of the pull request; otherwise next temporary pull request
        # may be created on a vanished reference.
```

jd (changes requested):
The commit message is not really clear and has typo 😢

sileht:
> The commit message is not really clear and has typo cry

I rephrase

jd (approved):

PR #3632, opened 2021-11-22
fix(config): do not require testing gpg keys

jd:


sileht (approved):

jd:
need to clean out the env once merged

PR #3638, opened 2021-11-23
fix(worker): stop workers correctly

sileht:
This makes workers unit tests failing randomly a lot.


jd (approved):

jd:
Is this linked to https://linear.app/mergify/issue/MRGFY-728/test-failure-due-to-task-was-destroyed-but-it-is-pending-error ?

PR #3640, opened 2021-11-23
fix(installer): force-upgrade vulnerable JS dependencies

jd:


installer/package.json
```diff
@@ -69,5 +69,13 @@
     "eslint-plugin-jsx-a11y": "^6.5.1",
     "eslint-plugin-react": "^7.27.1",
     "eslint-plugin-react-hooks": "^4.3.0"
+  },
+  "resolutions": {
```
sileht, line 73: This can inject nasty bug, since we will install this version instead of respecting packages dependencies...

jd: This is why we have tests, no? 😄

sileht:
react-scripts will never update its deps (eg: facebook/create-react-app#11174 and https://overreacted.io/npm-audit-broken-by-design/)

And we will need to update this list again and again.

Here, another approach to the issue: https://github.com/Mergifyio/mergify-engine/pull/3659

PR #3643, opened 2021-11-23
chore(rules): add more get_rule_checks_status() tests

sileht:
Change-Id: I70395cdc19c60e1f7ace9136d261ee0e089cfc4d

PR #3644, opened 2021-11-23
fix(queue/rules): handle correctly when we use negative check conditions

sileht:
If pending/failure checks have a negative conditions, the current code
doesn't works, as we explicit replace any `check-XXX` by `check` to
known if all required checks has reported they values.

This change remove the "-" from the conditions and look that required
checks are not pending.

This will to write rules like:

- check-success=seatbelt
- -check-pending=foo
- -check-failure=foo

Fixes MRGFY-730

Change-Id: I1ce946161bc1acafbc23fbfa00fb88e5baf30875

Depends-On: #3643

### DIFF
--- a/mergify_engine/actions/merge_base.py
+++ b/mergify_engine/actions/merge_base.py
@@ -83,30 +83,38 @@ async def get_rule_checks_status(
 ) -> check_api.Conclusion:
 
     if rule.conditions.match:
+        log.debug("everything match")
         return check_api.Conclusion.SUCCESS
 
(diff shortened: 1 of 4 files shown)


sileht:
1. chore(rules): add more get_rule_checks_status() tests ([#3643](https://github.com/Mergifyio/mergify-engine/pull/3643))
1. fix(queue/rules): handle correctly when we use negative check conditions ([#3644](https://github.com/Mergifyio/mergify-engine/pull/3644))

PR #3663, opened 2021-11-25
feat(api): add badges endpoint

sileht:
Fixes MRGFY-688

Change-Id: Ief6665d27f161f701084ade646684d6dc4c87b40


mergify_engine/web/api/queues.py
```diff
@@ -36,6 +36,13 @@
 
 LOG = daiquiri.getLogger(__name__)
 
+router = fastapi.APIRouter(
+    tags=["queues"],
+    dependencies=[
+        fastapi.Depends(security.require_authentication),
+    ],
+)
+
```
jd, line 45: unrelated?

jd (approved):

mergify_engine/web/api/queues.py
sileht: It's related, I relax the default security dependencies of /v1 endpoint, I move them to queues/applications routers instead. So /v1/badges don't need authentication and this is reflexed in OpenAPI generated file.

PR #3666, opened 2021-11-25
feat(template): allow templating in get_section()

sileht:
Interpret the string retrieved with get_section() has template.

Change-Id: I0db496d01fc2a36bd9ab9ddbb73f3a04fe0f6553

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -18,6 +18,7 @@ import base64
 import contextlib
 import dataclasses
 import datetime
+import functools
 import itertools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


mergify_engine/context.py
```diff
@@ -1690,7 +1693,10 @@ def _filter_get_section(v: str, section: str) -> str:
         if not found:
             raise TypeError("section not found")
 
-        return ("\n".join(section_lines)).strip()
+        # We don't allow get_section to ensure we can't have ending recursion
```
jd, line 1696: you mean never-ending?

jd (approved):

PR #3671, opened 2021-11-26
fix(application): use new token validation endpoint

sileht:
Fixes MRGFY-733

Change-Id: If3ae6475ae30d4ff46b1d5e7585044cf4fe547b4


sileht:
Waiting for the other side to get merged first.

jd (approved):

PR #3678, opened 2021-11-29
fix(queue/rules): use a Filter to found pending/final checks

sileht:
Instead of the heuristic that looks only for `check-success=XXX`.

This creates a Filter that introduces a new possible result
`IncompleteCheck`.

The filter has a different meaning for or/and when boolean are combined
with IncompleteCheck:
* or cannot return false if at least one is IncompleteCheck
* and cannot return false if at least one is IncompleteCheck

To compute if a check conditions is final or not, this change adds an
hook in Filter before the evaluation of the condition to return IncompleteCheck
if need instead of the real evaluation outcome.

Fixes MRGFY-729
Fixes MRGFY-730
Fixes MRGFY-731
Fixes MRGFY-736

Change-Id: I1ce946161bc1acafbc23fbfa00fb88e5baf30875

### DIFF
--- a/mergify_engine/rules/filter.py
+++ b/mergify_engine/rules/filter.py
@@ -16,6 +16,7 @@
 from collections import abc
 import dataclasses
 import datetime
+import enum
 import inspect
 import operator
 import re
@@ -233,6 +234,20 @@ class Filter(typing.Generic[FilterResultT]):
             raise InvalidArguments(nodes)
         return self._handle_multiple_op(multiple_op, nodes)
 
+    def _eval_binary_op(
+        self,
+        op: BinaryOperatorT[FilterResultT],
+        attribute_name: str,
+        attribute_values: typing.List[typing.Any],
+        ref_values_expanded: typing.List[typing.Any],
+    ) -> FilterResultT:
+        binary_op, iterable_op, compile_fn = op
+        return iterable_op(
+            binary_op(attribute_value, ref_value)
+            for attribute_value in attribute_values
+            for ref_value in ref_values_expanded
+        )
+
     def _handle_binary_op(
         self,
         op: BinaryOperatorT[FilterResultT],
@@ -258,10 +273,8 @@ class Filter(typing.Generic[FilterResultT]):
                     typing.Awaitable[typing.Any], ref_values_expanded
                 )
 
-            return iterable_op(
-                binary_op(attribute_value, ref_value)
-                for attribute_value in attribute_values
-                for ref_value in ref_values_expanded
+            return self._eval_binary_op(
+                op, attribute_name, attribute_values, ref_values_expanded
             )
 
         return _op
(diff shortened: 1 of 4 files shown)


mergify_engine/rules/filter.py
```diff
@@ -491,3 +511,174 @@ def NearDatetimeFilter(
...
+        return IncompleteCheck
+    return True
+
+
+def IncompleteChecksAny(
+    values: typing.Iterable[object],
+) -> IncompleteChecksResult:
+    values = typing.cast(typing.Iterable[IncompleteChecksResult], values)
+    if not values:
+        return True
```
jd, line 553: Not sure about this one:

```python
>>> any([])
False
```

mergify_engine/rules/filter.py
```diff
@@ -113,6 +114,18 @@ class Filter(typing.Generic[FilterResultT]):
     binary_operators: typing.Dict[str, BinaryOperatorT[FilterResultT]]
     multiple_operators: typing.Dict[str, MultipleOperatorT[FilterResultT]]
 
+    binary_op_hook: typing.Optional[
```
jd, line 117: add some documentation on this mechanism

mergify_engine/rules/filter.py
```diff
@@ -258,6 +271,13 @@ async def _op(obj: GetAttrObjectT) -> FilterResultT:
                     typing.Awaitable[typing.Any], ref_values_expanded
                 )
 
+            if self.binary_op_hook is not None:
+                res = self.binary_op_hook(
+                    binary_op, iterable_op, ref_values_expanded, attribute_name
+                )
+                if res is not None:
+                    return res
+
```
jd, line 280: I'm not a fan of the hook approach here. I'd rather do a classic inheritance.

mergify_engine/tests/unit/rules/test_get_rule_checks_status.py
```diff
@@ -29,6 +30,9 @@
 from mergify_engine.rules import conditions
 
 
+LOG = daiquiri.getLogger(__name__)
+
+
```
jd, line 35:
```suggestion
```

jd (changes requested):

mergify_engine/rules/filter.py
sileht: I can try, I was afraid to fight with mypy.

mergify_engine/rules/filter.py
```diff
@@ -491,3 +504,186 @@ def NearDatetimeFilter(
...
+            real_attr_name = attribute_name[1:]
+        else:
+            real_attr_name = attribute_name
+
+        if real_attr_name.startswith("check-") or real_attr_name.startswith("status-"):
+            if attribute_name.startswith(Filter.LENGTH_OPERATOR):
+                if len(self.pending_checks) != 0:
+                    return False
+            else:
+                if "pending" in real_attr_name:
```
jd, line 659:
```suggestion
                if real_attr_name.endswith("pending"):
```
let's avoid a bug later 😄

jd (dismissed):

mergify_engine/rules/filter.py
```diff
@@ -491,3 +504,189 @@ def NearDatetimeFilter(
...
+            real_attr_name = attribute_name
+
+        if real_attr_name.startswith("check-") or real_attr_name.startswith("status-"):
+            if attribute_name.startswith(Filter.LENGTH_OPERATOR):
+                if len(self.pending_checks) != 0:
+                    return False
+            else:
+                if real_attr_name in (
+                    "check-pending",
+                    "check-success-or-neutral-or-pending",
```
jd, line 661: do we need `status-` variant too? L654 seems to handle it.

sileht: status-pending doesn't exist.

jd (approved):

PR #3679, opened 2021-11-29
fix: track when speculative check start

sileht:
Partial-Fix: MRGFY-640

Change-Id: I62594694da0355d3402a6876a4c42534db650e31


mergify_engine/queue/merge_train.py
```diff
@@ -123,6 +123,9 @@ class TrainCar:
     still_queued_embarked_pulls: typing.List[EmbarkedPull]
     parent_pull_request_numbers: typing.List[github_types.GitHubPullRequestNumber]
     initial_current_base_sha: github_types.SHAType
+    creation_date: datetime.datetime = dataclasses.field(
+        default_factory=lambda: date.utcnow()
```
jd, line 127:
```suggestion
        default_factory=date.utcnow
```

jd (changes requested):

jd (approved):

PR #3695, opened 2021-12-01
feat(redis): add a generic migrations system

sileht:
This move out from the worker the migration system.

It changes how we stamps the fact we have done the migration script.

Change-Id: I8daeec7e29207ecba657929bbeedbe6b72473c18

### DIFF
--- a/mergify_engine/migrations/__init__.py
+++ b/mergify_engine/migrations/__init__.py
@@ -1,0 +1,63 @@
+# -*- encoding: utf-8 -*-
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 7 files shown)


mergify_engine/migrations/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+import typing
+
+import daiquiri
+import pkg_resources
+
+from mergify_engine import redis_utils
+from mergify_engine import utils
+
+
+LOG = daiquiri.getLogger(__name__)
```
jd, line 24: seems unused

jd (dismissed):

jd (approved):

PR #3697, opened 2021-12-01
fix(refresh): use pipeline on multiple refresh

sileht:
Fixes MRGFY-746

Change-Id: If174845d9b9e3186e21ed471d9df37a8257215fa

### DIFF
--- a/mergify_engine/delayed_refresh.py
+++ b/mergify_engine/delayed_refresh.py
@@ -74,7 +74,13 @@ async def send(
     redis_cache: utils.RedisCache,
 ) -> None:
     score = date.utcnow().timestamp()
-    for subkey in await redis_cache.zrangebyscore(DELAYED_REFRESH_KEY, "-inf", score):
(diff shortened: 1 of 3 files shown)


mergify_engine/delayed_refresh.py
```diff
@@ -110,5 +116,7 @@ async def send(
                 "source": "delayed-refresh",
             },  # type: ignore[typeddict-item]
         )
+        keys_to_delete.add(subkey)
 
-        await redis_cache.zrem(DELAYED_REFRESH_KEY, subkey)
+    await pipe.execute()
+    await redis_cache.zrem(DELAYED_REFRESH_KEY, *keys_to_delete)
```
jd, line 122: why not putting this in the pipe also?

sileht: that's not the same DB and it's just one redis call

jd (approved):

PR #3700, opened 2021-12-02
fix(cache): get each org its own summary-sha cache

sileht:
Fixes MRGFY-748

Change-Id: I2fb4f1cdccf0483fbac71e666642d6b0233d94d7


jd (commented):
Need rebase but excellent idea 👍

sileht:
> Need rebase but excellent idea +1

Yeah, but hash doesn't have expiration on Redis, this can't really work :(

PR #3703, opened 2021-12-02
feat(context): cache user/teams permissions

sileht:
This data is not going to change often we can keep it for the life time
of a context (maximum 30s).

Fixes MRGFY-750

Change-Id: Id3dabe978c4a826d1eb8fc7e6589ecc4b9a2cfba

### DIFF
--- a/mergify_engine/utils.py
+++ b/mergify_engine/utils.py
@@ -14,12 +14,15 @@
 # under the License.
 import asyncio
 import contextlib
+import dataclasses
+import functools
 import hashlib
 import hmac
 import math
 import os
 import socket
 import ssl
+import types
 import typing
 import uuid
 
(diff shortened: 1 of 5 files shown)


mergify_engine/utils.py
```diff
@@ -314,3 +315,22 @@ def split_list(
     while remaining:
         yield remaining[:size]
         remaining = remaining[size:]
+
+
+def async_cache(
+    user_function: typing.Callable[..., typing.Awaitable[_T]]
+) -> typing.Callable[..., typing.Awaitable[_T]]:
+    cache: typing.Dict[typing.Hashable, _T] = {}
```
jd, line 323: rather use the cache object from functools?
it has more stuff like stats etc.

sileht: I tried to keep it simple, for stats we need to catch miss/hit ourselves and fill the `functools._CacheInfo()` ourselves.

jd (approved):

PR #3709, opened 2021-12-03
fix(rules): fix regex conditions

sileht:
IncompleteChecksFilter use binary_op in the reverse way.
It works for all ops except regex that need to have the regex as second
argument.

This fixes that.

Fixes MERGIFY-ENGINE-2FJ

Change-Id: I37eb167ab80178e0e70d8a192d42826b93987d9c


jd (approved):

sileht:
@mergifyio rebase

sileht:
@Mergifyio rebase

PR #3715, opened 2021-12-03
fix(worker): remove space is tracer name

sileht:
Change-Id: I0e798392fe088cf317c3cc9c1c85508521974a85


jd (approved):
This is fine but not needed. :)

PR #3719, opened 2021-12-03
feat(context): cache user/teams permissions v2

sileht:
This change uses the same technique as other cached thing.

Fixes MRGFY-750

Change-Id: Idc27c44614edf6cfcda8109a54f6c73bf10d9ed8

### DIFF
--- a/mergify_engine/cache.py
+++ b/mergify_engine/cache.py
@@ -1,0 +1,61 @@
+#
+# Copyright © 2021 Mergify SAS
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+import dataclasses
+import enum
+import typing
+
+
+_K = typing.TypeVar("_K")
+_V = typing.TypeVar("_V")
+
+
+# NOTE(sileht): Sentinel object (eg: `marker = object()`) can't be expressed
+# with typing yet use the proposed workaround instead:
+#   https://github.com/python/typing/issues/689
+#   https://www.python.org/dev/peps/pep-0661/
+class UnsetType(enum.Enum):
+    _MARKER = 0
+
+
+Unset: typing.Final = UnsetType._MARKER
+
+
+@dataclasses.dataclass
+class Cache(typing.Generic[_K, _V]):
+    _cache: typing.Dict[_K, _V] = dataclasses.field(default_factory=dict)
+
+    def get(self, key: _K) -> typing.Union[_V, UnsetType]:
+        return self._cache.get(key, Unset)
+
+    def set(self, key: _K, value: _V) -> None:
+        self._cache[key] = value
+
+    def delete(self, key: _K) -> None:
+        try:
+            del self._cache[key]
+        except KeyError:
+            pass
+
+    def clear(self) -> None:
+        self._cache.clear()
+
+    def __setitem__(self, key: _K, value: _V) -> None:
+        return self.set(key, value)
+
+    def __getitem__(self, key: _K) -> typing.Union[_V, UnsetType]:
+        return self.get(key)
+
+    def __delitem__(self, key: _K) -> None:
+        self.delete(key)
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -38,6 +38,7 @@ import jinja2.utils
 import markdownify
 import tenacity
 
+from mergify_engine import cache
 from mergify_engine import check_api
 from mergify_engine import config
 from mergify_engine import constants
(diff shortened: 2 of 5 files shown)


mergify_engine/context.py
```diff
@@ -71,6 +71,12 @@ class PullRequestAttributeError(AttributeError):
     name: str
 
 
+class InstallationCache(typing.TypedDict, total=False):
```
jd, line 74: why not using a class?

sileht: I pick the same schema as other caches. Since the cache may cache "None", you need something to differentiate `not-cached` from `None` cached. The none presence of the key means not cached.

jd: That's true for the value of the key `team_members`, but the `InstallationCache` itself could be a class. 🤷 I don't see why it needs to be a `TypedDict`.

sileht: What about something like this: https://github.com/Mergifyio/mergify-engine/pull/3733. I think it's clearer because Cached / NotCached becomes explicit!

mergify_engine/context.py
```diff
@@ -220,27 +230,29 @@ async def clear_team_members_cache_for_org(
...
     ) -> typing.List[github_types.GitHubLogin]:
-
-        key = self._team_members_cache_key_for_repo(self.owner_id)
-        members_raw = await self.redis.hget(key, team_slug)
-        if members_raw is None:
-            members = [
-                github_types.GitHubLogin(member["login"])
-                async for member in self.client.items(
-                    f"/orgs/{self.owner_login}/teams/{team_slug}/members"
+        members = self._cache.team_members.get(team_slug, NotCached)
```
jd, line 233: Not a fan of having `NotCached` pass as an arg to `get`. It'd be better that `get` returns `NotCached` if it's not cached.

e.g. https://github.com/Mergifyio/mergify-dashboard/blob/main/mergify_dashboard/cache.py

jd (dismissed):

mergify_engine/context.py
sileht: I think you mix dashboard redis cache and engine in-memory cache. `self._cache.team_members` is a simple dict here. We pass `NotCached` to `.get()` to differentiate cached `None` and not cached.

jd: I don't mix anything, I'm just showing an example of a `Cache` class.

You shouldn't use a raw `dict` here because you are leaking the cache abstraction in the user by having it pass `NotCached` as the default argument.

Something like this would be cleaner:
```python
class Cache(dict[K, V]):
    def get(self, key: K) -> typing.Union[V, NotCached]:
        ...
```
(You can add support for the `default` arg but not sure you'd need it).

jd: Actually `NotCached` could be anything it's just a marker that you are abusing here.

So alternatively, if you want, you should rewrite the code like this:

```python
try:
   members = self._cache.team_members[team_slug]
except KeyError
   key = self._team_members_cache_key_for_repo(self.owner_id)
   ...
```

In that case, you don't need to use a marker, `NotCached` or any other.

mergify_engine/cache.py
```diff
@@ -0,0 +1,61 @@
...
+
+_K = typing.TypeVar("_K")
+_V = typing.TypeVar("_V")
+
+
+# NOTE(sileht): Sentinel object (eg: `marker = object()`) can't be expressed
+# with typing yet use the proposed workaround instead:
+#   https://github.com/python/typing/issues/689
+#   https://www.python.org/dev/peps/pep-0661/
+class _Unset(enum.Enum):
```
jd, line 28:
```suggestion
class UnsetType(enum.Enum):
```

mergify_engine/cache.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+Unset: typing.Final = _Unset._MARKER
+
+
+@dataclasses.dataclass
+class Cache(typing.Generic[_K, _V]):
+    _cache: typing.Dict[_K, _V] = dataclasses.field(default_factory=dict)
+
+    def get(self, key: _K) -> typing.Union[_V, _Unset]:
```
jd, line 39:
```suggestion
    def get(self, key: _K) -> typing.Union[_V, UnsetType]:
```

mergify_engine/cache.py
```diff
@@ -0,0 +1,61 @@
...
+        except KeyError:
+            pass
+
+    def clear(self) -> None:
+        self._cache.clear()
+
+    def __setitem__(self, key: _K, value: _V) -> None:
+        return self.set(key, value)
+
+    def __getitem__(self, key: _K) -> typing.Union[_V, _Unset]:
```
jd, line 57:
```suggestion
    def __getitem__(self, key: _K) -> typing.Union[_V, UnsetType]:
```

mergify_engine/context.py
```diff
@@ -492,22 +511,27 @@ async def get_user_permission(
...
-                github_types.GitHubRepositoryCollaboratorPermission,
-                await self.installation.client.item(
-                    f"{self.base_url}/collaborators/{user['login']}/permission"
-                ),
-            )["permission"]
-            pipe = await self.installation.redis.pipeline()
-            await pipe.hset(key, user["id"], permission)
-            await pipe.expire(key, self.USERS_PERMISSION_EXPIRATION)
-            await pipe.execute()
+        permission = self._cache.user_permissions.get(user["id"], NotCached)
```
jd, line 514: use the Cache class?

jd (changes requested):

mergify_engine/context.py
sileht: Damn, I squash it in the wrong patch...

PR #3727, opened 2021-12-04
feat(context): store less data in check-runs cache

sileht:
check-runs contains some useless data like, many app details
the app owner and permissions that we don't really need.

Fixes MRGFY-758

Change-Id: If31ed2b93014d2a900d30ef2ee1baaa04aa75fd7


mergify_engine/check_api.py
```diff
@@ -148,20 +180,24 @@ def compare_dict(
...
+    previous_check: github_types.GitHubCheckRunLight,
     expected_check: GitHubCheckRunParameters,
 ) -> bool:
     if compare_dict(
         typing.cast(typing.Dict[str, typing.Any], expected_check),
         typing.cast(typing.Dict[str, typing.Any], previous_check),
         ("head_sha", "status", "conclusion", "details_url"),
     ):
-        if previous_check["output"] == expected_check["output"]:
+        if previous_check["output"] is None and expected_check["output"] is None:
```
jd, line 191: a little bit worried by this being changed, I imagine it's related to this PR?

mergify_engine/github_types.py
```diff
@@ -472,6 +478,24 @@ class GitHubCheckRun(typing.TypedDict):
...
     output: GitHubCheckRunOutput
     conclusion: typing.Optional[GitHubCheckRunConclusion]
+    started_at: ISODateTimeType
+    completed_at: ISODateTimeType
+    html_url: str
+    details_url: str
+    check_suite: GitHubCheckRunCheckSuite
+
+
+class GitHubCheckRunLight(typing.TypedDict):
```
jd, line 488: Stupid question but how "light" it is? I don't see much data being removed?

sileht: A lot !!!! (btw, I plan will rename it `CachedGitHubCheckRun`)

The checks["app"] has always the same size and contains le owner details, the permissions details and some of its settings we don't care (I would say 30-40 keys)

The one that can be really big is `checks["output"]["summary"]` it's still present for our check-runs, but we drop it for other CIs.

mergify_engine/check_api.py
sileht: I will fix that separately, the condition has always been false, when I  added the missing Check-runs attributes, mypy complains.

mergify_engine/github_types.py
jd: If your plan for this PR or a next one?

mergify_engine/check_api.py
```diff
@@ -113,22 +114,53 @@ class Result:
     ended_at: typing.Optional[datetime.datetime] = None
 
 
+def to_check_run_light(
```
jd, line 117: I don't know how hard it is, but in theory calling `del` on the original and then casting it would be more efficient.

sileht: Using del on typing.TypedDict is mypy hell

sileht: you can't build a TypeDict from another TypedDict, the only ways is keeping typing.Any, delete keys, typing.cast. And you are not sure about what the dict really contains it you missing to del some keys or GitHub add new keys.

sileht: So copying is a dumb from memory handling perspective, but at least the final dict is the expected for sure.

jd (approved):

PR #3733, opened 2021-12-06
chore(context): make RepositoryCache more solid

sileht:
Change-Id: I6a99c07262dfdf33c4f52cf4144bafab6b161b51

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -18,6 +18,7 @@ import base64
 import contextlib
 import dataclasses
 import datetime
+import enum
 import functools
 import itertools
 import json
@@ -54,6 +55,17 @@ SUMMARY_SHA_EXPIRATION = 60 * 60 * 24 * 31 * 1  # 1 Month
 
 MARKDOWN_TITLE_RE = re.compile(r"^#+ ", re.I)
 MARKDOWN_COMMIT_MESSAGE_RE = re.compile(r"^#+ Commit Message ?:?\s*$", re.I)
+
+
+# NOTE(sileht): Sentinel object (eg: `marker = object()`) can't be expressed
+# with typing yet use the proposed workaround instead:
+#   https://github.com/python/typing/issues/689
+#   https://www.python.org/dev/peps/pep-0661/
+class _NotCached(enum.Enum):
+    _MARKER = 0
+
+
+NotCached: typing.Final = _NotCached._MARKER
 
 
 class MergifyConfigFile(github_types.GitHubContentFile):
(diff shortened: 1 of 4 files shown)


mergify_engine/context.py
```diff
@@ -56,6 +57,17 @@
...
 
+# NOTE(sileht): Sentinel object (eg: `marker = object()`) can't be expressed
+# with typing yet use the proposed workaround instead:
+#   https://github.com/python/typing/issues/689
+#   https://www.python.org/dev/peps/pep-0661/
+class _NotCached(enum.Enum):
+    _MARKER = 0
+
+
+NotCached: typing.Final = _NotCached._MARKER
```
jd, line 68: We're using this a lot, I'm wondering if having a general `Unset` marker in a util module wouldn't be simpler.

jd (commented):
👍

mergify_engine/context.py
sileht: Each thing should have it's marker in theory to avoid bug like:

```
def bar(o = Unset):
      if o is Unset:
          print("I'm unset)

def foo(o = Unset):
      bar(o)

>>> foo()
I'm unset
```
o has been set for bar, but detected as unset.

jd: I'm not sure it's a bug in this case. The alternative would be:
```python
def bar(o=UnsetBar):
   if o is UnsetBar:
       print("I'm unset")

def foo(o=UnsetFoo):
    bar(o)
```

Except that calling `bar` with `UnsetFoo` as value has no sense at all, so it'd likely be a bug.

Your example has semantic sense, mine does not.
Your example works as expected, mine would probably raise an error.

sileht: > Except that calling bar with UnsetFoo as value has no sense at all, so it'd likely be a bug.

A bug cached by mypy

jd: > A bug cached by mypy

Right, but it's still a bug. Why do you want to have a bug by default? 🤔

sileht: I'm just saying in my example, you also have the bug, but mypy will not complain as type match.

If `foo()` and `bar()` play well together and are related why not sharing the `Unset` as the meaning/compat is also shared. 

I don't like copy/pasting the enum hack for marker and hope mypy implement it nicely.

jd: That's my point: `Unset` always have the same meaning, e.g., `Unset` argument in this case.
That's why it makes total sense to use only one `Unset` type.

That being said I think your example is not well chosen for this case here.

jd (approved):

PR #3735, opened 2021-12-06
chore(caches): use cache module for ReposityCaches/ContextCaches

sileht:
Change-Id: Iadaaafa22d03d9a161604275c94a131f096f455c

### DIFF
--- a/mergify_engine/cache.py
+++ b/mergify_engine/cache.py
@@ -59,3 +59,17 @@ class Cache(typing.Generic[_K, _V]):
 
     def __delitem__(self, key: _K) -> None:
         self.delete(key)
+
+
+@dataclasses.dataclass
+class SingleCache(typing.Generic[_V]):
+    _cache: typing.Union[_V, UnsetType] = dataclasses.field(default=Unset)
+
+    def get(self) -> typing.Union[_V, UnsetType]:
+        return self._cache
+
+    def set(self, value: _V) -> None:
+        self._cache = value
+
+    def delete(self) -> None:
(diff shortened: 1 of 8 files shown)


mergify_engine/cache.py
```diff
@@ -59,3 +59,17 @@ def __getitem__(self, key: _K) -> typing.Union[_V, UnsetType]:
...
+    _cache: typing.Union[_V, UnsetType] = dataclasses.field(default=Unset)
+
+    def get(self) -> typing.Union[_V, UnsetType]:
+        return self._cache
+
+    def set(self, value: _V) -> None:
+        self._cache = value
+
+    def delete(self) -> None:
+        self._cache = Unset
```
jd, line 75: What about using the descriptor protocol here?

sileht: Looks fancy at first glance:

```
assert self._caches.mergify_config is Unset:
self._caches.mergify_config = "value"
assert self._caches.mergify_config == "value"
del self._caches.mergify_config
assert self._caches.mergify_config is Unset
```

jd: Is this a yes?

sileht: I just read the doc and it's `ContextCache` that have to be description protocol, but currently the `ContextCache` has `Cache` and `SingleCache` attributes. The value `SingleCache` as-is is the typing is simple and strong. If I have to description the same typing with a descriptor, I have to create an @overload for key the description protocol support.

jd (approved):

PR #3741, opened 2021-12-07
chore(tests): test if empty queue conditions wait the CIs

sileht:
It doesn't ....

Change-Id: I20320583fc3a00cc29cf98de69f35b670eba6208


sileht:
Just for the memory

PR #3742, opened 2021-12-07
feat(actions): new conditions thread-resolved and threads-unresolved

ErwanSimonetti:
Added 2 conditions related to review threads: resolved review threads and unresolved review-threads

Fixes MRGFY-591


mergify_engine/clients/github.py
```diff
@@ -387,6 +387,13 @@ async def delete(  # type: ignore[override]
             url, **self._prepare_request_kwargs(api_version, oauth_token, **kwargs)
         )
 
+    async def graphql_post(self, query: str) -> typing.Any:
+        response = (await self.post("/graphql", json={"query": query})).json()
+        data = response.get("data")
+        if not data:
+            raise ConnectionError(response)
```
sileht, line 394: ConnectionError is too generic. Create a new exception class like `GraphQLError`.

sileht (changes requested):

mergify_engine/context.py
```diff
@@ -752,6 +753,66 @@ async def retrieve_unverified_commits(self) -> typing.List[str]:
             if not commit["commit_verification_verified"]
         ]
 
+    async def retrieve_resolved_review_threads(
+        self, resolved: typing.Optional[bool] = True
+    ) -> ContextAttributeType:
```
sileht, line 758: The returns type is too broad, it should be just what return the function `typing.List[...]`

mergify_engine/context.py
```diff
@@ -752,6 +753,66 @@ async def retrieve_unverified_commits(self) -> typing.List[str]:
...
+                    "number": self.pull["number"],
+                },
+            ),
+            send_fn=self.client.graphql_post,
+        )
+        resolved_review_threads = []
+        query_number = 0
+        current_query = "Q" + str(query_number)
+        async for response in responses:
+            while current_query in response:
```
sileht, line 796: I don't get why this is a while loop. 

`for current_query, inner_response in response.items():` should work ?

sileht (changes requested):

mergify_engine/tests/functional/base.py
```diff
@@ -748,6 +748,82 @@ async def create_comment(
...
+        await self.client_admin.post(
+            f"{self.url_origin}/pulls/{pull_number}/comments",
+            json={"body": message, "in_reply_to": comment_id},
+        )
+        await self.wait_for("pull_request_review_comment", {"action": "created"})
+
+    async def resolve_review_thread(
+        self,
+        thread_id: str,
+    ) -> typing.Any:
```
sileht, line 813:
```suggestion
    ) -> bool:
```

mergify_engine/tests/functional/base.py
```diff
@@ -748,6 +748,82 @@ async def create_comment(
...
+                "body": message,
+                "path": path,
+                "commit_id": commits[-1]["sha"],
+                "line": line,
+            },
+        )
+        await self.wait_for("pull_request_review_comment", {"action": "created"})
+        return response.json()["id"]
+
+    async def get_review_threads(self, number):
```
sileht, line 771: typing is missing

mergify_engine/tests/functional/base.py
```diff
@@ -748,6 +748,82 @@ async def create_comment(
         )
         await self.wait_for("issue_comment", {"action": "created"})
 
+    async def create_review_thread(
+        self,
+        pull_number: github_types.GitHubPullRequestNumber,
+        message: str,
+        line: typing.Optional[int] = 1,
+        path: typing.Optional[str] = "test1",
+    ) -> typing.Any:
```
sileht, line 757:
```suggestion
    ) -> int:
```

mergify_engine/tests/functional/test_attributes.py
```diff
@@ -508,6 +510,194 @@ async def test_two_commits_verified(self):
         comments = await self.get_issue_comments(pr["number"])
         self.assertEqual("commits verified", comments[-1]["body"])
 
+    async def test_retrieve_zero_resolved_threads(self):
```
sileht, line 513:
```suggestion
    async def test_retrieve_zero_resolved_threads(self) -> None:
```

mergify_engine/tests/functional/test_attributes.py
```diff
@@ -508,6 +510,194 @@ async def test_two_commits_verified(self):
...
+            pr["number"], "you shouldn't write `foo` here", path="yves_testing_file"
+        )
+        await self.run_engine()
+        await self.wait_for("issue_comment", {"action": "created"})
+        comments = await self.get_issue_comments(pr["number"])
+        self.assertEqual("review-threads-resolved comment", comments[-1]["body"])
+        review_threads = await self.get_review_comments(pull_number=pr["number"])
+        self.assertEqual("you shouldn't write `foo` here", review_threads[-1]["body"])
+
+    async def test_retrieve_unresolved_threads(self):
```
sileht, line 537:
```suggestion
    async def test_retrieve_unresolved_threads(self) -> None:
```

mergify_engine/tests/functional/test_attributes.py
```diff
@@ -508,6 +510,194 @@ async def test_two_commits_verified(self):
...
+        self.assertEqual("much originality", review_threads[3]["body"])
+        self.assertEqual(
+            "Don't like this line too much either", review_threads[2]["body"]
+        )
+        self.assertEqual("How original", review_threads[1]["body"])
+        self.assertEqual(
+            "why are you still writing `foo` here ?", review_threads[0]["body"]
+        )
+
+    async def test_retrieve_resolved_and_unresolved_threads(self):
```
sileht, line 591:
```suggestion
    async def test_retrieve_resolved_and_unresolved_threads(self) -> None:
```

mergify_engine/tests/functional/test_attributes.py
```diff
@@ -508,6 +510,194 @@ async def test_two_commits_verified(self):
...
+        thread_2 = (await self.get_review_threads(pr["number"]))["repository"][
+            "pullRequest"
+        ]["reviewThreads"]["edges"][2]["node"]
+        assert thread_2["isResolved"]
+        await self.run_engine()
+        await self.wait_for("issue_comment", {"action": "created"})
+        comments = await self.get_issue_comments(pr["number"])
+        self.assertEqual("conditions matched; s u c c e s s", comments[-1]["body"])
+
+    async def test_retrieve_message_resolved_thread(self):
```
sileht, line 656:
```suggestion
    async def test_retrieve_message_resolved_thread(self) -> None:
```

mergify_engine/tests/functional/test_attributes.py
```diff
@@ -508,6 +510,194 @@ async def test_two_commits_verified(self):
...
+        }
+        await self.setup_repo(yaml.dump(rules))
+        pr, _ = await self.create_pr(files={"yves_testing_file": "foo"})
+        await self.create_review_thread(
+            pr["number"], "you shouldn't write `foo` here", path="yves_testing_file"
+        )
+        await self.run_engine()
+        await self.wait_for("issue_comment", {"action": "created"})
+        comments = await self.get_issue_comments(pr["number"])
+        self.assertEqual("review-threads-resolved comment", comments[-1]["body"])
```
sileht, line 533: We try to avoid using `self.assertEqual()`, and just use the raw `assert`.

mergify_engine/clients/github.py
```diff
@@ -43,6 +43,11 @@ class TooManyPages(Exception):
     response: httpx.Response
 
 
+@dataclasses.dataclass
+class GraphqlError(Exception):
+    response: httpx.Response
```
sileht, line 48: The typing looks wrong to me, this is not a `httpx.Response`, but a dictionary. 
Since the error format is not documented by GitHub, just put `str` for now.

mergify_engine/clients/github.py
```diff
@@ -387,6 +392,13 @@ async def delete(  # type: ignore[override]
             url, **self._prepare_request_kwargs(api_version, oauth_token, **kwargs)
         )
 
+    async def graphql_post(self, query: str) -> typing.Any:
+        response = (await self.post("/graphql", json={"query": query})).json()
+        data = response.get("data")
+        if not data:
+            raise GraphqlError(response)
+        return response
```
sileht, line 400: We should put the string into the exception for now, as we don't known that the error structure looks like.

```suggestion
        response = await self.post("/graphql", json={"query": query})
        data = response.json()
        if "data" not in data:
            raise GraphqlError(response.text)
        return data
```

mergify_engine/context.py
```diff
@@ -788,6 +790,69 @@ async def retrieve_unverified_commits(self) -> typing.List[str]:
...
+            ),
+        )
+        resolved_review_threads = []
+        query_number = 0
+        current_query = "Q" + str(query_number)
+        async for response in responses:
+            for current_query, _inner_response in response.items():
+                if (
+                    len(
+                        response[current_query]["pullRequest"]["reviewThreads"]["edges"]
```
sileht, line 839:
```suggestion
                        current_response["pullRequest"]["reviewThreads"]["edges"]
```

mergify_engine/context.py
```diff
@@ -788,6 +790,69 @@ async def retrieve_unverified_commits(self) -> typing.List[str]:
...
+                        "name": self.repository.repo["name"],
+                        "number": self.pull["number"],
+                    },
+                ),
+                send_fn=self.client.graphql_post,
+            ),
+        )
+        resolved_review_threads = []
+        query_number = 0
+        current_query = "Q" + str(query_number)
```
sileht, line 834: These variables are not needed. You already have the `response[current_query]` value on each loop, thank to `.items()`

https://docs.python.org/3/tutorial/datastructures.html#looping-techniques

```suggestion
```

mergify_engine/context.py
```diff
@@ -788,6 +790,69 @@ async def retrieve_unverified_commits(self) -> typing.List[str]:
...
+                        "reviewThreads"
+                    ]["edges"]:
+                        if review_comment["node"]["isResolved"] == resolved:
+                            resolved_review_threads.append(
+                                review_comment["node"]["comments"]["edges"][0]["node"][
+                                    "body"
+                                ]
+                            )
+                query_number += 1
+                current_query = "Q" + str(query_number)
```
sileht, line 853:
```suggestion
```

mergify_engine/context.py
```diff
@@ -788,6 +790,69 @@ async def retrieve_unverified_commits(self) -> typing.List[str]:
...
+        current_query = "Q" + str(query_number)
+        async for response in responses:
+            for current_query, _inner_response in response.items():
+                if (
+                    len(
+                        response[current_query]["pullRequest"]["reviewThreads"]["edges"]
+                    )
+                    > 0
+                ):
+                    for review_comment in response[current_query]["pullRequest"][
```
sileht, line 843:
```suggestion
                    for review_comment in current_response["pullRequest"][
```

mergify_engine/context.py
```diff
@@ -788,6 +790,69 @@ async def retrieve_unverified_commits(self) -> typing.List[str]:
...
+                    },
+                ),
+                send_fn=self.client.graphql_post,
+            ),
+        )
+        resolved_review_threads = []
+        query_number = 0
+        current_query = "Q" + str(query_number)
+        async for response in responses:
+            for current_query, _inner_response in response.items():
```
sileht, line 836:
```suggestion
            for current_query, current_response in response.items():
```

sileht (changes requested):

sileht (dismissed):
Can you add theses new attributes in the the `docs/`

mergify_engine/clients/github.py
```diff
@@ -43,6 +43,11 @@ class TooManyPages(Exception):
     response: httpx.Response
 
 
+@dataclasses.dataclass
+class GraphqlError(Exception):
+    response: str
```
sileht, line 48: nit: ```suggestion
    message: str
```

File: mergify_engine/github_graphql_types.py [hunk 20]
```diff
@@ -0,0 +1,70 @@
...
+class GraphqlComment(typing.TypedDict):
+    body: str
+
+
+class GraphqlCommentsEdges(typing.TypedDict):
+    node: GraphqlComment
+
+
+class GraphqlComments(typing.TypedDict):
+    edges: typing.List[GraphqlCommentsEdges]
```
sileht, line 28:
```suggestion
    edges: typing.List[GraphqlCommentsEdge]
```

releasenotes/notes/resolved-conversation-63f566d08c4b2b34.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    The attributes `review-threads-resolved` & `review-threads-unresolved` have been added.
```
sileht, line 4:
```suggestion
    The attributes ``review-threads-resolved`` & ``review-threads-unresolved`` have been added.
```

mergify_engine/github_graphql_types.py
```diff
@@ -0,0 +1,70 @@
...
+class GraphqlReviewThreadsQuery(typing.TypedDict):
+    repository: GraphqlRepository
+
+
+class GraphqlDataReviewThreads(typing.TypedDict):
+    repository: GraphqlRepository
+
+
+class GraphqlReviewThreadsQueryData(typing.TypedDict):
+    data: GraphqlDataReviewThreads
```
sileht, line 70: Some query has a type with `data` some not, we directly set the type of `response["data"]`.

We should choose one or the other.

I'm in favor of dropping types with just `data: ...`

mergify_engine/github_graphql_types.py
```diff
@@ -0,0 +1,70 @@
...
+
+class GraphqlRepository(typing.TypedDict):
+    pullRequest: GraphqlPullRequest
+
+
+class GraphqlReviewThreadsQuery(typing.TypedDict):
+    repository: GraphqlRepository
+
+
+class GraphqlDataReviewThreads(typing.TypedDict):
```
sileht, line 65: This type looks identical to the one above.

mergify_engine/github_graphql_types.py
```diff
@@ -0,0 +1,62 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2020 Mergify SAS
```
sileht, line 3:
```suggestion
# Copyright © 2021 Mergify SAS
```

mergify_engine/context.py
```diff
@@ -788,6 +790,60 @@ async def retrieve_unverified_commits(self) -> typing.List[str]:
...
+        async for response in responses:
+            for _current_query, current_response in response.items():
+                if len(current_response["pullRequest"]["reviewThreads"]["edges"]) > 0:
+                    for review_comment in current_response["pullRequest"][
+                        "reviewThreads"
+                    ]["edges"]:
+                        if review_comment["node"]["isResolved"] == resolved:
+                            resolved_review_threads.append(
+                                review_comment["node"]["comments"]["edges"][0]["node"][
+                                    "body"
```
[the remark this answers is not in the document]
ErwanSimonetti: @sileht wanted to use the body, my idea was to use the line and file associated to the review thread but you can't use dictionnaries in attributes :slightly_smiling_face: 
Also, it is indeed the body of the first message creating the thread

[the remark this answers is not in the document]
jd: I doubt the first body of the thread is that interesting. Might be better to use the author of the first review?
The real question is @sileht when can we export more complicated things like full Python dict (= object) with arrays and all and access them with `.`? That'd solve everything 😁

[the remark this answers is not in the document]
sileht: It's always hard to select the correct info without a valid use-case, a simpler solution could expose only the number of threads.

> The real question is @sileht when can we export more complicated things like full Python dict (= object) with arrays and all and access them with .?

Such an interesting topic to improve the "filters" for R&D next year. ;)

So my conclusion was to use something that nobody can use, so we can change it later :p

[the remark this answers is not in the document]
ErwanSimonetti: Should I change it to the ID then ?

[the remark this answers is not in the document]
jd: No, let's keep it that way for no. I think @sileht is right, even if it sounds stupid, to have something barely usable should make sure nobody uses it in an unattended way for now.

jd (approved):

jd (changes requested):
maybe fix PR title though?

jd (approved):

PR #3743, opened 2021-12-07
fix(queue/rules): use a Filter to found pending/final checks

sileht:
Instead of the heuristic that looks only for `check-success=XXX`.

This creates a Filter that introduces a new possible result
`IncompleteCheck`.

The filter has a different meaning for or/and when boolean are combined
with IncompleteCheck:
* or cannot return false if at least one is IncompleteCheck
* and cannot return false if at least one is IncompleteCheck

To compute if a check conditions is final or not, this change adds an
hook in Filter before the evaluation of the condition to return IncompleteCheck
if need instead of the real evaluation outcome.

Fixes MRGFY-729
Fixes MRGFY-730
Fixes MRGFY-731
Fixes MRGFY-736
Fixes MRGFY-752

Change-Id: Ib99d93348934d91c556f14756a8da85e81e19160

### DIFF
--- a/mergify_engine/rules/filter.py
+++ b/mergify_engine/rules/filter.py
@@ -16,6 +16,7 @@
 from collections import abc
 import dataclasses
 import datetime
+import enum
 import inspect
 import operator
 import re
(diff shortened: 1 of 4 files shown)


sileht:
I just added more tests but didn't reproduce the issue we got in production :(.

mergify_engine/rules/filter.py
```diff
@@ -491,3 +504,189 @@ def NearDatetimeFilter(
             "and": _minimal_datetime,
         },
     )
+
+
+# NOTE(sileht): Sentinel object (eg: `marker = object()`) can't be expressed
+# with typing yet use the proposed workaround instead:
+#   https://github.com/python/typing/issues/689
+#   https://www.python.org/dev/peps/pep-0661/
+class _IncompleteMarker(enum.Enum):
```
jd, line 513:
```suggestion
class IncompleteMarkerType(enum.Enum):
```

sileht:
Replaced by https://github.com/Mergifyio/mergify-engine/pull/3744

PR #3746, opened 2021-12-08
fix(queue): retry merge queue temporary branch creation

sileht:
The base branch may take a while to be created on certain repository.
This change retries the merge as we just created the branch and we known
it exists.

Fixes MRGFY-768

Change-Id: Ie6b8ff009be0677f0daa845ac26185a5d448308b

### DIFF
--- a/mergify_engine/queue/merge_train.py
+++ b/mergify_engine/queue/merge_train.py
@@ -22,6 +22,7 @@ from urllib import parse
 
 import daiquiri
 import first
+import tenacity
 
 from mergify_engine import branch_updater
 from mergify_engine import check_api
@@ -57,6 +58,10 @@ CHECK_ASSERTS = {
     "neutral": "https://raw.githubusercontent.com/Mergifyio/mergify-engine/master/assets/square-grey-16.png",
     "stale": "https://raw.githubusercontent.com/Mergifyio/mergify-engine/master/assets/square-grey-16.png",
 }
+
+
+def is_base_branch_not_exists_exception(exc: BaseException) -> bool:
+    return isinstance(exc, http.HTTPNotFound) and "Base does not exist" in exc.message
 
 
 class UnexpectedChange:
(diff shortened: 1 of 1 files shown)


mergify_engine/queue/merge_train.py
```diff
@@ -59,6 +60,10 @@
 }
 
 
+def is_base_branch_not_exists_exception(exc: BaseException) -> bool:
+    return isinstance(exc, http.HTTPNotFound) and "Base does not exist" in exc.message
```
jd, line 64: no error code to use rather?

mergify_engine/queue/merge_train.py
```diff
@@ -407,16 +412,39 @@ async def create_pull(
...
+                                "base": branch_name,
+                                "head": f"refs/pull/{pull_number}/head",
+                                "commit_message": f"Merge of #{pull_number}",
+                            },
+                        )
             except http.HTTPClientSideError as e:
-                if (
+                if is_base_branch_not_exists_exception(e):
+                    self.train.log.warning(
+                        "fail to create the queue pull request because base still doesn't exists, 1.5 seconds after it creation",
```
jd, line 437:
```suggestion
                        "fail to create the queue pull request because base still doesn't exist, 1.5 seconds after its creation",
```

mergify_engine/queue/merge_train.py
```diff
@@ -407,16 +412,39 @@ async def create_pull(
...
-                        "base": branch_name,
-                        "head": f"refs/pull/{pull_number}/head",
-                        "commit_message": f"Merge of #{pull_number}",
-                    },
-                )
+                # NOTE(sileht): From time to time, GitHub returns a 404 when we merge
+                # the pull request in branches because the branch doesn't exists yet
+                # even if the previous API call returns
+                async for attempt in tenacity.AsyncRetrying(
+                    wait=tenacity.wait_exponential(multiplier=0.1),
```
jd, line 419: Do we need exponential?
Or 4 times every 0.5s is good enough?
(no clue)

jd (changes requested):

mergify_engine/queue/merge_train.py
sileht: No this one has only `message` set.

mergify_engine/queue/merge_train.py
sileht: I checked the log, it only happens 40 times in a week (all org mixed), so I guess a very short retry should be enough to retry fast here ~1.5s and if it still fail, we PostPone the creation of the PR and on next worker pass we will retry.

sileht:
@mergify squash

jd (approved):

PR #3761, opened 2021-12-09
feat(queue): allow to configure checks timeout

sileht:
This inject a new condition that can unqueue pull request automatically.

Since this condition is a timedelta, the pull request is automatically
refresh when the timeout is reached.

Fixes MRGFY-640

Change-Id: I175b07c8cb0a4cb582b75aae6c7bd27dfcae4636


sileht:
image

sileht:
https://github.com/mergifyio-testing/functional-testing-repo-sileht/pull/3249/checks?check_run_id=4473765844

docs/source/actions/queue.rst
```diff
@@ -356,6 +356,12 @@ A ``queue_rules`` takes the following parameter:
        between 1 and 20.
        See :ref:`speculative checks`.
 
+   * - ``checks_timeout``
+     - string or int
```
jd, line 360:
```suggestion
     - string
```

We can convert it to int as needed 😉

docs/source/actions/queue.rst
```diff
@@ -356,6 +356,12 @@ A ``queue_rules`` takes the following parameter:
...
+     - none
```
jd, line 361:
```suggestion
     - 
```
We set it to empty when it's none.

docs/source/actions/queue.rst
```diff
@@ -356,6 +356,12 @@ A ``queue_rules`` takes the following parameter:
...
+     - The times Mergify wait for Checks to succeed before automatically unqueue pull requests.
+       at least 60s. 50s, 5m or 1h can also be used for 1 second, 1 minute or 1 hour.
```
jd, line 363:
```suggestion
     - The amonut of time Mergify wait for pending checks to return before unqueueing pull requests.
       This cannot be less than 60 seconds.
```

You need to document the format. It should be similar to the time ones we already used, though maybe they are not compatible since they used "ago".

mergify_engine/actions/merge_base.py
```diff
@@ -338,7 +344,7 @@ async def _run(
                 return check_api.Result(
                     check_api.Conclusion.CANCELLED,
                     "The pull request has been removed from the queue",
-                    "The queue conditions cannot be satisfied due to failing checks. "
+                    "The queue conditions cannot be satisfied due to failing checks or checks timeout. "
```
jd, line 347: can we know which one?

mergify_engine/context.py
```diff
@@ -1012,6 +1013,16 @@ async def _get_consolidated_data(self, name: str) -> ContextAttributeType:
                 return -1
             return position
 
+        elif name == "queue-speculative-check-elapsed-time":
```
jd, line 1016: that does not work for inplace?

jd (changes requested):

mergify_engine/actions/merge_base.py
sileht: The details is on the other check-run, this message continues with https://github.com/Mergifyio/mergify-engine/blob/main/mergify_engine/actions/queue.py#L48.

mergify_engine/context.py
sileht: Yes, it does :)

docs/source/actions/queue.rst
sileht: I hesitate on the format, this one is simple but doesn't look like the other.
The Postgres format may work if we remove the `ago` (eg: `2 days 18:02` or `18:02`). But it feels weird for a duration.
The ISO format is even worse: `P2DT18H02M0S`
If we keep this format do we support think like `1h 2m`?

jd: I think you want to read https://www.postgresql.org/docs/9.1/datatype-datetime.html point 8.5.4 😄

mergify_engine/context.py
jd: The name sounds strange then?

sileht: When even we rebase a PR instead of creating a draft one, it's still a speculative checking.

docs/source/actions/queue.rst
sileht: Yeah that the best one, we can't support Y and M has python timedelta doesn't support it

sileht: § 8.5.4 support many formats:



Example | Description | Should we support it ?
-- | -- | --
1-2 | SQL standard format: 1 year 2 months | NO
3 4:05:06 | SQL standard format: 3 days 4 hours 5 minutes 6 seconds | MAYBE
1 year 2 months 3 days 4 hours 5 minutes 6 seconds | Traditional Postgres format: 1 year 2 months 3 days 4 hours 5 minutes 6 seconds | YES
P1Y2M3DT4H5M6S | ISO 8601 "format with designators": same meaning as above | NO
P0001-02-03T04:05:06 | ISO 8601 "alternative format": same meaning as above | NO

mergify_engine/context.py
jd: Err no, in place cannot be speculative. Speculative means we _speculate_ the first one work, so the check of the first PR in the queue is *not* speculative.

sileht: Ok, I see your point! so maybe just `queue-checks-elapsed-time`.

docs/source/actions/queue.rst
jd: 100% agree. Your `maybe` is a `no` to me.
We actually say in the code IIRC we support Postgre format, not ISO nor SQL.

mergify_engine/date.py
```diff
@@ -261,3 +263,40 @@ def fromtimestamp(timestamp: float) -> datetime.datetime:
...
+    if not kw or not kw["filled"]:
+        raise InvalidDate("Invalid date interval")
+
+    td = datetime.timedelta(
+        days=int(kw["days"] or 0),
+        hours=int(kw["hours"] or 0),
+        minutes=int(kw["minutes"] or 0),
+        seconds=int(kw["seconds"] or 0),
+    )
+    if min_ is not None and td < min_:
```
jd, line 299: Not a fan of having this kind of feature in a parser.
Not a parsing problem.

mergify_engine/tests/functional/actions/test_queue.py
```diff
@@ -2624,6 +2628,54 @@ async def test_more_ci_in_pull_request_rules_failure(self):
...
+        self.addCleanup(p.stop)
+
+        config = {
+            "queue_rules": [
+                {
+                    "name": "default",
+                    "conditions": [
+                        "check-success=continuous-integration/fake-ci",
+                    ],
+                    "checks_timeout": "10 m",
```
jd, line 2649: why not settings `1 s` here?

jd (dismissed):

mergify_engine/tests/functional/actions/test_queue.py
sileht: The min is 60s

mergify_engine/date.py
sileht: I will move it in voluptuous side

mergify_engine/context.py
```diff
@@ -1012,6 +1013,16 @@ async def _get_consolidated_data(self, name: str) -> ContextAttributeType:
                 return -1
             return position
 
+        elif name == "queue-checks-elapsed-time":
```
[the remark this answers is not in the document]
sileht: Let's wait we really need it, it's easier to add things than to remove/rename/refactor them.

And while I'm writing that I think of some other feature:

```
conditions:
  - queue-checks-elapsed-time > 
actions:
  comment:
      message: Hey sileht, the CIs should have finished you may need to look at it.
```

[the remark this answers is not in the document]
jd: That was my thinking too.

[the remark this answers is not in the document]
jd: Though I think the name of the field is a bit of a misnomer. There's nothing check-related.
I agree time-in-queue is right now linked to checks-elapsed-time but that not be always the case.
Might be better to name this `queue-waiting-time` or something like that.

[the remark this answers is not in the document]
sileht: Or we should expose a relative time attribute:

```
conditions:
  - queue-checks-started-at > 1 hours ago 
```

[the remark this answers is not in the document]
jd: That'd work too, though I'd name it `queued-merge-started-at` or the like?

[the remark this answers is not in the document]
sileht: Yeah, `queued-merge-started-at` soon better, no ?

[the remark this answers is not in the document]
jd: I meant `queue-merge-started-at` (no `d`) but yes!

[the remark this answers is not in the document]
sileht: Should we also expose add `queued-at` attribute to ensure users don't mix them.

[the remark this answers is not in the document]
jd: Yes, I actually think we did! lol

docs/source/conditions.rst
```diff
@@ -204,6 +204,9 @@ Here's the list of pull request attribute that can be used in conditions:
      - The position of the pull request in its queue if queued. The first pull
        request in the queue has position ``0``. The value is set to ``-1`` if
        the pull request is not queued.
+   * - ``queue-merge-started-at``
+     - integer
```
sileht, line 208: timestamp or relative timestamp

releasenotes/notes/queue-merge-started-at-b8efc3b8ecf348f5.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Add ``queue-merge-started-at`` condition.
```
jd, line 4: Need another entry for `checks_timeout`.

jd (changes requested):

releasenotes/notes/queue-merge-started-at-b8efc3b8ecf348f5.yaml
sileht: It already exists, in the other file.

jd: arf why you do 2 files omg 😁

jd (approved):

PR #3763, opened 2021-12-10
fix(approved-by): should not contains requested_reviews

sileht:
GitHub Reviews API still show as APPROVED pull request with
requested_reviews.

From GitHub UI and GitHub branch protection perspective, the pull
request is not approved anymore.

This change improved approved-by to done the same.

Fixes MRGFY-773

Change-Id: Id42991a441b53151eedeecc0c4c0dd9f291e582e

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -975,6 +975,10 @@ class Context(object):
             approvals: typing.Dict[
                 github_types.GitHubLogin, github_types.GitHubReview
             ] = {}
+
+            requested_reviewer_ids = [
+                rr["id"] for rr in self.pull["requested_reviewers"]
+            ]
             valid_user_ids = {
                 r["user"]["id"]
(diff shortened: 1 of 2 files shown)


mergify_engine/context.py
```diff
@@ -975,6 +975,10 @@ async def consolidated_reviews(
             approvals: typing.Dict[
                 github_types.GitHubLogin, github_types.GitHubReview
             ] = {}
+
+            requested_reviewer_ids = [
+                rr["id"] for rr in self.pull["requested_reviewers"]
+            ]
```
jd, line 981:
```suggestion
            requested_reviewer_ids = {
                rr["id"] for rr in self.pull["requested_reviewers"]
            }
```

jd (changes requested):

sileht:
This is not the right solution.

PR #3770, opened 2021-12-13
fix(delayed-refresh): don't fail to refresh on team resolution failure

sileht:
Fixes MERGIFY-ENGINE-2GJ

Change-Id: Ib05e9590c08b55519486ea567a8b2e2437dd7816

### DIFF
--- a/mergify_engine/delayed_refresh.py
+++ b/mergify_engine/delayed_refresh.py
@@ -53,7 +53,10 @@ async def plan_next_refresh(
     for rule in _rules:
         f = filter.NearDatetimeFilter(rule.conditions.extract_raw_filter_tree())
         live_resolvers.configure_filter(ctxt.repository, f)
-        bet = await f(pull_request)
+        try:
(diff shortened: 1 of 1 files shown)


mergify_engine/delayed_refresh.py
```diff
@@ -53,7 +53,10 @@ async def plan_next_refresh(
     for rule in _rules:
         f = filter.NearDatetimeFilter(rule.conditions.extract_raw_filter_tree())
         live_resolvers.configure_filter(ctxt.repository, f)
-        bet = await f(pull_request)
+        try:
+            bet = await f(pull_request)
```
jd, line 57: If I read correctly `bet` can be undefined.

sileht: Hum, we don't use `bet` if the exception is raised or outside the loop, so we are good.

jd: right!

jd (approved):

PR #3776, opened 2021-12-14
fix(queue): requeue when priority changes

sileht:
Fixes MRGFY-756

Change-Id: Ibe2dd0d5f00a3b7ce63bad7f9fca5507e2d3aac2

### DIFF
--- a/mergify_engine/queue/merge_train.py
+++ b/mergify_engine/queue/merge_train.py
@@ -1190,9 +1190,23 @@ class Train(queue.QueueBase):
         await self.force_remove_pull(ctxt, exclude_ref=ctxt.pull["base"]["ref"])
 
         best_position = -1
+        need_to_be_readded = False
         for position, (embarked_pull, _) in enumerate(self._iter_embarked_pulls()):
             if embarked_pull.user_pull_request_number == ctxt.pull["number"]:
-                # already in queue, we are good
+                if (
+                    config["effective_priority"]
+                    != embarked_pull.config["effective_priority"]
+                    or config["name"] != embarked_pull.config["name"]
+                ):
+                    self.log.info(
+                        "pull request already in train but misplaced",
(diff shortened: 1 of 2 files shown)


mergify_engine/queue/merge_train.py
```diff
@@ -1190,9 +1190,23 @@ async def add_pull(
...
         for position, (embarked_pull, _) in enumerate(self._iter_embarked_pulls()):
             if embarked_pull.user_pull_request_number == ctxt.pull["number"]:
-                # already in queue, we are good
+                if (
+                    config["effective_priority"]
+                    != embarked_pull.config["effective_priority"]
+                    or config["name"] != embarked_pull.config["name"]
+                ):
+                    self.log.info(
+                        "pull request already in train, but misplaced",
```
jd, line 1202:
```suggestion
                        "pull request already in train but misplaced",
```

mergify_engine/queue/merge_train.py
```diff
@@ -1205,9 +1219,16 @@ async def add_pull(
...
             ):
-                # We found a car with lower priorit
+                # We found a car with lower priority
                 best_position = position
 
+        if need_to_be_readded:
+            # FIXME(sileht): this can be optimised by not dropping spec checks,
+            # if the position in the queue does not change
+            await self.remove_pull(ctxt)
+            await self.add_pull(ctxt, config)
```
jd, line 1229: Is it worth having a `move_pull` which might be more Redis friendly (e.g. one operation or pipe)

sileht: From Redis point of view, it would be 1  `hset()` instead 2. But this doesn't only move things in Redis, it may also clean up some resources on GitHub because some TrainCar has been deleted.

jd (approved):

PR #3778, opened 2021-12-14
feat(queue): allow to never interrupt speculative checks

sileht:
Fixes MRGFY-772

Change-Id: I20b319747a2858709ea9dad54cbbf44df233be5a

### DIFF
--- a/docs/source/actions/queue.rst
+++ b/docs/source/actions/queue.rst
@@ -348,6 +348,13 @@ A ``queue_rules`` takes the following parameter:
      - True
      - Allow to update/rebase the original pull request if possible to check its mergeability.
        If ``False``, a draft pull request is always created.
+   * - ``allow_speculative_checks_interruption``
+     - bool
+     - True
+     - Allow interrupting the ongoing speculative checks when a pull request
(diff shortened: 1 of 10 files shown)


docs/source/actions/queue.rst
```diff
@@ -348,6 +348,13 @@ A ``queue_rules`` takes the following parameter:
      - True
      - Allow to update/rebase the original pull request if possible to check its mergeability.
        If ``False``, a draft pull request is always created.
+   * - ``allow_speculative_checks_interruption``
+     - bool
+     - True
+     - Allow to interrupt the ongoing speculative checks when a pull request
+       with higher priority enter in queue.
+       If ``False``, pull request with higher priority will be inserted just after the pull requests that
+       have speculative checks running.
```
jd, line 357:
```suggestion
     - Allow interrupting the ongoing speculative checks when a pull request
       with higher priority enters in the queue.
       If ``False``, pull request with higher priority will be inserted just after the pull requests that
       have speculative checks running.
```

jd (changes requested):

jd (approved):

PR #3784, opened 2021-12-15
feat(queue): wait for pull request rules in in-place mode

sileht:
By default, it's safer to wait for pull request rules to match again
after the rebase, before checking if the queue rules match or not.

Fixes MRGFY-770

Change-Id: Ib0a313664d97f567c6146b1498eea077f9fbd75b


mergify_engine/actions/queue.py
```diff
@@ -304,11 +304,29 @@ async def _should_be_queued(self, ctxt: context.Context, q: queue.QueueT) -> boo
...
+            )
+            if pull_rule_checks_status == check_api.Conclusion.FAILURE:
+                return False
+
+            # NOTE(sileht): if the pull request rules are pending we wait their
+            # match before checking queue rules states, in case of one
+            # condition suddently unqueue the pull request.
+            # TODO(sileht): we may want to make this behavior configurable as
+            # people having slow/long CI running only on pull request rules, we
+            # may want to merge it before it finishes.
```
jd, line 326:
```suggestion
            # NOTE(sileht): if the pull request rules are pending we wait for their
            # matches before checking queue rules states, in case of one
            # condition suddently unqueueing the pull request.
            # TODO(sileht): we may want to make this behavior configurable as
            # people having slow/long CI running only on pull request rules, we
            # may want to merge it before it finishes.
```

jd (approved):

jd:
@sileht Does this fix the all MRGFY-770 story?

sileht:
> @sileht Does this fix the all [MRGFY-770](https://linear.app/mergify/issue/MRGFY-770) story?

yes

PR #3790, opened 2021-12-15
fix(queue): enable new get_rule_checks_status()

sileht:
We still log the old one for a while just in case.

Fixes MRGFY-729
Fixes MRGFY-730
Fixes MRGFY-731
Fixes MRGFY-736
Fixes MRGFY-752

Change-Id: I79116d109640be813b99abe30669777745ea77c5


sileht:
@mergifyio rebase

jd (approved):

PR #3797, opened 2021-12-17
chore(queue): rename some queue_rules options

sileht:
allow_speculative_checks_interruption -> allow_checks_interruption
allow_inplace_speculative_checks -> allow_inplace_checks

Fixes MRGFY-789
Fixes MRGFY-790

Change-Id: If2757260987ad5c50ae7cd87d57783da5c912e7e


jd (dismissed):
1. doc update?
2. is upgrade seamless for GHES?

sileht:
> 2. is upgrade seamless for GHES?

Yes the old names still works.

jd (approved):

PR #3798, opened 2021-12-17
fix(ghes): make graphql call working

sileht:
The graphql endpoint of GHES is /api/graphql instead of /graphql.

This introduce new environment variables to configure it.

Fixes MRGFY-788

Change-Id: I084cfe2631180fabd4609a26effe050c3fda5f4a


jd (changes requested):
Release note

jd (approved):

PR #3811, opened 2021-12-21
fix(review): ensure we don't review the same PR twice

sileht:
Fixes MRGFY-792

Change-Id: I31b645cf7b1453e45948f5f5e6ace0de65807715


jd (dismissed):

mergify_engine/actions/review.py
```diff
@@ -102,25 +102,15 @@ async def run(
...
-            reversed(
-                list(
-                    filter(
-                        lambda r: r["user"]["login"] is not review_user,
-                        await ctxt.reviews,
-                    )
+        reviews = reversed(
+            list(
+                filter(
+                    lambda r: r["user"]["login"] == review_user,
```
jd, line 108: did you mean `==` or `!=`?

sileht: `==` I have rewrite the test to prove it ;) and even check the new test fail with the old code.

jd (approved):

PR #3818, opened 2021-12-22
refactor(worker): Don't use login/repo from Redis key for tracing/logging

sileht:
This change will allow dropping repo_name and owner_login from Redis
keys without losing logging/tracing information.

Depends-On: #3821

Change-Id: I84189a340e6f458adad09bafade40f6c8ed06717

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -56,6 +56,7 @@ import typing
 import daiquiri
 from datadog import statsd
 from ddtrace import tracer
+import first
 import msgpack
 import tenacity
 import yaaredis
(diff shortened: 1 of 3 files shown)


mergify_engine/worker.py
```diff
@@ -555,15 +570,19 @@ async def _consume_buckets(
             pulls_processed += 1
             installation.client.set_requests_ratio(pulls_processed)
 
+            messages = await self.redis_stream.xrange(bucket_sources_key)
+            statsd.histogram("engine.buckets.events.read_size", len(messages))  # type: ignore[no-untyped-call]
+
+            tracing_repo_name = f"<unknown {repo_id}>"
+            if messages:
+                tracing_repo_name = messages[0].get(b"repo_name", tracing_repo_name)
```
jd, line 578: use `first` to find the first message with a `repo_name`?

mergify_engine/worker.py
```diff
@@ -323,6 +324,17 @@ class StreamProcessor:
     redis_cache: utils.RedisCache
     dedicated: bool
 
+    owners_mapping: typing.Dict[
+        github_types.GitHubAccountIdType, github_types.GitHubLogin
+    ] = dataclasses.field(default_factory=dict)
```
jd, line 329: just wondering if this is not going to explode in memory size? (at least in theory)
should we leverage Redis?

sileht: I hope to see it explode!

I don't think we should add redis yet for this.

This is just a dict of int/string of Mergify installation, we currently have 4000 installations, so if all org names are ~ 50 chars long, the dict size will be 4000 * 32 (int) +  4000 * 100 (char) + 10% (dict overhead), so less than 1Mo.

I think we can afford this in-memory cache. I can put a comment just in case to not forget it.

jd (approved):

PR #3821, opened 2021-12-22
fix(redis): load redis script on startup

sileht:
Since we use pipeline with redis script, if script loading may fail as
the ScriptError is not raise when we use evalsha(), but a generic error
is raise during the pipeline execution.

Instead, this change loads script on startup.

Depends-On: #3821

Change-Id: Id47d234cbb0447d94b4d54c3fc6268122dfae1f9

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -71,6 +71,7 @@ from mergify_engine import github_events
 from mergify_engine import github_types
 from mergify_engine import logs
 from mergify_engine import migrations
+from mergify_engine import redis_utils
 from mergify_engine import service
 from mergify_engine import signals
 from mergify_engine import utils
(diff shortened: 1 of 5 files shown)


mergify_engine/worker.py
```diff
@@ -1041,6 +1042,7 @@ async def start(self) -> None:
         self._redis_stream = utils.create_yaaredis_for_stream()
         self._redis_cache = utils.create_yaaredis_for_cache()
 
+        await redis_utils.load_scripts(self._redis_stream)
```
jd, line 1045: What about doing this in `create_yaaredis_for_stream`?

sileht: I wanted to avoid running that il all utilities `mergify-debug/mergify-refresher/mergify-worker-status/...`

jd (approved):

PR #3822, opened 2021-12-22
fix(depends-on): support when body has CRLF instead LF eol

sileht:
GitHub has changed the eol character in the UI, yes you correctly read it!

Body edited with UI has \r\n. Via API it have exactly was you sent.

Looks at body on old PR #3237, then on new PR #3821.

Since this breaks our multiline regex assumptions, this change replaces
all \r\n to \n.

Change-Id: I0b8bbc4d2acbb215626cc03d58fdf19deaa3554c


jd (commented):
OK

jd (approved):

PR #3823, opened 2021-12-22
chore(github-events): remove useless cache

sileht:
Since we always get the list of open pull requests, storing
the mapping between sha and number is useless.

Change-Id: Ie004f4007cd37d665e08e7897b8b57dc0fde2d57

### DIFF
--- a/mergify_engine/github_events.py
+++ b/mergify_engine/github_events.py
@@ -464,29 +464,18 @@ async def filter_and_dispatch(
 SHA_EXPIRATION = 60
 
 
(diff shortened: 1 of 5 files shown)


jd (commented):
Is there a case where you'd need the sha/number mapping of a PR that was (just) closed?

sileht:
> Is there a case where you'd need the sha/number mapping of a PR that was (just) closed?

We need this mapping for status and check-runs. And I have another plan for them ;)

jd (approved):

PR #3830, opened 2021-12-23
chore: no need to process check_suite not from mergify

sileht:
Change-Id: Id7c349debd94a5f1efdc59c05479b310817c68a7

### DIFF
--- a/mergify_engine/github_events.py
+++ b/mergify_engine/github_events.py
@@ -263,12 +263,11 @@ async def push_to_worker(
         if event["repository"]["archived"]:
             ignore_reason = "repository archived"
 
-        elif event["action"] != "rerequested":
-            ignore_reason = f"check_suite/{event['action']}"
+        elif event[event_type]["app"]["id"] != config.INTEGRATION_ID:
+            ignore_reason = "not a mergify check_suite"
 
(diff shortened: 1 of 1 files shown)


mergify_engine/github_events.py
```diff
@@ -263,12 +263,11 @@ async def push_to_worker(
         if event["repository"]["archived"]:
             ignore_reason = "repository archived"
 
-        elif event["action"] != "rerequested":
-            ignore_reason = f"check_suite/{event['action']}"
+        elif event[event_type]["app"]["id"] != config.INTEGRATION_ID:
+            ignore_reason = "not a mergify check_suite"
```
jd, line 267: I'm lacking context as to why we only process check suite from Mergify here?

sileht: We use the check-suite/rerequested event to run the engine when people push the GitHub UI re-run button.

jd: I'd thought we wanted to also be triggered on `check_suite` `completed` event just in case?

sileht: Hum, ok, that looks legit seatbelt.

PR #3851, opened 2021-12-29
ci(releasenotes): lint reno files

sileht:
Change-Id: I4a1131cc4c1cef5d591b7c7aac2bf7c1119a5028

### DIFF
--- a/.github/workflows/ci-pep8.yaml
+++ b/.github/workflows/ci-pep8.yaml
@@ -14,6 +14,8 @@ jobs:
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v2.4.0
+        with:
+          fetch-depth: 0
 
       - name: Setup Python 🔧
         uses: actions/setup-python@v2.3.1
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-pep8.yaml
```diff
@@ -14,6 +14,8 @@ jobs:
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v2.4.0
+        with:
+          fetch-depth: 0
```
jd, line 18: sounds unrelated to the pr?

jd (changes requested):
might be good splitting it

.github/workflows/ci-pep8.yaml
sileht: Reno needs full git clone to work

jd: ok, would have been great to write this in the commit message :)

jd (approved):

PR #3861, opened 2022-01-02
fix(worker): set owner_login_for_tracing

sileht:


jd (approved):
No sentry nor test? :(

PR #3885, opened 2022-01-06
fix(queue): statuses information may be None

sileht:
GitHub may returns None for these attributes

We previously displayed "None" in the checks list.

The new code raises an error.

Fixes MERGIFY-ENGINE-2H3


jd (approved):

sileht:
@mergifyio refresh

PR #3887, opened 2022-01-06
fix(queue): control is_behind one last time before merging a pull request

sileht:
After we rebase pull requests in-place if something gets merged manually
at a particular timing, the queue can be stuck has Mergify didn't check
one last time before merging the pull request that the pull request is
up-to-date, it blindly trusts the QUEUE_SUMMARY generated for the previous
base sha.

This change adds new control of is_behind, and the pull request will
be rebased one more time.

Fixes MRGFY-822

### DIFF
--- a/mergify_engine/actions/queue.py
+++ b/mergify_engine/actions/queue.py
@@ -306,7 +306,11 @@ Then, re-embark the pull request into the merge queue by posting the comment
         if not await q.is_first_pull(ctxt):
             return False
 
-        if not await ctxt.is_behind:
+        if await ctxt.is_behind:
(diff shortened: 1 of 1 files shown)


mergify_engine/actions/queue.py
```diff
@@ -306,7 +306,11 @@ async def _should_be_merged(
         if not await q.is_first_pull(ctxt):
             return False
 
-        if not await ctxt.is_behind:
+        if await ctxt.is_behind:
```
jd, line 309: I guess this is cached, no? Are you sure it's up to date when checking here?

sileht: Yes it's up to date.

jd (approved):

PR #3918, opened 2022-01-12
/bin/sh: line 1: qa: command not found

sileht:
This change:
* improves the error message
* set the state to pending to retry the action automatically
* set a refresh date to ensure the action will rerun when the issue is
  resolved.

Change-Id: I7b4e31e6127d775febaf79ad47a23c1595805dba


jd (approved):
great idea 👍

PR #3921, opened 2022-01-12
fix(queue): handle correctly batch when speculative_check=1

sileht:
Fixes MRGFY-844

Change-Id: I1c0165a19eccdfa5c516e3b7d99d37f968a2746f


sileht:
I will add tests

mergify_engine/engine/queue_runner.py
```diff
@@ -26,29 +26,11 @@
...
-            "train car has unexpectedly been synchronized",
-        )
-        return True
-
     unexpected_event = first(
         (source for source in ctxt.sources),
         key=lambda s: s["event_type"] == "pull_request"
         and typing.cast(github_types.GitHubEventPullRequest, s["data"])["action"]
-        in ["closed", "reopened"],
+        in ["closed", "reopened", "synchronize"],
```
jd, line 33: I think that's part of another PR, right?

sileht: yes, this is required to make that works.

mergify_engine/queue/merge_train.py
```diff
@@ -1442,10 +1443,16 @@ async def _split_failed_batches(self, queue_rules: rules.QueueRules) -> None:
                 # We move this car later at the end to not retest it
                 del self._cars[-1]
 
+                # NOTE(sileht): if speculative_checks == 1 we split the batch
+                # in two parts, but check only the first one
+                if queue_rule.config["speculative_checks"] == 1:
+                    parts = 2
+                else:
+                    parts = queue_rule.config["speculative_checks"]
```
jd, line 1451:
```suggestion
                parts = min(2, queue_rule.config["speculative_checks"])
```

mergify_engine/queue/merge_train.py
```diff
@@ -1460,17 +1467,22 @@ async def _split_failed_batches(self, queue_rules: rules.QueueRules) -> None:
...
-                    ):
-                        self.log.info(
-                            "failed to create draft pull request",
-                            car=car,
-                            exc_info=True,
-                        )
+                    # NOTE(sileht): if speculative_checks == 1 we must checks
+                    # only the first cars, keep the second one as pending.
+                    # _populate_cars() will create the second one, when the
+                    # first car has finished and passed
```
jd, line 1473:
```suggestion
                    # NOTE(sileht): if speculative_checks == 1 we must check
                    # only the first car, keep the second one as pending.
                    # _populate_cars() will create the second one when the
                    # first car has finished and passed
```

mergify_engine/queue/merge_train.py
```diff
@@ -1487,6 +1499,38 @@ async def _split_failed_batches(self, queue_rules: rules.QueueRules) -> None:
                 )
                 break
 
+        # NOTE(sileht): speculative_checks=1 may create car without the
+        # attached draft pull request if this car become the first, it means
+        # the previous car has been merged and we can start testing it
```
jd, line 1504: if spec check is 1, batch size is 2, and in place is allowed, is there a chance that there's no draft PR at all?
Does this work? Tested?

mergify_engine/tests/unit/test_train.py
```diff
@@ -80,6 +81,11 @@ def monkepatched_traincar(monkeypatch: pytest.MonkeyPatch) -> None:
 
 MERGIFY_CONFIG = """
 queue_rules:
+  - name: 1x5
```
jd, line 84: maybe test 1x2 with in-place like I wrote above?

jd (dismissed):

mergify_engine/queue/merge_train.py
sileht: yes it works, but I will add a test to prove it!

`TrainCar._create_pull()` is badly named since we introduce in-place, it should be `TrainCar._setup_car()` or `TrainCar._start_checking_car()`.

I will create a ticket for that.

jd (approved):

PR #3930, opened 2022-01-14
fix(config): change the way configuration change

sileht:
Using pull.base.sha is unreliable as it's the sha of when the pull
request is opened and not the real merge-base/fork-point of the pull
request.

This change looks at the diff between the base branch and the merge
commit of the pull request.

Fixes MRGFY-829

Change-Id: Ic31a7df439eca450148233d363cc98495f782109


mergify_engine/actions/dismiss_reviews.py
```diff
@@ -58,9 +58,7 @@ class DismissReviewsAction(actions.Action):
             [types.GitHubLogin],
             FROM_REQUESTED_REVIEWERS,
         ),
-        voluptuous.Required("message", default=None): voluptuous.Any(
-            None, types.Jinja2
-        ),
+        voluptuous.Required("message", default=None): types.Jinja2WithNone,
```
jd, line 61: unrelated :(

mergify_engine/tests/functional/test_configuration.py
```diff
@@ -92,6 +92,54 @@ async def test_invalid_yaml_configuration_in_repository(self):
             }
         ]
 
+    async def test_no_configuration_changed_with_wierd_base_sha(self):
```
jd, line 95:
```suggestion
    async def test_no_configuration_changed_with_weird_base_sha(self):
```

jd (dismissed):

mergify_engine/actions/dismiss_reviews.py
sileht: I prepared the record with the other change :)

jd (approved):

PR #3932, opened 2022-01-14
fix(rules): RuleConditionGroup description should be copied

sileht:
Change-Id: I3502b057443b255ed415f1218655f6e65d270cd4


jd (approved):
test?

PR #3976, opened 2022-01-25
feat(command): allow to restrict where command can be ran

sileht:
This change leverages the conditions system to configure on which pull
request commands can be ran or not.

Fixes MRGFY-859

Change-Id: Ia6d82e1fe789ce1c3cb0e7107e4a8b511a3ef205


jd (changes requested):
Documentation, relnote?

sileht:
image

sileht:
> Documentation, relnote?

Wanted to confirm the approach first.

jd:
> > Documentation, relnote?
> 
> Wanted to confirm the approach first.

I didn't read the code. Rather write the doc first 😅

docs/source/configuration.rst
```diff
@@ -75,6 +75,39 @@ Queue Rules
...
+
+   * - Key Name
+     - Value Type
+     - Value Description
+   * - ``conditions``
+     - list of :ref:`Conditions`
+     - A list of :ref:`Conditions` string that must match against the pull
+       request for the command to be allowed.
+
+For example, to limit backport command for pull request targeting the main branch:
```
jd, line 101: This is unclear. To me, the `conditions` match on the source pull request. They cannot match on a destination pull request since it does not exist yet 🤔 
Reading the code, this is what it actually does IIRC, so I think this explanation is just unclear.

```suggestion
For example, to limit backport commands for pull requests coming from the main branch:
```

sileht:
@merigfyio rebase

sileht:
@Mergifyio rebase

jd (dismissed):

mergify_engine/rules/__init__.py
```diff
@@ -555,6 +555,14 @@ def FullifyPullRequestRules(v):
     return v
 
 
+CommandsRestrictionsSchema = {
+    voluptuous.Required("conditions", default=[]): voluptuous.All(
```
jd, line 559: Do we allow `None` to allow `conditions:`?

jd (approved):

mergify_engine/rules/__init__.py
sileht: Hum, if it's empty the whole section is useless, so.

PR #3985, opened 2022-01-27
fix(worker): dd_tracer resource name

sileht:
As we use a new StreamProcessor each worker task we need
to store accounts mapping in Worker itself.

Fixes MRGFY-881

Change-Id: I203069146556274848b444af976faf93be289763

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -325,25 +325,36 @@ class SharedOrgBucketSelector:
 
 
 @dataclasses.dataclass
-class StreamProcessor:
-    redis_stream: utils.RedisStream
-    redis_cache: utils.RedisCache
-    dedicated: bool
-
+class OwnerLoginsCache:
     # NOTE(sileht): This could take some memory in the future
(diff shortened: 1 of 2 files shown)


mergify_engine/worker.py
```diff
@@ -325,26 +325,37 @@ async def next_org_bucket(self) -> typing.Optional[worker_lua.BucketOrgKeyType]:
...
-
+class OwnerLoginsCache:
     # NOTE(sileht): This could take some memory in the future
     # we can assume github_types.GitHubLogin is ~ 104 bytes
     # and github_types.GitHubAccountIdType 32 bytes
     # and 10% dict overhead
-    owners_mapping: typing.Dict[
+    _mapping: typing.Dict[
         github_types.GitHubAccountIdType, github_types.GitHubLogin
     ] = dataclasses.field(default_factory=dict)
```
jd, line 335: add `repr=False`?

jd (dismissed):

jd (approved):

PR #4009, opened 2022-01-31
refactor(worker): allow to separate dedicated worker

sileht:
This will allow spawning dedicated tasks and shared tasks in different workers/processes.

Fixes MRGFY-887

Change-Id: I855cec767da4bcbe93b02635c0d83327eb14c755


jd (changes requested):
Would be great to have more description as of why in the PR body 😢

jd (approved):

PR #4026, opened 2022-02-03
fix(test): remove arguments from pytest requests

Syffe:
removing "--pyargs mergify_engine" in tox.ini pytest requests

Fixes MRGFY-895


Syffe:
Changed the desc, didn't saw the template at first sight ! Thanks !

jd (approved):
It'd be more interesting to write *why* you're doing this in the desc, rather than what you do (that we can read in the diff 😉 )

PR #4029, opened 2022-02-04
feat(queue): allow to set a bot account for draft pull requests

sileht:
CIs like CircleCI may restrict who can run tests. And does not support
GitHub user of type "Bot". So we should allow to use a custom account
for that.

Fixes MRGFY-878

Change-Id: Ie909808a00ee473ba8513c826ee697226936e84b


docs/source/actions/queue.rst
```diff
@@ -375,6 +375,16 @@ A ``queue_rules`` takes the following parameter:
      - The amount of time Mergify waits for pending checks to return before unqueueing pull requests.
        This cannot be less than 60 seconds.
 
+   * - ``draft_bot_account``
+     - string
+     -
+     - |premium plan tag|
+       Mergify can impersonate a GitHub user to create draft pull request.
```
jd, line 382:
```suggestion
       Mergify can impersonate a GitHub user to create draft pull requests.
```

docs/source/actions/queue.rst
```diff
@@ -375,6 +375,16 @@ A ``queue_rules`` takes the following parameter:
...
+       If no ``draft_bot_account`` is set, Mergify will create the draft pull request
+       itself. The user account **must** have already been
+       logged in Mergify dashboard once and have **write** or **maintain** permission.
```
jd, line 385:
```suggestion
       If no ``draft_bot_account`` is set, Mergify creates the draft pull request
       itself. The user account **must** have already been
       logged in Mergify dashboard once and have **write** or **maintain** permission.
```

or admin permission?

mergify_engine/queue/merge_train.py
```diff
@@ -451,6 +452,18 @@ async def create_pull(
...
         self.creation_state = "created"
 
+        bot_account = queue_rule.config["draft_bot_account"]
+        github_user: typing.Optional[user_tokens.UserTokensUser] = None
+        if bot_account:
+            tokens = await self.train.repository.installation.get_user_tokens()
+            github_user = tokens.get_token_for(bot_account)
+            if not github_user:
+                await self._set_creation_failure(
+                    f"Unable to comment: user `{bot_account}` is unknown. "
```
jd, line 462:
```suggestion
                    f"Unable to create pull request: user `{bot_account}` is unknown. "
```

jd (changes requested):

mergify_engine/rules/__init__.py
```diff
@@ -521,6 +522,9 @@ def ChecksTimeout(v: str) -> datetime.timedelta:
                 voluptuous.Required("checks_timeout", default=None): voluptuous.Any(
                     None, voluptuous.All(str, voluptuous.Coerce(ChecksTimeout))
                 ),
+                voluptuous.Required("draft_bot_account", default=None): voluptuous.Any(
+                    None, str
```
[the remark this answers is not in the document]
sileht: It can't because the template needs a `context.Context`, we can't pick the draft one as it has not been created yet. And we cannot pick the pull requests that compose the TrainCar because we need only one `context.Context` and we can't random pick one pull request.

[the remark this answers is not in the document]
sileht: The template for `bot_account` was useful only to use `{{ author }}`. This doesn't make sense to pick the author of one of the original pull request and this will confuse a lot people to see merge-queue PR opened with names.

jd (approved):

PR #4030, opened 2022-02-04
fix(api): return 403 error when github rate limit exception raised

Syffe:
The Github rate limit exception was raised by the engine but unhandled by the REST API thus
sending back a code 500. This fix now handles the exception and sends back a 403
code and an error message. 

Fixes MRGFY-882

### DIFF
--- a/mergify_engine/web/config_validator.py
+++ b/mergify_engine/web/config_validator.py
@@ -19,10 +19,12 @@ import base64
 import hashlib
 
 import fastapi
+from starlette import requests
 from starlette import responses
 from starlette.middleware import cors
 
 from mergify_engine import context
+from mergify_engine import exceptions as engine_exceptions
 from mergify_engine import github_types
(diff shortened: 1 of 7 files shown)


mergify_engine/web/config_validator.py
```diff
@@ -70,3 +72,13 @@ async def config_validator(
         message = "The configuration is valid"
 
     return responses.PlainTextResponse(message, status_code=status)
+
+
+@app.exception_handler(engine_exceptions.RateLimited)
```
jd, line 77: do we have to repeat this in all blueprints?

sileht: This is not a blueprint (router in Scarlette/FastAPI wording), but a full app.

We have multiples apps instead of multiple routers to not set up CORS on all endpoints.

sileht (approved):

PR #4048, opened 2022-02-08
fix(command): improvement on command loader trigger regex

Syffe:
The regex was triggering the command loader for every comment
that contained the "@mergify/io" tag, regardless of its position in the comment.
Now it is only triggered if the tag is placed at the start of the comment.

Fixes MRGFY-797

### DIFF
--- a/mergify_engine/engine/commands_runner.py
+++ b/mergify_engine/engine/commands_runner.py
@@ -37,7 +37,7 @@ from mergify_engine.rules import conditions
 
 LOG = daiquiri.getLogger(__name__)
 
-COMMAND_MATCHER = re.compile(r"@Mergify(?:|io) (\w*)(.*)", re.IGNORECASE)
+COMMAND_MATCHER = re.compile(r"^@Mergify(?:|io) (\w*)(.*)", re.IGNORECASE)
 COMMAND_RESULT_MATCHER_OLD = re.compile(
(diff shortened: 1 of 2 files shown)


mergify_engine/engine/commands_runner.py
```diff
@@ -83,14 +86,23 @@ class CommandInvalid(Exception):
...
     mergify_config: rules.MergifyConfig,
     message: str,
 ) -> Command:
     """Load an action from a message."""
     action_classes = actions.get_commands()
     match = COMMAND_MATCHER.search(message)
-    if match and match[1] in action_classes:
+
+    if not match:
+        raise NotACommand(COMMENT_IS_NOT_A_COMMAND)
```
jd, line 103: You can probably just embed the str here, no need to have a variable IMHO. We're not going to reuse it and it makes the code a little bit harder to follow.

mergify_engine/engine/commands_runner.py
```diff
@@ -323,6 +335,9 @@ def log(comment_out: str, result: typing.Optional[check_api.Result] = None) -> N
         log(e.message)
         await post_comment(ctxt, e.message + footer)
         return
+    except NotACommand as e:
+        log(e.message)
```
jd, line 339: I don't think we need to log this — it'll just make noise in production log.

jd (changes requested):
nitpick otherwise LGTM

PR #4050, opened 2022-02-08
Fallback to .github repo for each config file

mkurz:
See [this discussion](https://github.com/Mergifyio/mergify-engine/discussions/2205).

If a config file can't be found in the "local" repo, we try to fall back to the `.github` repo.
That is (almost) the same like what Probot does: Look at [their `getConfigFiles` method](https://github.com/probot/octokit-plugin-config/blob/main/src/util/get-config-files.ts#L24-L104):
```js
/**
 * Load configuration from selected repository file. If the file does not exist
 * it loads configuration from the owners `.github` repository.
 *
 * If the repository file configuration includes an `_extends` key, that file
 * is loaded. Same with the target file until no `_extends` key is present.
 * ...
 */
```

"Almost" because I did not implement the `_extends` mechanism in this pull request. Just look at their code, it's easy to understand. IMHO mergify should behave the same.
However this pull request can be merged nonetheless, we can make `_extends` work later.

--

I also want to give you an example how `_extends` work:
Look at followingextends [release drafter](https://github.com/release-drafter/release-drafter/) config files, all are placed in the `.github` repo of our organization:
https://github.com/playframework/.github/tree/main/.github/release-drafts
If you open the configs you will see all extend from the base config, see https://github.com/playframework/.github/blob/main/.github/release-drafts/increasing-major-version.yml#L4

Now in a workflow we [tell release drafter to use the config file `release-drafts/increasing-minor-version.yml`](https://github.com/playframework/play-json/blob/main/.github/workflows/release-drafter.yml#L15). First probot looks into the "local" repo, but in the `play-json` repo, the file `.github/release-drafts/increasing-minor-version.yml` does not exist: https://github.com/playframework/play-json/tree/main/.github Therefore it falls back to the `.github` repo and loads [this file](https://github.com/playframework/.github/blob/main/.github/release-drafts/increasing-minor-version.yml) (and also resolved the `_extends` property).

However if a "local" file can be found, it will use this one first e..g here: https://github.com/playframework/play-soap/blob/main/.github/workflows/release-drafter.yml#L15, we tell release drafter to use `release-drafts/increasing-patch-version.yml` which **does** exist in this repo: https://github.com/playframework/play-soap/tree/main/.github/release-drafts
Now it is using this one instead of falling back to the `.github` repo. But it is getting better: The "local" config file can now still extend the config file with the same name located in the `.github` repo: https://github.com/playframework/play-soap/blob/main/.github/release-drafts/increasing-patch-version.yml#L1

This way it is possible to organize config files at a central place and just override the configs you want to change for a specifc repo.


mmlb:
oh nice!

mkurz:
@mmlb It would be even nicer if the `_extends` mechanism works. Actually maybe the code I referenced from Probo can be used as template to implement it here as well?
However this pull request on its own can be merged meanwhile.

jd (commented):
Thanks a lot! That was indeed in our roadmap.
This would need a few more things to be complete, at least documentation and test.
Let us know if you want to care of those, or we'll follow up and might take over this PR. 😄

mkurz:
> Let us know if you want to care of those, or we'll follow up and might take over this PR. smile

If you could take over that would be really nice. I am low on time and not really into Python.
This feature would be really really nice to have, also it's [the most upvoted discussion](https://github.com/Mergifyio/mergify-engine/discussions/2205). Also if you could implement `_extends` that would be ace. Again, look at Probot's code, it's basically just merging config files like merging a json object.
Looking forward to this feature! Organisations will love it :+1:

mkurz:
Conflict resolved!

jd:
This will need more testing but we'll have it on our roadmap. Closing for now :)

mkurz:
Thanks @jd, looking forward to this feature!

mkurz:
@jd Any progress regarding this feature (also see https://github.com/Mergifyio/mergify/discussions/2205) ?

PR #4055, opened 2022-02-09
feat(queue): track when checks end

sileht:
This allow to track when checks of a TrainCar are finished and Mergify just wait for
all GitHub round-strip to merge all PRs of the TrainCar.

Fixes MRGFY-909

Change-Id: Idd2770baa596dd374a14a176aba5270a1b4f9824


mergify_engine/queue/merge_train.py
```diff
@@ -182,6 +182,7 @@ class TrainCar:
     last_checks: typing.List[QueueCheck] = dataclasses.field(default_factory=list)
     last_evaluated_conditions: typing.Optional[str] = None
     has_timed_out: bool = False
+    checks_ended_date: typing.Optional[datetime.datetime] = None
```
jd, line 185:
```suggestion
    checks_ended_time: typing.Optional[datetime.datetime] = None
```
or `timestamp`?

jd (changes requested):
tagging as -1 since i did not get an answer

jd (approved):

PR #4060, opened 2022-02-10
ci(tox): ensure tox don't fallback when python version is wrong

sileht:
Change-Id: I1e5307d7ec25d7c4b4194740d3c5c8fee6e1a474

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,4 +1,6 @@
 [tox]
+# this disables the fallback from non existing `pyXYZ` target to make sure we use always use basepython
+ignore_basepython_conflict = true
 envlist = py310,pep8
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -1,4 +1,5 @@
 [tox]
+ignore_basepython_conflict = true
```
jd, line 2: can you add a comment?
```suggestion
# this disables the fallback from non existing `pyXYZ` target to make sure we use always use basepython
ignore_basepython_conflict = true
```

jd (dismissed):

sileht:
@mergifyio squash

jd (approved):

PR #4067, opened 2022-02-11
ci(updatecli): Add tools to automatically update python version

sileht:
Change-Id: I20fdfac958766cd8b92824ebf2ae6bf598166cf3

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -1,0 +1,57 @@
+name: Updatecli
+permissions: write-all
+
+on:
+  workflow_dispatch:
+  # FIXME(sileht): enable the schedule once
+  # https://github.com/updatecli/updatecli/issues/522 is fixed
+  # schedule:
+  #   - cron: '0 0 * * *'
+
+jobs:
+  updatecli:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      # NOTE(sileht): We don't use the GitHub action as it uses docker:
+      # * some files are missing inside the docker container of updatecli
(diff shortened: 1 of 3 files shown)


.github/workflows/updatecli.yaml
```diff
@@ -0,0 +1,51 @@
...
+        uses: actions/checkout@v2
+
+      # NOTE(sileht): We don't use the GitHub action as it uses docker:
+      # * some files are missing inside the docker container of updatecli
+      # * python is not installed in this container
+      - name: Updatecli
+        id: updatecli
+        run: |
+          url="https://github.com/updatecli/updatecli/releases/download/v0.19.2/updatecli_Linux_x86_64.tar.gz"
+          curl -L -o - $url | tar -xzf - -C /
```
jd, line 25: Could we add a checksum check here?

.github/workflows/updatecli.yaml
```diff
@@ -0,0 +1,54 @@
...
+      # * some files are missing inside the docker container of updatecli
+      # * python is not installed in this container
+      - name: Updatecli
+        id: updatecli
+        run: |
+          version="v0.16.2"
+          url="https://github.com/updatecli/updatecli/releases/download/${version}/updatecli_Linux_x86_64.tar.gz"
+          sha="0a9b8c9a21a3637ebee19b7a499172965d3cfa93884b2a2e25e9068b5e0a5fdf"
+          curl -L -o - $url | tar -xzf - -C /
+          echo "$sha /updatecli" | sha256sum -c -
```
jd, line 28: rather check the tarball

jd (changes requested):

jd (approved):

PR #4068, opened 2022-02-11
refactor(worker): rename some variables

sileht:
Since we also have dedicated stream tasks now and these variables are only used
for shared stream tasks. Let's choice better names.

Change-Id: I2534160dc521178c37e288e590570a366a9325ff


jd (approved):

sileht:
@mergifyio rebase

sileht:
@Mergifyio rebase

PR #4072, opened 2022-02-11
feat(attributes): new repo name attributes (short and full)

ErwanSimonetti:
Adding new attributes `repository-name-short` and `repository-name-full`

Fixes MRGFY-905


ErwanSimonetti:
I'm just missing the release-note because I can't find the exact `reno` command i'm supposed to use

jd (changes requested):
Name update plz 😁

mergify_engine/context.py
```diff
@@ -795,6 +795,14 @@ def base_url(self) -> str:
         # TODO(sileht): remove me when context split if done
         return self.repository.base_url
 
+    @property
```
sileht, line 798: We don't need these new properties, they don't add much value.

mergify_engine/context.py
```diff
@@ -1337,6 +1345,10 @@ async def _get_consolidated_data(self, name: str) -> ContextAttributeType:
                 for t in await self.retrieve_review_threads()
                 if not t["isResolved"]
             ]
+        elif name == "repository-name":
+            return self.repo_name
```
sileht, line 1349: You can directly use `self.repository.repo["name"]`

sileht (changes requested):
Everything looks good to me. Just a couple of lines I think we can remove.

mergify_engine/context.py
```diff
@@ -795,6 +795,14 @@ def base_url(self) -> str:
...
         return self.repository.base_url
 
+    @property
+    def repo_name(self) -> str:
+        return self.repository.repo["name"]
+
+    @property
+    def repo_full_name(self) -> str:
+        return self.repository.repo["full_name"]
+
```
jd, line 805:
```suggestion
```

mergify_engine/context.py
jd, line 1349:
```suggestion
            return self.repository.repo["name"]
```

mergify_engine/context.py
```diff
@@ -1337,6 +1345,10 @@ async def _get_consolidated_data(self, name: str) -> ContextAttributeType:
                 for t in await self.retrieve_review_threads()
                 if not t["isResolved"]
             ]
+        elif name == "repository-name":
+            return self.repo_name
+        elif name == "repository-full-name":
+            return self.repo_full_name
```
jd, line 1351:
```suggestion
            return self.repository.repo["full_name"]
```

jd (changes requested):

jd (approved):

PR #4089, opened 2022-02-14
refactor(worker): avoid calling sismembers too many times

sileht:
The set DEDICATED_WORKERS_KEY can change only every 1 minute. And it
doesn't change so often in reality.

This change creates an in memory cache of this list in all processes and
sync it from Redis every 30s.

Fixes MRGFY-917

Change-Id: Ic44bfea7bfecafd3b55813c6920a7c78711bdd0a

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -257,82 +257,6 @@ async def run_engine(
         logger.debug("engine in thread end")
 
 
-class BucketSelector(typing.Protocol):
-    async def next_org_bucket(self) -> typing.Optional[worker_lua.BucketOrgKeyType]:
-        ...
-
-
-@dataclasses.dataclass
-class DedicatedOrgBucketSelector:
-    redis_stream: utils.RedisStream
-    owner_id: github_types.GitHubAccountIdType
-
-    def _is_org_bucket_for_me(self, org_bucket: bytes) -> bool:
-        owner_id = int(org_bucket.split(b"~")[1])
-        return owner_id == self.owner_id
-
-    async def next_org_bucket(self) -> typing.Optional[worker_lua.BucketOrgKeyType]:
-        # FIXME(sileht): we can use ZSCORE instead of ZRANGEBYSCORE when we
-        # have drop all login/name from bucket names.
-        # NOTE(sileht): Should we drop the 30s latency ? That may impact the rate limit...
-        now = time.time()
-        for org_bucket in await self.redis_stream.zrangebyscore(
-            "streams",
-            min=0,
-            max=now,
-        ):
-            if self._is_org_bucket_for_me(org_bucket):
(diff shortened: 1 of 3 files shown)


mergify_engine/worker.py
```diff
@@ -1130,6 +1110,14 @@ async def start(self) -> None:
         await redis_utils.load_scripts(self._redis_stream)
         await migrations.run(self._redis_cache)
 
+        self._dedicated_workers_syncer_task = asyncio.create_task(
```
jd, line 1113: Can we avoid having a task just for this?
I'm not uber confident that this is not going to be the source of race conditions in the future.
I'd prefer to have a simpler scheme where we check the latest update timestamp in the code path that uses it and refresh it if it's too old before using it.

jd (changes requested):

mergify_engine/worker.py
sileht: The reason behind the dedicated task is,
We have multiple consumers of this data, especially each worker task (so around 60 per-process in production)
In the worse case, we lookup in this set every 0.42s per worker task.
If I use the method using a timestamp and check for expiration, the 60 tasks may update the data at the same time for no value (And I don't think we want to introduce a lock to avoid the update concurrency).

Another solution could be refresh it AFTER using it, so other workers will still use old data during the refresh but don't update the cache multiple times in parallel. This version has exactly the same kind of future source of race conditions of the dedicated task, as workers may use not yet refreshed data.

The code already protects some kind of race conditions around this set. If the subscription and the type of worker (dedicated/shared) don't match, we just update this set in Redis and retry later until everything are updated correctly.

mergify_engine/worker.py
```diff
@@ -884,6 +809,23 @@ async def _consume_pull(
             logger.error("failed to process pull request", exc_info=True)
             raise UnexpectedPullRetry()
 
+    async def should_handle_owner(
```
jd, line 812:
```suggestion
   def should_handle_owner(
```

mergify_engine/worker.py
```diff
@@ -884,6 +809,23 @@ async def _consume_pull(
...
+        self,
+        owner_id: github_types.GitHubAccountIdType,
+        dedicated_worker_owner_ids: typing.Set[github_types.GitHubAccountIdType],
```
jd, line 815: I find it odd that the StreamProcessor does not catch this itself since it's the main user of it.

jd (changes requested):

mergify_engine/worker.py
sileht: We have 30 StreamProcessor per process and don't want to duplicate the set or sync 30 times the set.

jd (approved):

PR #4093, opened 2022-02-16
fix(worker): update root span resource when cache is updated

jd:


sileht (approved):

sileht:
I fixed it and added set_tag, fill free to remove the label if it's ok for you.

PR #4099, opened 2022-02-16
ci(semgrep): ignore ssl usage warning

sileht:
Change-Id: I01b483c0273b2e6b52d0e6cc0a565610dff74052


mergify_engine/utils.py
```diff
@@ -46,7 +46,7 @@ def redis_from_url(url: str, **options: typing.Any) -> yaaredis.StrictRedis:
         final_url = f"redis://{url[len(ssl_scheme):]}"
         ctx = ssl.create_default_context()
         ctx.check_hostname = False
-        ctx.verify_mode = ssl.CERT_NONE
+        ctx.verify_mode = ssl.CERT_NONE  # nosemgrep contrib.dlint.dlint-equivalent.insecure-ssl-use
```
jd, line 49:
```suggestion
        ctx.verify_mode = ssl.CERT_NONE  # nosemgrep: contrib.dlint.dlint-equivalent.insecure-ssl-use
```

jd (approved):

PR #4135, opened 2022-02-17
fix(queue): fix deletion of draft pull request

sileht:
If for some reason the draft pull request branch is not deleted
correctly. The next time we create it, we should first delete the old
one and create the new one.

The current code was deleting the old one, but miss to recreate it and
then when we tried to merge the TrainCar PRs inside it was failing.

It's still unclearer how the draft pull request branch does not get
delete initially.

This change ensures we will recover correctly if this occurs.

### DIFF
--- a/mergify_engine/queue/merge_train.py
+++ b/mergify_engine/queue/merge_train.py
@@ -443,6 +443,32 @@ class TrainCar:
             [str(ep.user_pull_request_number) for ep in self.initial_embarked_pulls]
         )
 
+    @tenacity.retry(
+        retry=tenacity.retry_if_exception_type(tenacity.TryAgain),
+        stop=tenacity.stop_after_attempt(2),
+    )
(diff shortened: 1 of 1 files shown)


mergify_engine/queue/merge_train.py
```diff
@@ -443,6 +443,32 @@ def _get_pulls_branch_ref(self) -> str:
...
+            )
+        except http.HTTPClientSideError as exc:
+            if exc.status_code == 422 and "Reference already exists" in exc.message:
+                try:
+                    await self._delete_branch()
+                except http.HTTPClientSideError as exc_patch:
+                    await self._set_creation_failure(exc_patch.message)
+                    raise TrainCarPullRequestCreationFailure(self) from exc_patch
+
+                raise tenacity.TryAgain
```
sileht, line 467: This was missing ...

jd (approved):

PR #4137, opened 2022-02-17
chore(queue): remove useless retry

sileht:
Since we fix the root bug, we can remove that.


jd (approved):
Would be great to add a link to the previous fix.

PR #4156, opened 2022-02-22
fix(template): raise correct exception

sileht:
We need to raise a Jinja exception to get the error reported
correctly.

Fixes MERGIFY-ENGINE-2JW
Fixes MRGFY-945

Change-Id: Icdbd6f11931db107a731ec31c8c77675769bc10e


jd (approved):

sileht:
@mergifyio refresh

PR #4160, opened 2022-02-22
feat(queue): add queue freeze API tool

Syffe:
The queue freeze API tool offers the ability to
freeze/unfreeze some or all merge queues through new API endpoints.
This feature will allow users to have more control
on their merge queues since the problem was mentionned several
times in various issues.

Fixes [MRGFY-883](https://linear.app/mergify/issue/MRGFY-883)


sileht (dismissed):

mergify_engine/actions/queue.py
```diff
@@ -394,15 +400,23 @@ async def _should_be_queued(self, ctxt: context.Context, q: queue.QueueT) -> boo
...
 
     async def _should_be_merged(
-        self, ctxt: context.Context, rule: "rules.EvaluatedRule", q: queue.QueueT
+        self,
+        ctxt: context.Context,
+        q: queue.QueueBase,
+        p: typing.Optional[QueuePauseData],
     ) -> bool:
-        if not await q.is_first_pull(ctxt):
+        car = typing.cast(merge_train.Train, q).get_car(ctxt)
```
jd, line 408: you don't need to move this here

mergify_engine/actions/queue.py
```diff
@@ -463,9 +477,12 @@ async def get_queue_status(
         ctxt: context.Context,
         rule: "rules.EvaluatedRule",
         q: typing.Optional[queue.QueueBase],
+        p: typing.Optional[QueuePauseData] = None,
     ) -> check_api.Result:
         if q is None:
             title = "The pull request will be merged soon"
+        elif p is not None:
+            title = f'The queue "{p.name}" is currently on pause, with the following message: "{p.message}"'
```
jd, line 485:
```suggestion
            title = f'The queue "{p.name}" is currently on pause, for the following reason: "{p.message}"'
```

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,133 @@
...
+
+@dataclasses.dataclass
+class QueuePauseData:
+
+    # Stored in redis
+    _name: str = dataclasses.field(metadata={"description": "Queue name"})
+    _priority_level: int = dataclasses.field(
+        metadata={"description": "Queue priority level"}
+    )
+    _message: str = dataclasses.field(metadata={"description": "Pause message"})
```
jd, line 37: I'd name this `reason`.

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,133 @@
...
+    _user_app: str = dataclasses.field(
```
jd, line 38: it's not clear what is actually store here: a name, an id?

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,133 @@
...
+        metadata={"description": "App responsible for pause"}
+    )
+    _date: datetime.datetime = dataclasses.field(default_factory=date.utcnow)
```
jd, line 41: date of what? i'd rather use a more explicit name

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,133 @@
...
+    _status: bool = True
```
jd, line 42: status of the pause? not really clear either 😁

mergify_engine/rules/__init__.py
```diff
@@ -375,6 +375,12 @@ def get(self, key: QueueName) -> typing.Optional[QueueRule]:
         except KeyError:
             return None
 
+    def get_prority_level(self, key: QueueName) -> typing.Optional[int]:
+        try:
+            return self[key].config["priority"]
+        except KeyError:
```
jd, line 381: not a fan because you can have a `KeyError` on `self[key]` or `config["priority"]` and I'm sure the second one would be a bug.
Rather catch the one and only one you want or use `in`.

jd (changes requested):

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +324,435 @@ async def repository_queues(
...
         return queues
+
+
+QueuePauseSchema = voluptuous.Schema({voluptuous.Required("message"): str})
+
+
+@router.post(
+    "/repos/{owner}/{repository}/queue/{queue_name}/pause_from",  # noqa: FS003
+    summary="Pauses merge queue(s)",
+    description="Pauses the merge of all merge queues equal to and under the requested queue priority level",
```
jd, line 336:
```suggestion
    description="Pauses the merge of a queue and the queues following it",
```
There's no "priority" across queues per se, just an order of processing. Priority is confusing as it's used to order PR inside the queue.

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +324,435 @@ async def repository_queues(
...
             queues.queues.append(queue)
-
         return queues
+
+
+QueuePauseSchema = voluptuous.Schema({voluptuous.Required("message"): str})
+
+
+@router.post(
+    "/repos/{owner}/{repository}/queue/{queue_name}/pause_from",  # noqa: FS003
```
jd, line 334: why `_from`?

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +324,435 @@ async def repository_queues(
...
+
+        sub = await subscription.Subscription.get_subscription(
+            redis_cache, repo["owner"]["id"]
+        )
+
+        installation = context.Installation(
+            installation_json, sub, client, redis_cache, redis_queue
+        )
+        repository_ctxt = installation.get_repository_from_github_data(repo)
+        config_file_scheme = await repository_ctxt.get_mergify_config_file()
```
jd, line 398: why `_scheme`?

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +324,435 @@ async def repository_queues(
...
+
+        if config_file_scheme and installation.redis_queue:
+            config_file = get_mergify_config(config_file_scheme)
+
+            repo_id = repo["id"]
+            owner_id = installation.owner_id
+            queue_rules = config_file["queue_rules"]
+
+            if queue_name not in [rule.name for rule in queue_rules]:
```
jd, line 407:
```suggestion
            if all(queue_name != rule.name for rule in queue_rules):
```
this makes sure you don't need to iterate over all `queue_rules`

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +324,435 @@ async def repository_queues(
...
+                )
+
+            queue_name_priority = queue_rules.get_prority_level(queue_name)
+            redis_hash = f"merge-pause~{owner_id}"
+
+            for rule in queue_rules:
+                redis_key = f"{repo_id}~{rule.name}"
+
+                if queue_name_priority is not None and (
+                    rule.config["priority"] > queue_name_priority
```
jd, line 530: why are we dealing with priorities here?

mergify_engine/web/api/queues.py
```diff
@@ -312,5 +324,435 @@ async def repository_queues(
...
+            return responses.JSONResponse(
+                status_code=200,
+                content={
+                    "description": f'The queue(s) "{joined_list}" have been unpaused'
+                },
+            )
+
+
+@router.get(
+    "/repos/{owner}/{repository}/queue/{queue_name}/is_paused",  # noqa: FS003
```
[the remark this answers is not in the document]
jd: or `paused`?

jd (changes requested):

mergify_engine/web/api/queues.py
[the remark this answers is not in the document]
Syffe: if "paused" is understandable enough to imply that we "ask if the queue is paused", then we can go for it

mergify_engine/web/api/queues.py
Syffe: Because of the "falling" behavior of the pause system that we discussed, I thought initially that "we pause all queues from the one requested to the last queue behind in processing order of the queues". Formalized like this, I thought adding "from" would be more intuitive to understand the behavior, but it's related to @sileht question above about if we need just one API call of multiple call to pause several queues

mergify_engine/web/api/queues.py
Syffe: Because at first sight I saw that I needed two distinct variable to store the results of get_mergify_config_file and get_mergify_config ! I named them config_file_scheme and config_file, but I just forgot to refactor them to simply config_file and config

mergify_engine/web/api/queues.py
Syffe: As I mentionned uper, though I misunderstood "priority" and "process order", it also answer to the induced falling behavior of the pause system, "if we pause the requested queue, and all queues behind in the process order, we don't need to interact with the queues before in the process order". Thus I check the process order to avoid this

mergify_engine/web/api/queues.py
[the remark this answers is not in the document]
jd: yes if you `GET` or something it's to get the status

mergify_engine/web/api/queues.py
```diff
@@ -312,5 +324,435 @@ async def repository_queues(
...
+                        return responses.JSONResponse(
+                            status_code=200,
+                            content={
+                                "description": f'The queue "{queue_name}" has been unpaused'
+                            },
+                        )
+
+
+@router.get(
+    "/repos/{owner}/{repository}/queue/{queue_name}/unpause_from",  # noqa: FS003
```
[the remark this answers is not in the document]
jd: To me this is not a shortcut: this is the only sane way to make it work.

If you have queues `[A, B, C, D]` and you `pause` `B`, calling `unpause` on `C` and having only `C` unpaused is impossible from the design we decided on.

So there should be only one `unpause` (which is actually `unpause_from`) and that returns an error if you try to unpause `C` while the pause was set on `B`.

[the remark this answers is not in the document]
Syffe: Also, I implemented two different kind of unpause (1:1 and 1:many) following a case that @sileht mentioned to me: 

"Imagine a scenario in which you have two queues "high" and "low", the "low" queue only is set on pause because it is the "no refactor week", then, you have an incident and you need to pause the "high" queue, you solve your incident, but because it is still the "no refactor week", you need to unpause only the "high" queue, letting the "low" queue still on pause" 

To solve this scenario, it is somehow mandatory to have the ability to pause one and only one queue through a 1:1 endpoint, what do you think ?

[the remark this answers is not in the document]
jd: In your scenario you called:
1. first `pause low`
2. then `pause high`

to get back to the desired state you should call
3. `pause low`

🤷 

This also makes sure those pause calls are idempotent, which is a great thing.

Maybe `pause` is not the best word for all of that? We could keep `pause` for a future usage where you can `pause` a individual queue.

Maybe using the `freeze` word would make it "clearer" for this use case, and we can keep `pause` for something else? Or add a future parameter to freeze to allow to freeze/unfreeze a single queue.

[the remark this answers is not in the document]
Syffe: I think we can work around a better termination indeed, freeze seems better ! 

But in the case you described once you've done: 
1. first pause low
2. then pause high

To get back to the state where there is only "low" that is paused, using "unpause-from", and then re-using "pause low", there would be a window of time where there is no paused queue. Or maybe you imply that when you pause from a queue, you check if there is paused queues above, and if yes, you unpause them, in this case "high" ?

mergify_engine/actions/queue.py
```diff
@@ -394,15 +400,22 @@ async def _should_be_queued(self, ctxt: context.Context, q: queue.QueueT) -> boo
...
     async def _should_be_merged(
-        self, ctxt: context.Context, rule: "rules.EvaluatedRule", q: queue.QueueT
+        self,
+        ctxt: context.Context,
+        q: queue.QueueBase,
+        p: typing.Optional[QueuePauseData],
     ) -> bool:
-        if not await q.is_first_pull(ctxt):
+        is_first_pull = await q.is_first_pull(ctxt)
+        if not is_first_pull:
```
jd, line 409: that does not seem to required a change 🤷

mergify_engine/actions/queue.py
```diff
@@ -394,15 +400,22 @@ async def _should_be_queued(self, ctxt: context.Context, q: queue.QueueT) -> boo
...
             return False
 
         car = typing.cast(merge_train.Train, q).get_car(ctxt)
         if car is None:
             return False
 
+        if p is not None:
+            return False
```
jd, line 417: Might be worth checking this first?
That'd avoid requesting stuff from what I understand.

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,124 @@
...
+@dataclasses.dataclass
+class QueuePauseData:
+
+    repository: Repository
+
+    # Stored in redis
+    name: str = dataclasses.field(metadata={"description": "Queue name"})
+    reason: str = dataclasses.field(metadata={"description": "Pause message"})
+    user_app_name: str = dataclasses.field(
+        metadata={"description": "App responsible for pause"}
```
jd, line 38:
```suggestion
        metadata={"description": "Application responsible for pause"}
```

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,124 @@
...
+            redis_hash = f"merge-pause~{installation.owner_id}"
+            redis_key = f"{repo_id}~{queue_name}"
+
+            with utils.yaaredis_for_queue() as redis_queue:
+                queue_pause_data_raw = await redis_queue.hget(redis_hash, redis_key)
+
+                if queue_pause_data_raw:
+                    # NOTE(Syffe): timestamp parameter means that timestamp variables will be converted to
+                    # datetime (value 3=to_datetime()). Other values can be used: 1=to_float(), 2=to_unix_ns()
+                    queue_pause_data = msgpack.unpackb(
```
jd, line 67: any reason we use msgpack here?
i like it but I've seen json mostly used in other places

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,124 @@
...
+                        pause_date=queue_pause_data["pause_date"],
+                        pause_state=queue_pause_data["pause_state"],
+                    )
+                    return queue_pause
+
+                else:
+                    return None
+
+        except Exception:
+            return None
```
jd, line 85: This looks like a trap where we'd miss any bug in this code!

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @classmethod
+    async def get(
+        cls, repository: Repository, queue_name: str
+    ) -> typing.Optional["QueuePauseData"]:
+
+        try:
+            repo_id = repository.repo["id"]
+            installation = repository.installation
+            redis_hash = f"merge-pause~{installation.owner_id}"
```
jd, line 58: you want to use `self._get_redis_hash`

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,124 @@
...
+            redis_key = f"{repo_id}~{queue_name}"
```
jd, line 59: ditto

mergify_engine/queue/queue_pause.py
```diff
@@ -0,0 +1,124 @@
...
+        except Exception:
+            return None
+
+    def _get_redis_hash(self) -> str:
+        return f"merge-pause~{self.repository.installation.owner_id}"
+
+    def _get_redis_key(self) -> str:
+        return f"{self.repository.repo['id']}~{self.name}"
+
+    async def get_json(self) -> typing.Dict[str, typing.Any]:
```
jd, line 93: why `async`?

mergify_engine/web/api/queues.py
```diff
@@ -312,5 +324,435 @@ async def repository_queues(
...
+        redis.get_redis_train
+    ),
+    installation_json: github_types.GitHubInstallation = fastapi.Depends(  # noqa: B008
+        security.get_installation
+    ),
+) -> responses.JSONResponse:
+
+    try:
+        raw_json = await request.json()
+        data = QueuePauseSchema(raw_json)
```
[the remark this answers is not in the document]
jd: This comment still applies, you have 2 lines caught.

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +324,406 @@ async def repository_queues(
...
+
+    except voluptuous.Invalid:
+        return responses.JSONResponse(
+            status_code=400, content={"error": "Invalid json"}
```
jd, line 374: You need to pass the details from voluptuous so the user understand what's wrong.

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +324,406 @@ async def repository_queues(
...
+            redis_cache, repo["owner"]["id"]
+        )
+
+        installation = context.Installation(
+            installation_json, sub, client, redis_cache, redis_queue
+        )
+        repository_ctxt = installation.get_repository_from_github_data(repo)
+        config_file = await repository_ctxt.get_mergify_config_file()
+
+        if config_file and installation.redis_queue:
```
jd, line 401: check for `redis_queue` before? there's no point in getting the `config_file` if you don't use it if `redis_queue` is None.

And I'd tend to use `is not None` to be clear of what you're checking

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +324,406 @@ async def repository_queues(
...
+                    content={"error": f'The queue "{queue_name}" does not exists'},
+                )
+
+            queue_name_priority = queue_rules.get_prority_level(queue_name)
+            paused_queue_list = []
+            for rule in queue_rules:
+
+                if (
+                    queue_name_priority is not None
+                    and rule.config["priority"] > queue_name_priority
```
jd, line 417: I'm still confused on this `priority` but maybe that's just me. @sileht ?

jd (dismissed):

jd:
I thought about the design of all of this and I wanted to add a few cases that are not covered right now: queue addition and deletion (renaming being addition+deletion).

The way it works right now is if you have queue `[A, B, C]` and you `pause_from B`, you get only `A` merging.
Now, if you rename the queue `B` to `X`, there's no freeze anymore. 🤷

I don't think this is a problem per se, and it might be fine working this way. What I don't like is that we're going to store **forever** a free on the queue `B` and there's no way to show it or remove it.

Worse (?), if a user rename back the `X` queue to `B` or add a fourth queue name `B`, then their queue will be paused because we add a pause on a queue named `B` laying around,

That means we might want to either:
1. Make sure we delete pauses from queues that do not exist anymore. This should be made resistant to any race condition.
2. Provide another REST API endpoint that shows the pause list.

I'd tend to make 1. rather than 2.

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,124 @@
...
+    repository: Repository
+
+    # Stored in redis
+    name: str = dataclasses.field(metadata={"description": "Queue name"})
+    reason: str = dataclasses.field(metadata={"description": "Freeze message"})
+    user_app_name: str = dataclasses.field(
+        metadata={"description": "Application responsible for freeze"}
+    )
+    freeze_date: datetime.datetime = dataclasses.field(default_factory=date.utcnow)
+    freeze_state: bool = True
```
sileht, line 41: Since it's always true, I don't think we need to store it.

sileht (changes requested):

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@dataclasses.dataclass
+class QueueFreeze:
+
+    repository: Repository
+
+    # Stored in redis
+    name: str = dataclasses.field(metadata={"description": "Queue name"})
+    reason: str = dataclasses.field(metadata={"description": "Freeze message"})
```
jd, line 36:
```suggestion
    reason: str = dataclasses.field(metadata={"description": "Freeze reason"})
```

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,124 @@
...
+            else:
+                return None
+
+    def _get_redis_hash(self) -> str:
+        return f"merge-freeze~{self.repository.installation.owner_id}"
+
+    def _get_redis_key(self) -> str:
+        return f"{self.repository.repo['id']}~{self.name}"
+
+    def get_json(self) -> typing.Dict[str, typing.Any]:
```
jd, line 93:
```suggestion
    def jsonify(self) -> typing.Dict[str, typing.Any]:
```
I think this is the name we use usually.

Might be worth using a `TypedDict` also?

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,124 @@
...
+        return {
+            "name": self.name,
+            "reason": self.reason,
+            "user_app_name": self.user_app_name,
+            "freeze_date": str(self.freeze_date),  # TO CHANGE
```
jd, line 98: you mean `.isoformat()`?

mergify_engine/tests/functional/actions/test_queue.py
```diff
@@ -2585,6 +2587,766 @@ async def test_queue_manual_merge(self):
 
         await self._assert_cars_contents(q, None, [])
 
+    async def test_create_queue_freeze(self):
```
jd, line 2590: this test is very very very long
it'd be better to split in several `test_*` methods that test the different scenarios

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +325,367 @@ async def repository_queues(
...
+
+        if config_file is not None:
+            config = get_mergify_config(config_file)
+            queue_rules = config["queue_rules"]
+
+            if all(queue_name != rule.name for rule in queue_rules):
+                return responses.JSONResponse(
+                    status_code=400,
+                    content={"error": f'The queue "{queue_name}" does not exists'},
+                )
```
jd, line 412: I thought we'd allow this, right?

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +325,367 @@ async def repository_queues(
...
+
+            qf = await QueueFreeze.get(repository_ctxt, queue_name)
+            if qf is not None:
+                if qf.reason != reason:
```
jd, line 416: This whole block is a little hard to read, rather write:
```python
if qf is None:
   return responses.JSONResponse(…) # already frozen

if qf.reason != reason:
    qf.reason = reason
    await qf.save()
    return responses.JSONResponse(…)

qf = QueueFreeze(…)
```
😉

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +325,367 @@ async def repository_queues(
...
+                        content={
+                            "updated_queue_freeze_data": qf.get_json(),
+                            "description": f'The queue "{queue_name}" was already frozen. Its data have been successfully updated',
+                        },
+                    )
+                else:
+                    return responses.JSONResponse(
+                        status_code=200,
+                        content={
+                            "description": f'The queue "{queue_name}" was already frozen. Nothing happened.'
```
jd, line 430:
```suggestion
                            "description": f'The queue "{queue_name}" is already frozen.'
```

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +325,367 @@ async def repository_queues(
...
+            qf = await QueueFreeze.get(repository_ctxt, queue_name)
+            if qf is not None:
+                if qf.reason != reason:
+                    qf.reason = reason
+                    await qf.save()
+                    return responses.JSONResponse(
+                        status_code=200,
+                        content={
+                            "updated_queue_freeze_data": qf.get_json(),
+                            "description": f'The queue "{queue_name}" was already frozen. Its data have been successfully updated',
```
jd, line 423:
```suggestion
                            "description": f'The queue "{queue_name}" is already frozen. Its data have been successfully updated.',
```
Try to be consistent as if you want to use a final dot `.` or not in the description everywhere.

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +325,367 @@ async def repository_queues(
...
+                status_code=200,
+                content={
+                    "description": f'The queue "{queue_name}" has been successfully frozen'
+                },
+            )
+        else:
+            return responses.JSONResponse(
+                status_code=400,
+                content={
+                    "error": "Configuration file .mergify.yml is missing",
```
jd, line 453:
```suggestion
                    "error": "Mergify configuration file is missing",
```

it could be in `.config/mergify.yml` or the like

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +325,367 @@ async def repository_queues(
...
+                },
+            )
+
+
+@router.get(
+    "/repos/{owner}/{repository}/queues/freeze",  # noqa: FS003
```
jd, line 610:
```suggestion
    "/repos/{owner}/{repository}/queues/freezes",  # noqa: FS003
```
?
Since it returns a list?

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +325,367 @@ async def repository_queues(
...
+            queue_rules = config["queue_rules"]
+
+            frozen = []
+            for rule in queue_rules:
+
+                qf = await QueueFreeze.get(repository_ctxt, rule.name)
+                if qf is None:
+                    continue
+
+                frozen.append(qf.get_json())
```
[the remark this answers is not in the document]
jd: Agreed!

jd (changes requested):
Good stuff!

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +325,367 @@ async def repository_queues(
...
+
+    try:
+        raw_json = await request.json()
+    except json.JSONDecodeError:
+        return responses.JSONResponse(
+            status_code=400, content={"error": "Invalid json"}
+        )
+
+    try:
+        data = QueueFreezeSchema(raw_json)
```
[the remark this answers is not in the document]
Syffe: We might want to discuss if the way I implemented it is okay, but now it seems better indeed !

mergify_engine/tests/functional/actions/test_queue.py
Syffe: Might need some more indication on the best way to separate test scenarios, I tried to it but I'm sure it can be improve

sileht (changes requested):

mergify_engine/tests/functional/actions/test_queue.py
```diff
@@ -2585,6 +2587,974 @@ async def test_queue_manual_merge(self):
...
+                "Content-type": "application/json",
+            },
+        )
+
+        assert r.status_code == 422
+        assert r.json() == {
+            "detail": [
+                {
+                    "loc": ["body"],
+                    "msg": "QueueFreezePayload.__init__() got an unexpected keyword argument 'false_key'",
```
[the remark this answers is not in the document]
Syffe: Yes, I just reported the payload that fastapi+pydantic are sending back when this error case is hit. We can manage to change it though !

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,143 @@
...
+@dataclasses.dataclass
+class QueueFreeze:
+
+    repository: Repository
+
+    # Stored in redis
+    name: str = dataclasses.field(metadata={"description": "Queue name"})
+    reason: str = dataclasses.field(metadata={"description": "Freeze reason"})
+    user_app_name: str = dataclasses.field(
+        metadata={"description": "Application responsible for freeze"}
```
jd, line 38:
```suggestion
        metadata={"description": "Application responsible for the freeze"}
```

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    freeze_date: datetime.datetime = dataclasses.field(default_factory=date.utcnow)
```
jd, line 40: description?

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,143 @@
...
+                qf_list.append(
+                    queues.QueueFreeze(
+                        name=qf["name"],
+                        reason=qf["reason"],
+                        user_app_name=qf["user_app_name"],
+                        freeze_date=qf["freeze_date"],
+                    )
+                )
+
+        return qf_list
```
jd: I would rather `yield` those rather than build the list; if the caller does not need the full list, we'll save time and memory.

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,143 @@
...
+            freeze_date=datetime.datetime.utcnow(),
+        )
+
+        with utils.yaaredis_for_queue() as redis_queue:
+            queue_freeze_data_raw = await redis_queue.hget(
+                queue_freeze._get_redis_hash(),
+                queue_freeze._get_redis_key(),
+            )
+
+            if queue_freeze_data_raw is not None:
```
jd:
```suggestion
            if queue_freeze_data_raw is None:
                return None
```

just saying 😁

mergify_engine/tests/functional/actions/test_queue.py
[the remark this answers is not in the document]
jd: We really should. This really does not look like an error a user would be happy to see, it really makes me thing there's an uncaught error on the *server-side* rather than a problem on the *client-side*.

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,143 @@
+# -*- encoding: utf-8 -*-
```
jd, line 1: nit: you could rename the file `freeze.py` to keep it shorter.

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,143 @@
...
+
+        queue_freeze = cls(
+            repository=repository,
+            name=queue_name,
+            reason="",
+            user_app_name="",
+            freeze_date=datetime.datetime.utcnow(),
+        )
+
+        with utils.yaaredis_for_queue() as redis_queue:
```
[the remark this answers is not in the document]
Syffe: Yes, I forgot to change this following your fix: https://github.com/Mergifyio/mergify-engine/pull/4194/files 

before that I had an error telling me that "None couldn't call hget()" If i remember good !

mergify_engine/web/api/queues.py
```diff
@@ -137,6 +141,43 @@ class Queues:
...
+class QueueFreeze:
+    name: str = dataclasses.field(
+        default_factory=str, metadata={"description": "Queue name"}
+    )
+    reason: str = dataclasses.field(
+        default_factory=str, metadata={"description": "The reason of the queue freeze"}
+    )
+    user_app_name: str = dataclasses.field(
+        default_factory=str,
+        metadata={"description": "Application responsible for freeze"},
```
jd, line 161:
```suggestion
        metadata={"description": "Application responsible for the freeze"},
```

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +358,309 @@ async def repository_queues(
...
+        security.get_installation
+    ),
+) -> QueueFreezeResponse:
+
+    if queue_freeze_payload.reason == "":
+        queue_freeze_payload.reason = "No freeze reason was specified."
+
+    async with github.aget_client(installation_json) as client:
+        try:
+            # Check this token as access to this repository
```
jd, line 400:
```suggestion
            # Check this token has access to this repository
```

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +358,309 @@ async def repository_queues(
...
+        installation = context.Installation(
+            installation_json, sub, client, redis_cache, redis_queue
+        )
+        repository_ctxt = installation.get_repository_from_github_data(repo)
+        config_file = await repository_ctxt.get_mergify_config_file()
+
+        if config_file is None:
+            raise fastapi.HTTPException(
+                status_code=404, detail="Mergify configuration file is missing."
+            )
```
jd, line 420: Do this before getting the `sub`; it'd avoid retrieving the sub if there's no config file.

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +358,309 @@ async def repository_queues(
...
+
+        config = get_mergify_config(config_file)
+        queue_rules = config["queue_rules"]
+
+        if all(queue_name != rule.name for rule in queue_rules):
+            raise fastapi.HTTPException(
+                status_code=404, detail=f'The queue "{queue_name}" does not exists.'
+            )
```
jd, line 428: Do we allow this to make it possible to create a freeze for a queue that does not exist yet?
Just asking, I think it's fine this way too.

mergify_engine/tests/functional/actions/test_queue.py
```diff
@@ -2585,6 +2587,974 @@ async def test_queue_manual_merge(self):
...
+                "Authorization": f"bearer {self.api_key_admin}",
+                "Content-type": "application/json",
+            },
+        )
+
+        assert r.status_code == 404
+        assert r.json() == {
+            "detail": "There is no frozen queues to show.",
+        }
+
```
jd, line 3557: None of the tests in this file actually test that the freeze *works*, e.g., that it does what it means.
We need a test that checks the PR is not merged and that the given reason makes sense for the user in what we report to GitHub.
And than then when we unfreeze, it merges again.

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +358,309 @@ async def repository_queues(
...
+                queue_freeze=[
+                    QueueFreeze(
+                        name=qf.name,
+                        reason=qf.reason,
+                        user_app_name=qf.user_app_name,
+                        freeze_date=qf.freeze_date,
+                    )
+                ],
+                description=f'The queue "{queue_name}" has been successfully frozen.',
+            )
```
jd, line 451:
```suggestion
            description = f'The queue "{queue_name}" has been successfully frozen.'
```

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +358,297 @@ async def repository_queues(
                 )
 
             queues.queues.append(queue)
-
```
jd, line 320: unrelated 😁

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +358,297 @@ async def repository_queues(
...
+                queue_freeze=[
+                    QueueFreeze(
+                        name=qf.name,
+                        reason=qf.reason,
+                        user_app_name=qf.user_app_name,
+                        freeze_date=qf.freeze_date,
+                    )
+                ],
+                description=f'The queue "{queue_name}" is already frozen. Its data have been successfully updated.',
+            )
```
jd, line 466:
```suggestion
        elif qf.reason != queue_freeze_payload.reason:
            qf.reason = queue_freeze_payload.reason
            description = f'The queue "{queue_name}" is already frozen. Its data have been successfully updated.'
        else:
            description = f'The queue "{queue_name}" is already frozen.'
```

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +358,297 @@ async def repository_queues(
...
+        return QueueFreezeResponse(
+            queue_freeze=[
+                QueueFreeze(
+                    name=qf.name,
+                    reason=qf.reason,
+                    user_app_name=qf.user_app_name,
+                    freeze_date=qf.freeze_date,
+                )
+            ],
+            description=f'The queue "{queue_name}" is already frozen.',
```
jd, line 477:
```suggestion
            description=description,
```

mergify_engine/queue/queue_freeze.py
```diff
@@ -0,0 +1,130 @@
...
+                # in order to be serialized
+                datetime=True,
+            ),
+        )
+
+    async def delete(self) -> None:
+        await self.repository.installation.redis_queue.hdel(
+            self._get_redis_hash(self.repository),
+            self._get_redis_key(self.repository, self.name),
+        )
```
jd, line 130:
```suggestion
    async def delete(self) -> bool:
        return bool(await self.repository.installation.redis_queue.hdel(
            self._get_redis_hash(self.repository),
            self._get_redis_key(self.repository, self.name),
        ))
```
`hdel` returns `1` if the deletion happened

mergify_engine/web/api/queues.py
```diff
@@ -317,5 +358,297 @@ async def repository_queues(
...
+        repository_ctxt = installation.get_repository_from_github_data(repo)
+
+        qf = await queue_freeze.QueueFreeze.get(repository_ctxt, queue_name)
+        if qf is None:
+            raise fastapi.HTTPException(
+                status_code=404,
+                detail=f'The queue "{queue_name}" does not exists or is not currently frozen.',
+            )
+
+        await qf.delete()
```
jd, line 533:
```suggestion
        qf = await queue_freeze.QueueFreeze(repository_ctxt, name=queue_name)
        if not await qf.delete()
            raise fastapi.HTTPException(
                status_code=404,
                detail=f'The queue "{queue_name}" does not exists or is not currently frozen.',
            )
```

This avoids doing a `HGET` + `HDEL` and just executes a `HDEL` for the same result, avoiding race conditions and saving time!

jd (changes requested):
We also need a subscription feature flag and we need to check the sub has this flag before doing any operation.

mergify_engine/web/api/queues.py
Syffe: We can allow this, but by doing it we allow the possibility of creating a freeze for a future queue with potential naming errors, if the queue that is created afterwards has a different name, the user will have to get_all->delete->create or get_all->create/update if this case if encountered. Though it doesn't generate any big problems, and it's already solvable so as you want !

mergify_engine/tests/functional/actions/test_queue.py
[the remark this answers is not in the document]
Syffe: I added an exception handler for those cases in order to not sendback the fastapi+pydantic error payload, light need some improvement in the verbose but it seems more appropriate like this

mergify_engine/actions/queue.py
```diff
@@ -258,10 +259,14 @@ async def run(
...
-                    if await self._should_be_merged(ctxt, rule, q):
+                    qf = await freeze.QueueFreeze.get(
+                        ctxt.repository, self.config["name"]
+                    )
+                    if await self._should_be_merged(ctxt, q, qf):
                         result = await self._merge(ctxt, rule, q, merge_bot_account)
                     else:
-                        result = await self.get_queue_status(ctxt, rule, q)
+                        result = await self.get_queue_status(ctxt, rule, q, qf)
+
```
jd, line 269: I see no test on this part in the functional tests.

mergify_engine/actions/queue.py
```diff
@@ -463,9 +475,12 @@ async def get_queue_status(
         ctxt: context.Context,
         rule: "rules.EvaluatedRule",
         q: typing.Optional[queue.QueueBase],
+        qf: typing.Optional[freeze.QueueFreeze] = None,
     ) -> check_api.Result:
         if q is None:
             title = "The pull request will be merged soon"
+        elif qf is not None:
+            title = f'The queue "{qf.name}" is currently frozen, for the following reason: "{qf.reason}"'
```
jd, line 483:
```suggestion
            title = f'The queue "{qf.name}" is currently frozen, for the following reason: {qf.reason}'
```

mergify_engine/web/api/queues.py
```diff
@@ -319,3 +360,272 @@ async def repository_queues(
...
+
+    async with github.aget_client(installation_json) as client:
+        try:
+            # Check this token has access to this repository
+            repo = typing.cast(
+                github_types.GitHubRepository,
+                await client.item(f"/repos/{owner}/{repository}"),
+            )
+        except (http.HTTPNotFound, http.HTTPForbidden, http.HTTPUnauthorized):
+            raise fastapi.HTTPException(status_code=404)
```
jd, line 407: It seems we're (re)using this in multiple place. Can we factorize it?

mergify_engine/web/api/queues.py
jd: Ok, let's keep it that way for now.

mergify_engine/web/api/queues.py
```diff
@@ -319,3 +360,272 @@ async def repository_queues(
...
+                    freeze_date=qf.freeze_date,
+                )
+            ],
+            description=description,
+        )
+
+
+@router.delete(
+    "/repos/{owner}/{repository}/queue/{queue_name}/unfreeze",  # noqa: FS003
+    summary="Unfreezes merge queue",
```
jd, line 463:
```suggestion
    summary="Unfreeze the merge queue",
```

mergify_engine/web/api/queues.py
```diff
@@ -319,3 +360,272 @@ async def repository_queues(
...
+    description="Unfreezes the specified merge queue",
```
jd, line 464:
```suggestion
    description="Unfreeze the specified merge queue",
```

mergify_engine/web/api/queues.py
```diff
@@ -319,3 +360,272 @@ async def repository_queues(
...
+
+        sub = await subscription.Subscription.get_subscription(
+            redis_cache, repo["owner"]["id"]
+        )
+        installation = context.Installation(
+            installation_json, sub, client, redis_cache, redis_queue
+        )
+        repository_ctxt = installation.get_repository_from_github_data(repo)
+
+        qf = await freeze.QueueFreeze.get(repository_ctxt, queue_name)
```
jd, line 506: Don't do a `get` here, see my original comment

mergify_engine/web/api/queues.py
```diff
@@ -319,3 +360,272 @@ async def repository_queues(
...
+        )
+        repository_ctxt = installation.get_repository_from_github_data(repo)
+
+        qf_response = QueueFreezeResponse()
+        qf_response.queue_freeze = [
+            qf async for qf in freeze.QueueFreeze.get_all(repository_ctxt)
+        ]
+        if qf_response.queue_freeze:
+            qf_response.description = (
+                f"there is {len(qf_response.queue_freeze)} queue(s) currently frozen."
```
jd, line 625:
```suggestion
                f"There is {len(qf_response.queue_freeze)} queue(s) currently frozen."
```
I'm sure you can compute if the `s` is needed or not 😉

mergify_engine/web/api/queues.py
```diff
@@ -319,3 +360,272 @@ async def repository_queues(
...
+            )
+            return qf_response
+
+        raise fastapi.HTTPException(
+            status_code=404, detail="There is no frozen queues to show."
```
jd, line 630:
```suggestion
            status_code=404, detail="There is no frozen queues."
```
We should return an empty list, not 404 IMHO.
404 means client did an error and the path does not exist.
the path exists: we just need to return an empty list.

jd (changes requested):
We still need the feature flag check

mergify_engine/actions/queue.py
Syffe: `assert (
            check["output"]["title"]
            == 'The queue "default" is currently frozen, for the following reason: "test freeze reason"'
        )`
there is this part in test_create_queue_freeze(), it asserts the PR status in the queue

jd: I missed it thanks!

Syffe:
flag sub feature is on the way !

mergify_engine/web/api/security.py
```diff
@@ -85,3 +88,45 @@ async def get_installation(
...
+        sub = await subscription.Subscription.get_subscription(
+            redis_cache, repo["owner"]["id"]
+        )
+
+        installation = context.Installation(
+            installation_json, sub, client, redis_cache, redis_queue
+        )
+
+        if installation.subscription.has_feature(subscription.Features.QUEUE_FREEZE):
+            return subscription.Features.QUEUE_FREEZE
```
jd, line 130: Actually, you should return the `installation` here or the like.

The problem is that your current code gets twice the subscription for example, here, and in `get_respository_context`; this is quite suboptimal.

So I think if this function returns the `installation` that it built, you can then pass it to `get_repository_context` and save some code — or you don't even need `get_repository_context` maybe?

jd (changes requested):
LGTM
Also, you need to add doc right?

jd (dismissed):

mergify_engine/web/api/queues.py
```diff
@@ -137,6 +141,43 @@ class Queues:
...
+
+@pydantic.dataclasses.dataclass
+class QueueFreeze:
+    name: str = dataclasses.field(
+        default_factory=str, metadata={"description": "Queue name"}
+    )
+    reason: str = dataclasses.field(
+        default_factory=str, metadata={"description": "The reason of the queue freeze"}
+    )
+    user_app_name: str = dataclasses.field(
```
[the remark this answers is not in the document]
sileht: good catch.

sileht (dismissed):

mergify_engine/web/api/security.py
```diff
@@ -85,3 +88,50 @@ async def get_installation(
...
+        )
+
+        # Check this sub has access to queue_freeze feature
+        if installation.subscription.has_feature(subscription.Features.QUEUE_FREEZE):
+            return installation.get_repository_from_github_data(repo)
+
+        else:
+            raise fastapi.HTTPException(
+                status_code=403,
+                detail=f"⚠ The [subscription](https://dashboard.mergify.com/github/{owner}/subscription) needs to be updated to enable the `queue_freeze` feature.",
```
jd, line 136:
```suggestion
                detail=f"⚠ The [subscription](https://dashboard.mergify.com/github/{owner}/subscription) needs to be upgraded to enable the `queue_freeze` feature.",
```

jd (dismissed):
👍

mergify_engine/web/api/security.py
Syffe: this might need to be changed in other places since I C/C the text !

docs/source/actions/queue.rst
```diff
@@ -213,6 +213,31 @@ For example, if your queue is 15 pull requests long and you have
...
+|premium plan tag|
+
+Mergify allows to freeze the merge of one or several queues at the same time. In order to provide maximum control
+and flexibility on how and when you want the code to be merged.
+
+When a queue is frozen, all the checks and CI configured on the queue rules are still running, it is simply and
+uniquely the merge that is blocked. This allows you to continue your work, development and testing while having
+the security that nothing will be deployed.
+
+The freeze tool is provided through the Mergify API, where you can see the references at: `API references <https://docs.mergify.com/api#/operations/create_queue_freeze_repos__owner___repository__queue__queue_name__freeze_put>`_
```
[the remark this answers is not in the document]
Syffe: I wasn't sure about the generic link at first, but indeed it is risky to put directly the specific link, it is better for us to expect the user to find the different links by him/herself ! gonna change that

docs/source/actions/queue.rst
```diff
@@ -213,6 +213,31 @@ For example, if your queue is 15 pull requests long and you have
...
+
+For example, let's say you have three queues with priorities, respectively "Urgent", "Default" and "Low".
+You are in the "No refactor week" of your project, and want your production branch
+to not deploy anything related to refactoring. By creating a freeze on the "Low" queue, you are now
+assured that nothing from it will be deployed. Meanwhile, you have an important incident and you have to
+freeze all the queues, by creating a freeze on "Urgent". You are now assured that nothing from the
+"Urgent", "Default" and "Low" queue will be deployed, even if no freeze has been created on "Default",
+because "Urgent" have the highest priority. Once your incident is resolved, you can unfreeze the queue
+"Urgent" thus returning to the previous state, having only the queue "Low" frozen.
+
```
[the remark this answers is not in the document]
Syffe: Yes, I think with time and the API growing we might find a solution for better use case explanations !

mergify_engine/web/api/queues.py
```diff
@@ -319,3 +359,165 @@ async def repository_queues(
...
+    repository_ctxt: context.Repository = fastapi.Depends(  # noqa: B008
+        security.get_repository_context_with_queue_freeze_feat_check
+    ),
+) -> QueueFreezeResponse:
+
+    if queue_freeze_payload.reason == "":
+        queue_freeze_payload.reason = "No freeze reason was specified."
+
+    async with github.aget_client(installation_json) as client:
+        repository_ctxt.installation.client = client
```
sileht, line 390: Why do you override the client? You should not need to set up a new client here.

mergify_engine/web/api/security.py
```diff
@@ -85,3 +88,50 @@ async def get_installation(
...
+            redis_cache, installation_json["account"]["id"]
+        )
+
+        installation = context.Installation(
+            installation_json, sub, client, redis_cache, redis_queue
+        )
+
+        # Check this sub has access to queue_freeze feature
+        if installation.subscription.has_feature(subscription.Features.QUEUE_FREEZE):
+            return installation.get_repository_from_github_data(repo)
```
sileht, line 131: I think you should `yield` the data instead of returning it. So the client will not be closed when the function return, but when the HTTP request finishes.

sileht (changes requested):

docs/source/actions/queue.rst
```diff
@@ -213,6 +213,31 @@ For example, if your queue is 15 pull requests long and you have
...
+For example, let's say you have three queues, respectively ``hotfix``, ``default`` and ``lowprio``, in this order.
+
+You are in the middle of an incident for your project, and want your production branch
+to not receive anything else than bug fixes. By creating a freeze on the ``default`` queue, you are now
+assured that nothing from this queue  — nor the queue after it, ``lowprio`` — will be merged.
+
+Meanwhile, you can still merge pull requests that are sent to the ``hotfix`` queue.
+
+Once your incident is resolved, you can unfreeze the queue ``default`` and returns to the previous state.
```
jd, line 240:
```suggestion
Mergify allows freezing the merge of one or several queues simultaneously to provide maximum control
and flexibility on how and when you want the code to be merged.

The freeze feature is available through `the Mergify API <https://docs.mergify.com/api>`_. A set of API endpoints allows you to create, get, update and delete freezes on your queues.

When a queue is frozen, all the checks and CI configured on the queue rules are still scheduled and triggered;
only the merge is blocked. Freezing allows you to continue your work, development, and testing while ensuring that no pull requests will be merged from the queue.

For example, let's say you have three queues, respectively ``hotfix``, ``default`` and ``lowprio``, in this order.

You are in the middle of an incident for your project and want your production branch
not to receive anything other than bug fixes. By creating a freeze on the ``default`` queue, you are now
assured that nothing from this queue  — nor the queues after it, ``lowprio`` — will be merged.

Meanwhile, you can still merge pull requests pushed to the ``hotfix`` queue.

Once your incident is resolved, you can unfreeze the queue ``default`` and returns to the previous state.
```

mergify_engine/web/api/queues.py
Syffe: Indeed, i did this to solve the client closing, but yielding it solves it way better

jd (dismissed):
last doc change hopefully and good to go

jd (commented):
LGTM
@sileht I'll leave it to you to approve

sileht (commented):
It's all good for me!

sileht (approved):

PR #4161, opened 2022-02-22
feat(queue): added RedisQueue database

Syffe:
In order to implement the QueuePause feature, we
introduce the new RedisQueue database that will store
operating datas about Trains/Queues during engine running

Fixes MRGFY-946

### DIFF
--- a/mergify_engine/config.py
+++ b/mergify_engine/config.py
@@ -176,6 +176,7 @@ Schema = voluptuous.Schema(
             "REDIS_CACHE_WEB_MAX_CONNECTIONS", default=None
         ): voluptuous.Any(None, voluptuous.Coerce(int)),
         voluptuous.Required("STORAGE_URL", default="redis://localhost:6379?db=8"): str,
+        voluptuous.Required("QUEUE_URL", default=None): voluptuous.Any(None, str),
         voluptuous.Required("STREAM_URL", default=None): voluptuous.Any(None, str),
         voluptuous.Required("SHARED_STREAM_PROCESSES", default=1): voluptuous.Coerce(
             int
@@ -259,6 +260,7 @@ BOT_USER_ID: int
 BOT_USER_LOGIN: str
 STORAGE_URL: str
 STREAM_URL: str
+QUEUE_URL: str
 BUCKET_PROCESSING_MAX_SECONDS: int
 INTEGRATION_ID: int
 SUBSCRIPTION_BASE_URL: str
@@ -317,6 +319,9 @@ globals().update(Schema(CONFIG))
 
 if globals()["STREAM_URL"] is None:
     STREAM_URL = globals()["STORAGE_URL"]
+
+if globals()["QUEUE_URL"] is None:
+    QUEUE_URL = globals()["STORAGE_URL"]
 
 # NOTE(sileht): Docker can't pass multiline in environment, so we allow to pass
 # it in base64 format
--- a/mergify_engine/tests/unit/test_github_events.py
+++ b/mergify_engine/tests/unit/test_github_events.py
@@ -58,7 +58,7 @@ async def _do_test_event_to_pull_check_run(
         }
     )
     installation = context.Installation(
-        installation_json, mock.Mock(), mock.Mock(), redis_cache
+        installation_json, mock.Mock(), mock.Mock(), redis_cache, mock.Mock()
     )
     pulls = await github_events.extract_pull_numbers_from_event(
         installation,
--- a/mergify_engine/tests/unit/test_worker.py
+++ b/mergify_engine/tests/unit/test_worker.py
@@ -545,7 +545,12 @@ async def test_consume_unexisting_stream(
     get_subscription.side_effect = fake_get_subscription
     get_installation_from_account_id.side_effect = fake_get_installation_from_account_id
     p = worker.StreamProcessor(
-        redis_stream, redis_cache, "shared-8", None, worker.OwnerLoginsCache()
+        redis_stream,
+        redis_cache,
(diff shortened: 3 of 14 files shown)


mergify_engine/tests/conftest.py
```diff
@@ -71,6 +71,19 @@ async def redis_stream() -> typing.AsyncGenerator[utils.RedisStream, None]:
             await utils.stop_pending_yaaredis_tasks()
 
 
+@pytest.fixture()
+async def redis_train() -> typing.AsyncGenerator[utils.RedisTrain, None]:
+    with utils.yaaredis_for_train() as client:
+        await client.flushdb()
+        await redis_utils.load_scripts(client)
```
jd, line 78: Do we need to load all scripts?

sileht (commented):
looks good.

Just wondering if we shouldn't name it `QUEUE_URL`, in case we implement another style of queue and store the data at the same place?

mergify_engine/tests/conftest.py
sileht: mypy should have complain about this one.

mergify_engine/utils.py
```diff
@@ -95,6 +96,25 @@ def yaaredis_for_stream() -> typing.Iterator[RedisCache]:
...
+    max_connections: typing.Optional[int] = None,
+) -> RedisTrain:
+    r = redis_from_url(
+        config.TRAIN_URL, max_idle_time=max_idle_time, max_connections=max_connections
+    )
+    return RedisTrain(r)
+
+
+@contextlib.contextmanager
+def yaaredis_for_train() -> typing.Iterator[RedisCache]:
```
sileht, line 110:
```suggestion
def yaaredis_for_train() -> typing.Iterator[RedisTrain]:
```

mergify_engine/tests/conftest.py
Syffe: Without it, I counldn't generate the fake_repository fixture which made some tests run with errors

Syffe:
> looks good.
> 
> Just wondering if we shouldn't name it `QUEUE_URL`, in case we implement another style of queue and store the data at the same place?

I was thinking the same thing, it might be more appropriate, gonna refactor this !

mergify_engine/tests/unit/test_github_events.py
```diff
@@ -58,7 +58,7 @@ async def _do_test_event_to_pull_check_run(
         }
     )
     installation = context.Installation(
-        installation_json, mock.Mock(), mock.Mock(), redis_cache
+        installation_json, mock.Mock(), mock.Mock(), redis_cache, mock.ANY
```
sileht, line 61: I guess you mean `mock.Mock()` here. `mock.ANY` is when we don't care about a value/object during an asserting. If we don't care about an object in our code we should use mock.Mock() or mock.MagickMock(). Here it's not a big deal as redis_queue is never used by these tests.

```suggestion
        installation_json, mock.Mock(), mock.Mock(), redis_cache, mock.mock()
```

mergify_engine/web/api/queues.py
```diff
@@ -228,6 +228,9 @@ async def repository_queues(
     redis_cache: utils.RedisCache = fastapi.Depends(  # noqa: B008
         redis.get_redis_cache
     ),
+    redis_train: utils.RedisTrain = fastapi.Depends(  # noqa: B008
+        redis.get_redis_cache
```
sileht, line 232:
```suggestion
        redis.get_redis_queue
```

or 

```suggestion
        redis.get_redis_train
```

mergify_engine/config.py
```diff
@@ -259,6 +260,7 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
 BOT_USER_LOGIN: str
 STORAGE_URL: str
 STREAM_URL: str
+QUEUE_URL: str
```
sileht, line 263: If we go with QUEUE_URL, we should use the same terminology everywhere eg: redis_queue, create_yaaredis_queue, ...

mergify_engine/tests/conftest.py
sileht: This hides another issue I think.

Also, Mypy is sick here, `load_scripts` takes a `RedisCache` while we pass a  `RedisTrain`. This should have raised an error. I wonder why it didn't catch it.

sileht (changes requested):

mergify_engine/tests/unit/test_github_events.py
Syffe: Understood !

mergify_engine/config.py
Syffe: Gonna refactor everything from "train" to "queue" seems indeed more appropriate

mergify_engine/tests/unit/test_worker.py
```diff
@@ -1207,10 +1212,10 @@ async def test_stream_processor_retrying_after_read_error(
 
     owners_cache = worker.OwnerLoginsCache()
     p = worker.StreamProcessor(
-        redis_stream, redis_cache, "shared-0", None, owners_cache
+        redis_stream, redis_cache, mock.Mock(), "shared-0", None, owners_cache
     )
 
-    installation = context.Installation(FAKE_INSTALLATION, {}, None, None)
+    installation = context.Installation(FAKE_INSTALLATION, {}, None, None, None)
```
sileht, line 1218: Unrelated to your change, but this should not be None in theory. Mypy didn't catch it because the method signature is not typed :(

sileht (approved):

PR #4166, opened 2022-02-23
feat(github-app): use new dashboard <> engine api keys

sileht:
Fixes MRGFY-947

Change-Id: Iec3bbf896e3bc7d9b5516fa29662d7fe59d9b9a1


mergify_engine/clients/http.py
```diff
@@ -206,7 +206,7 @@ def __init__(
         headers: typing.Optional[httpx_types.HeaderTypes] = None,
         timeout: httpx_types.TimeoutTypes = DEFAULT_TIMEOUT,
         base_url: httpx_types.URLTypes = "",
-    ):
+    ) -> None:
```
jd, line 209: could be in a different pr but ok :)

mergify_engine/config.py
```diff
@@ -164,7 +174,7 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
         voluptuous.Required(
             "WEBHOOK_FORWARD_EVENT_TYPES", default=None
         ): voluptuous.Any(None, CommaSeparatedStringList),
-        # Mergify
+        # Mergify Engine tweaks
```
jd, line 177: tweaks?

jd (approved):

mergify_engine/clients/http.py
sileht: I don't get because mypy didn't see it before...

mergify_engine/config.py
```diff
@@ -155,6 +159,12 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
         voluptuous.Required("APPLICATION_APIKEYS", default=""): voluptuous.Coerce(
             ApplicationAPIKeys
         ),
+        # Saas Special config
+        voluptuous.Required("ENGINE_TO_DASHBOARD_API_KEY"): str,
+        voluptuous.Required("DASHBOARD_TO_ENGINE_API_KEY"): str,
```
jd, line 164: Will this block the startup of onprem?

sileht: I have generated a random secret on startup for onpremise, so that will not block and make this APIs unusable.

jd (approved):

PR #4169, opened 2022-02-23
feat(webhook): allow a second webhook secret

sileht:
This will allow to rotate webhook secret

Fixes MRGFY-948

Change-Id: Iaadc443b8fb33452e5c8887c86c4bb2fb8aa8c39


mergify_engine/config.py
```diff
@@ -252,6 +255,7 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
 WEBHOOK_APP_FORWARD_URL: str
 WEBHOOK_FORWARD_EVENT_TYPES: str
 WEBHOOK_SECRET: str
+WEBHOOK_SECRET_TEMP: typing.Optional[str]
```
jd, line 258: better name maybe?

jd (approved):

PR #4199, opened 2022-03-01
feat(datadog): add tracer for YAML safe_load

Syffe:
Sometimes, YAML.safe_load() takes too long to resolve.
In order to monitor which client is impacted, we need to
trace back the runtime of the function by decorating it
with a tracer span.

Fixes MRGFY-922

### DIFF
--- a/mergify_engine/rules/__init__.py
+++ b/mergify_engine/rules/__init__.py
@@ -22,6 +22,7 @@ import operator
 import typing
 
 import daiquiri
+from ddtrace import tracer
 import voluptuous
 import yaml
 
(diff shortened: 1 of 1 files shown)


mergify_engine/rules/__init__.py
```diff
@@ -411,6 +412,7 @@ def get_annotations(self, path: str) -> typing.List[github_types.GitHubAnnotatio
         return []
 
 
+@tracer.wrap("stream_yaml_load", span_type="worker")
```
jd, line 415:
```suggestion
@tracer.wrap("yaml.load")
```
1. why stream?
2. don't set a type here; worker is not great but I don't think any other is great either.

jd (changes requested):

jd:
The `test` prefix for semantic is for changing tests. That does not look like a test related PR :)

jd (approved):

PR #4220, opened 2022-03-07
feat(queue): implement disallow_checks_interruption_from_queue

sileht:
Change-Id: I84e3c7cd6a697c47c55a97c82a2ac322a9fb29e3


sileht:
I will add a couple of tests that describe the exact use-case of the ticket and I think we are good.

jd (changes requested):

sileht:
@mergifyio squash

jd (approved):

PR #4242, opened 2022-03-11
fix(ghes): handle when Mergify got uninstalled from an org

sileht:
On GHES, if an org uninstall Mergify app, when should just ignore it
from users counts.

Fixes MRGFY-987

### DIFF
--- a/mergify_engine/count_seats.py
+++ b/mergify_engine/count_seats.py
@@ -332,25 +332,34 @@ class Seats:
                     installation["account"]["id"], installation["account"]["login"]
                 )
                 async with github.aget_client(installation) as client:
-                    async for repository in client.items(
-                        "/installation/repositories", list_items="repositories"
-                    ):
-                        repository = typing.cast(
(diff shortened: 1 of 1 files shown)


mergify_engine/count_seats.py
```diff
@@ -331,26 +331,35 @@ async def populate_with_collaborators_with_write_users_access(self) -> None:
...
-                    ):
-                        repository = typing.cast(
-                            github_types.GitHubRepository, repository
-                        )
-                        repo = SeatRepository(repository["id"], repository["name"])
-                        async for collaborator in client.items(
-                            f"{repository['url']}/collaborators"
+                try:
+                    async with github.aget_client(installation) as client:
+                        async for repository in client.items(
```
jd, line 336:
```suggestion

                    async with github.aget_client(installation) as client:
                       try:
                            async for repository in client.items(
```
no?

jd (approved):

PR #4243, opened 2022-03-11
fix(rules): cast InvalidDatetime exception

sileht:
Fixes MERGIFY-ENGINE-2K8


jd (dismissed):
should we have a test?

PR #4244, opened 2022-03-11
fix(date): allow for no space between number and units

jd:
This allows things like "30s" which are quite natural.
This allows to use mins? for minutes too.


sileht (commented):
This becomes != from the postgres parser but ok.

sileht (approved):

PR #4269, opened 2022-03-16
fix(queue): checks_timeout + inplace

sileht:
In case of inplace optimization, we looks at pull_request_rules to known
if we need to unqueue the pull request or not. If we don't known we
wait and don't look at queue_rules conditions.

This change fix the case the queue_rules conditions tell us we must
unqueue the pull request even if pull_request_rules are still waiting.

Fixes MRGFY-993

Change-Id: I8c02319be04dacddaa6597dde90965b19068b148


Syffe (commented):
LGTM @jd I'll let you approve it !

jd (approved):

PR #4270, opened 2022-03-16
feat(crypto): Add mechanism to rotate oauth tokens SECRET

sileht:
Crypto in engine is only for encryting data dashboard cached. We can
just temporary try to decode the data with an old key for the rotation.
The whole cache is re-validated every 3600 and the retention is 3 days.
So we just have to wait 3 days and we are sure all encrypted data are
encrypted with the new key.

Fixes MRGFY-956

### DIFF
--- a/mergify_engine/crypto.py
+++ b/mergify_engine/crypto.py
@@ -16,24 +16,35 @@
 import base64
 import binascii
 import os
+import typing
 
 import cryptography.exceptions
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import ciphers
 from cryptography.hazmat.primitives import hashes
+from datadog import statsd
 
 from mergify_engine import config
 
 
-digest = hashes.Hash(hashes.SHA256(), backend=default_backend())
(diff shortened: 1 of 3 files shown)


mergify_engine/crypto.py
```diff
@@ -56,28 +66,38 @@ def encrypt(value: bytes) -> bytes:
...
+    tag = decoded[IV_BYTES_NEEDED : IV_BYTES_NEEDED + TAG_SIZE_BYTES]
+    value = decoded[IV_BYTES_NEEDED + TAG_SIZE_BYTES :]
+
     try:
-        return decryptor.update(decrypted) + decryptor.finalize()  # type: ignore[no-any-return]
+        try:
+            return _decrypt(iv, tag, value, SECRET_KEY)
+        except cryptography.exceptions.InvalidTag:
+            if SECRET_KEY_OLD is not None:
+                return _decrypt(iv, tag, value, SECRET_KEY_OLD)
```
jd, line 100: do we need metrics around this?

Syffe:
LGTM

jd (approved):

PR #4271, opened 2022-03-16
fix(engine): improve reporting of weird configuration issues

sileht:
This change returns an error if multiple configuration files are found.
Or if the configuration is deleted by a PR.

Fixes MRGFY-994

Change-Id: I8103968f4e185e8e15d5d9b65bd90a2092cf7198


mergify_engine/engine/__init__.py
```diff
@@ -84,26 +89,45 @@ async def _check_configuration_changes(
...
+    config_files: typing.Dict[str, context.MergifyConfigFile] = {}
     async for config_file in ctxt.repository.iter_mergify_config_files(
         ref=ctxt.pull["merge_commit_sha"], preferred_filename=preferred_filename
     ):
-        if (
-            current_mergify_config_file is None
-            or config_file["path"] != current_mergify_config_file["path"]
+        config_files[config_file["path"]] = config_file
+
+    if len(config_files) >= 2:
```
jd, line 98:
```suggestion
    if len(config_files) > 1:
```

jd (approved):

PR #4277, opened 2022-03-17
fix(config): don't always run configuration detector

sileht:
This code can be run only when the pull request is opened or
synchronized.

Depends-On: #4271
Fixes MRGFY-986

Change-Id: I43b29cae06bfc25a8d51758586e71e2307b3fcbf

### DIFF
--- a/mergify_engine/engine/__init__.py
+++ b/mergify_engine/engine/__init__.py
@@ -79,6 +79,25 @@ async def _check_configuration_changes(
         # As the PR is closed, we don't care about the config change detector.
         return False
 
+    # NOTE(sileht): This heuristic works only if _ensure_summary_on_head_sha()
+    # is called after _check_configuration_changes().
+    # If we don't have a summary yet, it means the pull request has just been
+    # open or synchronize or we never see it.
+    summary = await ctxt.get_engine_check_run(constants.SUMMARY_NAME)
+    if (
+        summary
+        and summary["output"]["title"]
+        != constants.CONFIGURATION_MUTIPLE_FOUND_SUMMARY_TITLE
+    ):
+        if await ctxt.get_engine_check_run(constants.CONFIGURATION_CHANGED_CHECK_NAME):
+            return True
+        elif await ctxt.get_engine_check_run(
+            constants.CONFIGURATION_DELETED_CHECK_NAME
+        ):
+            return True
+        else:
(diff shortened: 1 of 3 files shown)


mergify_engine/engine/__init__.py
```diff
@@ -79,6 +79,25 @@ async def _check_configuration_changes(
         # As the PR is closed, we don't care about the config change detector.
         return False
 
+    # NOTE(sileht): This heuristic works only if _ensure_summary_on_head_sha()
+    # is called after _check_configuration_changes().
+    # If we don't have a summary yet, it means the pull request has just been
+    # open or synchronize or we never see it.
+    summary = await ctxt.get_engine_check_run(constants.SUMMARY_NAME)
+    print(summary)
```
jd, line 87:
```suggestion
```

mergify_engine/engine/__init__.py
```diff
@@ -79,6 +79,25 @@ async def _check_configuration_changes(
...
+        and summary["output"]["title"]
+        != constants.CONFIGURATION_MUTIPLE_FOUND_SUMMARY_TITLE
+    ):
+        configuration_changed_check = await ctxt.get_engine_check_run(
+            constants.CONFIGURATION_CHANGED_CHECK_NAME
+        )
+        configuration_deleted_check = await ctxt.get_engine_check_run(
+            constants.CONFIGURATION_DELETED_CHECK_NAME
+        )
+        return bool(configuration_changed_check) or bool(configuration_deleted_check)
```
jd, line 99:
```suggestion
        configuration_changed_check = await ctxt.get_engine_check_run(
            constants.CONFIGURATION_CHANGED_CHECK_NAME
        )
        if configuration_changed_check:
            return True
        configuration_deleted_check = await ctxt.get_engine_check_run(
            constants.CONFIGURATION_DELETED_CHECK_NAME
        )
        return bool(configuration_deleted_check)
```
avoid running 2 awaits if not needed?

jd (dismissed):

mergify_engine/engine/__init__.py
sileht: all checks are always retrieved.

jd (changes requested):
still one comment

jd (approved):

PR #4283, opened 2022-03-18
fix(request-reviewers): support long team names

sileht:
The code was already there for slug only team, but not for fully
qualified teams.

Fixes MRGFY-997

Change-Id: Id33fd3d68b68fc57f9bd905feb998b7b5d7b0077


Syffe (approved):

jd (commented):
no test? 😢

PR #4284, opened 2022-03-18
fix(test): fix empty config parameter in test

Syffe:
The test responsible for checking the case of an empty configuration
file was missing the empty string parameter

Fixes MRGFY-994


sileht (approved):

sileht:
You should rerecord it I guess

sileht (approved):

PR #4293, opened 2022-03-21
fix(api): queue_freeze delete request with response's content size error

Syffe:
The endpoint was returning None, because by default FastAPI uses JSONResponse,
the value None was converted to `b'null'` and thus raising an error because no content
was expected.

Fixes MRGFY-999
Fixes MERGIFY-ENGINE-2KN


sileht (approved):

sileht:
I added sentry ticket to the description

jd:
@sileht @Syffe that would be worth a test though

PR #4294, opened 2022-03-21
feat(queue): add unqueue command

sileht:
Change-Id: Ibbbe6799ecea52b5a026b16dce8efd31e10947bd


docs/source/commands.rst
```diff
@@ -118,4 +118,21 @@ Re-evaluates your Mergify rules on this pull request.
   * - Example
     - ``@Mergifyio refresh``
 
+unqueue
```
Syffe, line 121:
```suggestion
.. _unqueue command:

unqueue
```
for other commands, the ref `.. _action command:` is before the title right ? 

it also seems like the squash command is missing its ref too `.. _squash command:`

Syffe:
otherwise LGTM !

docs/source/commands.rst
sileht: I removed it.

".. _ref" is useful to create anchor or RST reference. We don't really need it as "unqueue" is a title and a title always have an anchor already. We don't need the ref either as it's used nowhere.

releasenotes/notes/unqueue-command-b64fdbd6985ebaa7.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    ``unqueue`` command has been added.
```
jd, line 4:
```suggestion
    The ``unqueue`` command has been added.
```

jd (dismissed):

Syffe (approved):

PR #4303, opened 2022-03-22
feat(sentry): set gh_owner in worker

sileht:
Fixes MRGFY-998

Change-Id: Iffb79a488dac58928bef757c5bb672d1f8b15403

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -39,6 +39,7 @@
 # Pull key format: f"bucket-sources~{repo_id}~{pull_number or 0}"
 #
 
+
 import argparse
 import asyncio
 import collections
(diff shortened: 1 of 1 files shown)


mergify_engine/worker.py
```diff
@@ -1005,9 +1010,11 @@ async def _stream_worker_task(self, stream_processor: StreamProcessor) -> None:
                 resource=owner_login_for_tracing,
             ) as span:
                 span.set_tag("gh_owner", owner_login_for_tracing)
-                await stream_processor.consume(
-                    bucket_org_key, owner_id, owner_login_for_tracing
-                )
+                with sentry_sdk.push_scope() as scope:
+                    scope.set_tag("gh_owner", owner_login_for_tracing)
+                    await stream_processor.consume(
```
Syffe, line 1015: Does this line needs to be in the `with` block, even if it's not using scope ?

jd: Pretty sure it does otherwise the scope is closed and then you get back to the original information.
It's a bit like how the tracer works.

mergify_engine/worker.py
```diff
@@ -39,6 +39,7 @@
 # Pull key format: f"bucket-sources~{repo_id}~{pull_number or 0}"
 #
 
+
```
jd, line 42:
```suggestion
```

jd (approved):

PR #4304, opened 2022-03-22
refactor(api): uses get_repository_context helper in /queues too

sileht:
This change uses the security.get_repository_context() helper in /queues
endpoint.

Change-Id: I769de9454769ff06fe941bc963be25e921fa4b5a


jd:
2 commits, suspicious?

sileht:
> 



> 2 commits, suspicious?

I miss to put it as draft

jd (approved):

PR #4324, opened 2022-03-24
docs(queue): simplify conditions in queue_rules

jd:
There is no need to repeat the CI there, so don't do it to keep things
simpler.


sileht (approved):

sileht:
@Mergifyio rebase

PR #4360, opened 2022-03-31
chore(queue): always queue_pull_request_number

sileht:
We now set it for "in-place" checked car too, this reduces code path
branches.

Change-Id: I1294a5b2a4c8c91228b75b712fb0fdba548d464c


mergify_engine/queue/merge_train.py
```diff
@@ -385,19 +385,11 @@ def _get_embarked_refs(
...
     ) -> typing.List[context.BasePullRequest]:
-        if self.creation_state == "updated":
-            if len(self.still_queued_embarked_pulls) != 1:
-                raise RuntimeError("multiple embarked_pulls but state==updated")
-            ctxt = await self.train.repository.get_pull_request_context(
-                self.still_queued_embarked_pulls[0].user_pull_request_number
-            )
-            return [ctxt.pull_request]
-        elif self.creation_state == "created":
+        if self.creation_state in ["created", "updated"]:
```
jd, line 388:
```suggestion
        if self.creation_state in ("created", "updated"):
```

mergify_engine/queue/merge_train.py
```diff
@@ -823,22 +816,15 @@ async def update_state(
                     self.has_timed_out = True
                     break
 
-        if self.creation_state == "created":
+        if self.creation_state in ["created", "updated"]:
```
jd, line 819:
```suggestion
        if self.creation_state in ("created", "updated"):
```

jd (approved):

sileht:
@mergifyio squash

PR #4367, opened 2022-04-01
fix(queue): change checks_timeout scope and processing

Syffe:
Due to new and future contexts around queue processing, the timeout checks
needed to be adapted to handle these new cases in order to not trigger
false information and behavior like unqueuing PRs that should stay in queue.

Fixes MRGFY-923


mergify_engine/queue/merge_train.py
```diff
@@ -812,14 +840,12 @@ async def update_state(
...
-                    and not condition.match
-                ):
-                    self.has_timed_out = True
-                    break
+        if (
+            self.checks_conclusion != check_api.Conclusion.FAILURE
+            and await self.checks_have_timed_out(date.utcnow())
+        ):
+            self.has_timed_out = True
+            self.checks_conclusion = check_api.Conclusion.FAILURE
```
sileht, line 848: I don't think we should change `self.checks_conclusion`, self.has_timed_out is enough.

mergify_engine/queue/merge_train.py
```diff
@@ -645,9 +649,33 @@ async def _set_initial_state(
...
+            # Circular import
+            from mergify_engine import delayed_refresh
+
+            await delayed_refresh.plan_refresh_at_least_at(
+                self.train.repository,
+                self.queue_pull_request_number,
+                self.creation_date + self.checks_timeout,
+            )
+
+    async def checks_have_timed_out(self, checks_duration: datetime.datetime) -> bool:
```
sileht, line 672:
```suggestion
    def checks_have_timed_out(self, checks_duration: datetime.datetime) -> bool:
```

mergify_engine/queue/merge_train.py
```diff
@@ -645,9 +649,33 @@ async def _set_initial_state(
...
+
         await self.update_state(check_api.Conclusion.PENDING, evaluated_queue_rule)
         await self.update_summaries(check_api.Conclusion.PENDING, force_refresh=True)
 
+    async def set_checks_timeout_context(self, timeout: datetime.timedelta) -> None:
+        self.checks_timeout = timeout
+
+        if self.queue_pull_request_number is not None:
+            # Circular import
+            from mergify_engine import delayed_refresh
```
sileht, line 664: 😢 , I'm surprised we need that, but we can try to break the loop later.

mergify_engine/queue/merge_train.py
```diff
@@ -891,9 +917,26 @@ async def update_summaries(
...
+            expected_finish = (
+                date.pretty_datetime(
+                    date.RelativeDatetime(expected_finish_timestamp).value
+                )
+                if expected_finish_timestamp.date() > date.utcnow().date()
+                else date.pretty_time(
+                    date.RelativeDatetime(expected_finish_timestamp).value
+                )
+            )
+            self.last_evaluated_conditions += (
```
sileht, line 935: We should avoid setting TrainCar state attributes in update_summary(), if we need to update the state we should use update_state().

sileht (dismissed):
Just a couple of thing to move, otherwise that looks good!

mergify_engine/queue/merge_train.py
Syffe: I tried to remove it but could'nt find a workaround, there another place also where this is used in `merge_train.py`

mergify_engine/queue/merge_train.py
Syffe: `self.last_evaluated_conditions` is no longer modified now !

sileht:
I finally got an idea to test that that will be 100% reproducible without the need of freeze/unfreeze.
Create a queue_rules conditions with a schedule condition and prepare the test timing to have the CI timed out outside of the schedule. With the old code, the PR would have been unqueue, while with yours it should stay queued.

Syffe:
Gonna try that !

mergify_engine/queue/merge_train.py
```diff
@@ -645,9 +649,33 @@ async def _set_initial_state(
...
+                self.train.repository,
+                self.queue_pull_request_number,
+                self.creation_date + self.checks_timeout,
+            )
+
+    def checks_have_timed_out(self, checks_duration: datetime.datetime) -> bool:
+        return (
+            (checks_duration - self.creation_date) > self.checks_timeout
+            if self.checks_timeout is not None
+            else False
```
jd, line 676:
```suggestion
        return (
            self.checks_timeout is not None and (checks_duration - self.creation_date) > self.checks_timeout
        )
```

mergify_engine/queue/merge_train.py
```diff
@@ -355,7 +361,9 @@ def deserialize(
             last_checks=last_checks,
             last_evaluated_conditions=data.get("last_evaluated_conditions"),
             has_timed_out=data.get("has_timed_out", False),
+            checks_timeout=data.get("checks_timeout"),
```
sileht, line 364: Since `checks_timeout` is computed from the configuration we should avoid storing it in Redis. In case of the user value change, this value will not be updated. You can get the config with something like:

```
rules = self.train.get_queue_rules()
rule = rules[self.initial_embarked_pulls[0].config["name"]]
```

sileht (changes requested):

mergify_engine/queue/merge_train.py
```diff
@@ -806,20 +830,44 @@ async def update_state(
...
+        if all(ci_conditions_list):
+            self.ci_has_passed = True
+
+        rules = await self.train.get_queue_rules()
+        timeout = (
+            (
+                rules[self.initial_embarked_pulls[0].config["name"]].config[
+                    "checks_timeout"
+                ]
+                if "checks_timeout"
```
sileht, line 850: On parsed configuration all keys are always present (but maybe set to None if voluptuous allow it).

mergify_engine/queue/merge_train.py
```diff
@@ -806,20 +830,44 @@ async def update_state(
...
+        ci_conditions_list = [
+            condition.match
+            for condition in evaluated_queue_rule.conditions.walk()
+            if condition.get_attribute_name().startswith("check-")
+            or condition.get_attribute_name().startswith("status-")
+        ]
+        if all(ci_conditions_list):
+            self.ci_has_passed = True
+
+        rules = await self.train.get_queue_rules()
```
sileht, line 844: If the excepted rule is no more present we should raise an exception. In theory it's impossible since refresh does the cleanup earlier, but who known we don't have or will have a bug one day, so we need to be explicit instead assuming no rules means it's like not having timeout.

mergify_engine/queue/merge_train.py
```diff
@@ -645,9 +649,29 @@ async def _set_initial_state(
             self.train.repository.log,
             False,
         )
+
+        if queue_rule.config["checks_timeout"] is not None:
+            await self.set_checks_timeout_context(queue_rule.config["checks_timeout"])
```
sileht, line 654: I think we should update the delayed refresh date each time the train car state is updated not only when is created? That was the case in previous revision I think, no?

sileht (dismissed):

mergify_engine/queue/merge_train.py
```diff
@@ -254,6 +256,8 @@ class Serialized(typing.TypedDict):
         last_evaluated_conditions: typing.Optional[str]
         has_timed_out: bool
         checks_ended_timestamp: typing.Optional[datetime.datetime]
+        ci_has_passed: bool
+        checks_timeout: bool
```
sileht, line 260: Why do we store that ?

mergify_engine/queue/merge_train.py
```diff
@@ -891,12 +945,40 @@ async def update_summaries(
...
+        timeout_summary = ""
+
+        rules = await self.train.get_queue_rules()
+        timeout = (
+            rules[self.initial_embarked_pulls[0].config["name"]].config[
+                "checks_timeout"
+            ]
+            if rules is not None
+            else None
+        )
```
sileht, line 959: We have this code twice, maybe we can create an helper?

sileht (dismissed):

mergify_engine/queue/merge_train.py
```diff
@@ -797,6 +801,20 @@ async def _set_creation_failure(
                     source="draft pull creation error",
                 )
 
+    async def get_timeout(self) -> typing.Optional[datetime.timedelta]:
+        queue_name = self.initial_embarked_pulls[0].config["name"]
+        rules = await self.train.get_queue_rules()
+
+        if not rules or queue_name not in [rule.name for rule in rules]:
+            raise RuntimeError("Rules are missing from configuration")
```
Syffe, line 809: this is always called no matter the context and thus making the RuntimeError fails other tests like `test_create_pull_basic`, need to work on this

sileht: I see, if that's only tests in TestTrainApiCalls, they should be updated to fix the configuration file.

mergify_engine/queue/merge_train.py
```diff
@@ -797,6 +801,22 @@ async def _set_creation_failure(
                     source="draft pull creation error",
                 )
 
+    async def assert_rules_existence(
```
sileht, line 804: `get_rule()` looks more suited.

sileht (changes requested):

sileht (approved):

PR #4394, opened 2022-04-06
fix(edit): requires GHES >= 3.2

sileht:
Change-Id: Iac2936b7140c19e9f9bd750b751a57d064c5cf06

### DIFF
--- a/mergify_engine/tests/functional/actions/test_edit.py
+++ b/mergify_engine/tests/functional/actions/test_edit.py
@@ -14,12 +14,18 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import pytest
 import yaml
 
+from mergify_engine import config
 from mergify_engine.tests.functional import base
 
 
 class TestEditAction(base.FunctionalTestBase):
+    @pytest.mark.skipif(
+        not config.GITHUB_URL.startswith("https://github.com"),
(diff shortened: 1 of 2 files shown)


mergify_engine/tests/functional/actions/test_edit.py
```diff
@@ -14,12 +14,18 @@
...
 
+from mergify_engine import config
 from mergify_engine.tests.functional import base
 
 
 class TestEditAction(base.FunctionalTestBase):
+    @pytest.mark.skipif(
+        not config.GITHUB_URL.startswith("https://github.com"),
+        reason="requires GHES 3.2",
+    )
```
jd, line 28: This means we won't test it against GHES >= 3.2 either? 😢

sileht: For now yes, the integration run 3.1 so we are OKish, I created a ticket to retrieve the server version and be able to write a better skipif.

Syffe: would it be more understandable with `requires GHES==3.2` ? Otherwise LGTM

jd: It's actually >= 3.2

PR #4406, opened 2022-04-07
chore(tests): change how we track the engine works

sileht:
Instead of just sleeping and looking at Redis content to know if the
engine works, this change track when StreamProcessor.consume run().

This allow to reduce a lot all sleep timing when the engine ran.

test_worker for example takes 15s instead of 24s on my machine.

### DIFF
--- a/mergify_engine/tests/functional/base.py
+++ b/mergify_engine/tests/functional/base.py
@@ -42,6 +42,7 @@ from mergify_engine import github_types
 from mergify_engine import gitter
 from mergify_engine import utils
 from mergify_engine import worker
+from mergify_engine import worker_lua
 from mergify_engine.clients import github
 from mergify_engine.clients import http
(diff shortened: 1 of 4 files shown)


mergify_engine/tests/functional/base.py
```diff
@@ -473,7 +507,7 @@ async def asyncTearDown(self):
         # also to avoid the "git clone fork" failure that Github returns when
         # we create repo too quickly
         if RECORD:
-            await asyncio.sleep(0.5)
+            await asyncio.sleep(self.WAIT_TIME_BERFORE_TEARDOWN)
```
jd, line 510:
```suggestion
            await asyncio.sleep(self.WAIT_TIME_BEFORE_TEARDOWN)
```

jd (changes requested):

PR #4415, opened 2022-04-08
chore(tests): start replacing pytest-httpserver by respx

sileht:
httpserver starts a real http server that is very slow.
respx mock the heart of httpx make tests ran a lot faster.

test_clients takes now ~3s instead of 20s

Change-Id: I376c36c55a30f066b41bc394c210eaea367fde1f

### DIFF
--- a/mergify_engine/tests/unit/test_clients.py
+++ b/mergify_engine/tests/unit/test_clients.py
@@ -15,13 +15,15 @@
 # under the License.
 
 import datetime
+import typing
 from unittest import mock
 
+import httpx
 import pytest
-from pytest_httpserver import httpserver
+import respx
 from werkzeug.http import http_date
-from werkzeug.wrappers import Response
 
+from mergify_engine import config
 from mergify_engine import date
 from mergify_engine import exceptions
 from mergify_engine import github_types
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -58,6 +58,7 @@ test =
     pytest-httpserver==1.0.4
```
jd, line 58: Still used?

mergify_engine/tests/unit/test_clients.py
```diff
@@ -229,47 +205,57 @@ async def test_client_connection_error() -> None:
...
 async def test_client_retry_429_retry_after_as_absolute_date(
-    httpserver: httpserver.HTTPServer,
+    respx_mock: respx.MockRouter,
 ) -> None:
-    retry_after = http_date(date.utcnow() + datetime.timedelta(seconds=3))
-    await _do_test_client_retry_429(httpserver, retry_after, 3)
+    expected_retry = date.utcnow() + datetime.timedelta(seconds=2)
+    retry_after = http_date(expected_retry)
+    when = await _do_test_client_retry_429(respx_mock, retry_after)
+    # ms are cut by http_date, so we allow 1 seconds delta :(
```
Syffe, line 249:
```suggestion
    # ms are cut by http_date, so we allow a 1 second delta :(
```

setup.cfg
sileht: yes by one other file, I will drop very soon.

jd (approved):

PR #4418, opened 2022-04-09
fix(queue): deserialization head_branch

sileht:
That's not a huge issue as head_branch is not used when
`create_state != "created"`, but it's better to set it correctly.

Change-Id: Icc2c85467a0be077956888449da10eb5cf79b7db


jd (approved):

jd (approved):

sileht:
@mergifyio refresh

PR #4436, opened 2022-04-11
PoC(tests): API calls instead of Git command

sileht:
At first glance, it's not faster, but more complicated to maintain.
Not sure it worses it.


sileht:
Just for the record.

PR #4444, opened 2022-04-11
refactor(queue): move existence check of draft pull request

sileht:
This moves the check later in the process so it will be used only when
really needed.

Change-Id: Idffdf529299f3930b1561063d031ea5bf3a03118


jd (approved):

sileht:
@mergifyio rebase

PR #4456, opened 2022-04-12
feat(config): cache loaded config in Repository context

sileht:
Loading the yaml and parsing it with voluptuous is slow.

This change cache the configuration in Repository context.

Since Repository context is recreated each 30s, the cached configuration will
stay valid for 30s.

Fixes MRGFY-1071

Change-Id: Ib81251e0f182feb6e31e9e0698cf22e1c0a32014

### DIFF
--- a/mergify_engine/context.py
+++ b/mergify_engine/context.py
@@ -53,6 +53,9 @@ from mergify_engine.dashboard import subscription as subscription_mod
 from mergify_engine.dashboard import user_tokens
 
 
+if typing.TYPE_CHECKING:
+    from mergify_engine import rules
+
 SUMMARY_SHA_EXPIRATION = 60 * 60 * 24 * 31 * 1  # 1 Month
 
 MARKDOWN_TITLE_RE = re.compile(r"^#+ ", re.I)
@@ -63,6 +66,15 @@ MARKDOWN_COMMENT_RE = re.compile("(<!--.*?-->)", flags=re.DOTALL | re.IGNORECASE
 
 class MergifyConfigFile(github_types.GitHubContentFile):
     decoded_content: bytes
+
+
+DEFAULT_CONFIG_FILE = MergifyConfigFile(
+    decoded_content=b"",
+    type="file",
+    content="<default>",
+    sha=github_types.SHAType("<default>"),
+    path="<default>",
+)
 
 
 class T_PayloadEventSource(typing.TypedDict):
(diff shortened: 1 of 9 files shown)


mergify_engine/context.py
```diff
@@ -371,10 +386,39 @@ async def iter_mergify_config_files(
                 ),
             )
 
+    async def get_mergify_config(self) -> "rules.MergifyConfig":
+        # circular import
+        from mergify_engine import rules
+
+        mergify_config_or_exception = self._caches.mergify_config.get()
+        if mergify_config_or_exception is not cache.Unset:
+            if isinstance(mergify_config_or_exception, rules.InvalidRules):
```
jd, line 395:
```suggestion
            if isinstance(mergify_config_or_exception, Exception):
```

jd (approved):

jd (dismissed):
nitpick

jd:
Also the title is not that great IMHO, the "30s" has no link with the change.

sileht:
> Also the title is not that great IMHO, the "30s" has no link with the change.

I added why it's 30s

mergify_engine/context.py
```diff
@@ -371,10 +386,39 @@ async def iter_mergify_config_files(
                 ),
             )
 
+    async def get_mergify_config(self) -> "rules.MergifyConfig":
+        # circular import
+        from mergify_engine import rules
```
Syffe, line 391: why do we need to import it here since it is mentionned here line 57 ?
`if typing.TYPE_CHECKING:
    from mergify_engine import rules`

mergify_engine/context.py
```diff
@@ -371,10 +386,39 @@ async def iter_mergify_config_files(
...
+
+        mergify_config_or_exception = self._caches.mergify_config.get()
+        if mergify_config_or_exception is not cache.Unset:
+            if isinstance(mergify_config_or_exception, Exception):
+                raise mergify_config_or_exception
```
Syffe, line 396:
```
raise mergify_config_or_exception if isinstance(mergify_config_or_exception, Exception) else return mergify_config_or_exception
``` ?

**Comment by Syffe** [comment]
otherwise LGTM

File: mergify_engine/context.py (hunk 2 above)
Comment (sileht) [reply mergify_engine/context.py]: Fixed

File: mergify_engine/context.py (hunk 3 above)
Comment (Syffe) [reply mergify_engine/context.py:393]: Understood !

File: mergify_engine/tests/unit/rules/test_rules.py [hunk 4]
```diff
@@ -524,7 +524,7 @@ async def item(
     client.item.return_value = item()
     fake_repository.installation.client = client
 
-    fake_repository._caches.mergify_config.delete()
```
Syffe, line 527: don't we also need to keep this line like in the train tests ?

sileht: No this test never goes through `fake_repository.get_mergify_config()`, it plays directly with `rules.get_mergify_config(file)`.

Syffe (approved):

PR #4461, opened 2022-04-12
fix(context): don't retry on-flight anymore

sileht:
We have to run it one time to ensure the payload is complete in case of
the context is built from an event or a GET "/pulls" api call.

This change increases the number of retry by the worker and remove the
"on-flight" retry. This will allow to process other pull requests in
the meantime.

Change-Id: I9e48ce3e454f80318c85392741e015b9fc6305b7


sileht:
CleanShot 2022-04-12 at 19 09 52

jd (approved):

PR #4465, opened 2022-04-13
fix(queue): add missing import

sileht:
Fixes MERGIFY-ENGINE-2MJ

Change-Id: If1fa37f3f6635ee44ddef6e622bbcd6e689ebdf1


jd (approved):
might be great to have a test for this?

PR #4470, opened 2022-04-13
fix(engine): don't handle pull request with more than 500 check-runs

sileht:
Fixes MRGFY-1076

### DIFF
--- a/mergify_engine/check_api.py
+++ b/mergify_engine/check_api.py
@@ -21,9 +21,11 @@ import typing
 
 from mergify_engine import config
 from mergify_engine import date
+from mergify_engine import exceptions
 from mergify_engine import github_types
 from mergify_engine import json
 from mergify_engine import utils
+from mergify_engine.clients import github
 from mergify_engine.clients import http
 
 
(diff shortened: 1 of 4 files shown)


mergify_engine/clients/github.py
```diff
@@ -431,6 +431,7 @@ async def items(
         oauth_token: typing.Optional[github_types.GitHubOAuthToken] = None,
         list_items: typing.Optional[str] = None,
         params: typing.Optional[typing.Dict[str, str]] = None,
+        page_limit: int = 100,
```
jd, line 434: Wondering if we should decrease that in another PR.
I'm having a hard time imagining what will work with 100 pages, whatever the request is.

mergify_engine/check_api.py
```diff
@@ -162,9 +164,15 @@ async def get_checks_for_ref(
...
                     list_items="check_runs",
                     params=params,
+                    page_limit=5,
                 ),
             )
         ]
+    except github.TooManyPages as e:
+        raise exceptions.UnprocessablePullRequest(
+            f"The pull request reports more than {(e.last_page - 1) * 100} check-runs, "
+            "reaching the limit of 500 check-runs."
```
jd, line 174:
```suggestion
            f"The pull request reports more than {(e.last_page - 1) * 100} check runs, "
            "reaching the limit of {page_limit * 100} check runs."
```

jd (dismissed):

mergify_engine/clients/github.py
sileht: Agreed!

sileht: I will propose another PR to fix that.

jd (approved):

PR #4471, opened 2022-04-13
fix(github): set a page_limit everywhere

sileht:
Not all api calls need a page limit of 100, so make it mandatory.

This also allows to report page limit to user everywhere.

### DIFF
--- a/mergify_engine/actions/copy.py
+++ b/mergify_engine/actions/copy.py
@@ -242,7 +242,11 @@ class CopyAction(actions.Action):
                     branch["name"]
                     async for branch in typing.cast(
                         typing.AsyncGenerator[github_types.GitHubBranch, None],
-                        ctxt.client.items(f"{ctxt.base_url}/branches"),
+                        ctxt.client.items(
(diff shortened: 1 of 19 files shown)


sileht:
I tracked every API call that doesn't scale, and keep 100 for those I'm not sure which value we should put, comments are welcome!

mergify_engine/actions/copy.py
```diff
@@ -242,7 +242,11 @@ async def run(
                     branch["name"]
                     async for branch in typing.cast(
                         typing.AsyncGenerator[github_types.GitHubBranch, None],
-                        ctxt.client.items(f"{ctxt.base_url}/branches"),
+                        ctxt.client.items(
+                            f"{ctxt.base_url}/branches",
+                            resource_name="branches",
+                            page_limit=100,
```
jd, line 248: that's a lot?

sileht: fixed

jd (approved):

PR #4485, opened 2022-04-15
fix(engine): move serialized summary on top

sileht:
Check-run summary can be truncated, this ensure we always have the
serialized summary stored.


mergify_engine/engine/actions_runner.py
```diff
@@ -211,11 +211,10 @@ async def get_summary_check_result(
 ) -> typing.Optional[check_api.Result]:
     summary_title, summary = await gen_summary(ctxt, pull_request_rules, match)
 
-    serialized_conclusions = serialize_conclusions(conclusions)
-    summary_for_logging = summary + serialized_conclusions
+    summary += serialize_conclusions(conclusions) + summary
```
Syffe, line 214:
```suggestion
    summary = serialize_conclusions(conclusions) + summary
```
like this right ? otherwise we have summary + serialize_conclusions(conclusions) + summary ?

jd:
@sileht status?

sileht:
I have to re-record everything so…

jd (commented):
Is this retro compatible with existing data?

sileht:
> Is this retro compatible with existing data?

Now yes

jd (approved):

PR #4491, opened 2022-04-15
fix(summary): truncate summary correctly

sileht:
The current code was checking the number of char instead bytes length,
before truncating the string.

This change ensures everything is done in bytes.

Fixes MERGIFY-ENGINE-2MQ
Fixes MERGIFY-ENGINE-2J8
Fixes MRGFY-1082

### DIFF
--- a/mergify_engine/tests/unit/test_utils.py
+++ b/mergify_engine/tests/unit/test_utils.py
@@ -12,6 +12,8 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
+import typing
+
 import msgpack
 import pytest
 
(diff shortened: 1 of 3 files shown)


sileht:
@mergify refresh

sileht:
@mergifyio refresh

jd (changes requested):

mergify_engine/tests/unit/test_utils.py
```diff
@@ -12,35 +12,78 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
+import typing
```
Syffe, line 15: wrong ordering of imports here ?

jd (approved):

PR #4492, opened 2022-04-15
chore(command): remove useless code

sileht:
Fixes MRGFY-1024
Fixes MRGFY-1054


mergify_engine/engine/commands_runner.py
```diff
@@ -314,11 +311,6 @@ def log(comment_out: str, result: typing.Optional[check_api.Result] = None) -> N
             result=result,
         )
 
-    if "@mergifyio" in comment.lower():  # @mergify have been used instead
-        footer = ""
-    else:
-        footer = "\n\n" + WRONG_ACCOUNT_MESSAGE
```
jd, line 320: You want to remove `WRONG_ACCOUNT_MESSAGE` var entirely?

mergify_engine/engine/commands_runner.py
```diff
@@ -297,9 +297,6 @@ async def handle(
     user: typing.Optional[github_types.GitHubAccount],
     rerun: bool = False,
 ) -> None:
-    if "@mergify " not in comment.lower() and "@mergifyio " not in comment.lower():
```
jd, line 300: Why? Where is that checked otherwise?

jd (dismissed):

mergify_engine/engine/commands_runner.py
sileht: The regex in load_command.

PR #4510, opened 2022-04-19
fix(train): handle error when draft PR is not supported by user

Syffe:
Some old plans of GitHub are not able to handle draft PR. We
needed to handle this case properly when train tries to create
a draft PR

Fixes MRGFY-1085

### DIFF
--- a/mergify_engine/clients/http.py
+++ b/mergify_engine/clients/http.py
@@ -64,11 +64,26 @@ class HTTPClientSideError(httpx.HTTPStatusError):
     def message(self) -> str:
         # TODO(sileht): do something with errors and documentation_url when present
         # https://developer.github.com/v3/#client-errors
-        return typing.cast(str, self.response.json()["message"])
+        to_cast = (
+            self.response.json()["errors"][0]["message"]
+            if "errors" in self.response.json()
+            else self.response.json()["message"]
+        )
+        return typing.cast(str, to_cast)
 
     @property
     def status_code(self) -> int:
         return self.response.status_code
+
+    def contains(self, msg_to_find: str) -> bool:
+
+        if "errors" in self.response.json():
+            return any(
+                msg_to_find in error["message"]
+                for error in self.response.json()["errors"]
+            )
+
+        return msg_to_find in self.response.json()["message"]
 
 
 class HTTPForbidden(HTTPClientSideError):
--- a/mergify_engine/queue/merge_train.py
+++ b/mergify_engine/queue/merge_train.py
@@ -539,9 +539,8 @@ class TrainCar:
                 ).json(),
             )
         except http.HTTPClientSideError as e:
-            if e.response.json()["errors"][0]["message"].startswith(
-                "A pull request already exists for"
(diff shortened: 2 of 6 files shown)


mergify_engine/queue/merge_train.py
```diff
@@ -539,6 +539,11 @@ async def _create_draft_pull_request(
                 ).json(),
             )
         except http.HTTPClientSideError as e:
+
+            if "errors" not in e.response.json():
```
jd, line 543: I find this cryptic. Can we add a comment at least?
Or even an example of the JSON you'd expect?

jd (dismissed):

mergify_engine/queue/merge_train.py
jd: I'd even check for the message in the JSON to be clear.

Syffe: Indeed referencing the expected json should do the trick

mergify_engine/queue/merge_train.py
```diff
@@ -539,6 +539,11 @@ async def _create_draft_pull_request(
...
+                await self._set_creation_failure(e.response.json()["message"])
```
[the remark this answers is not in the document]
Syffe: I'm not sure i quite understand the behavior, if the draft PR is not supported by user, we should trigger `raise tenacity.TryAgain`?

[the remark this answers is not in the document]
jd: What he means is that you're doing a minimal version of what the code does in L547-582.
So it'd be better to rewrite L547-549 to include your condition here.

mergify_engine/clients/http.py
```diff
@@ -64,12 +64,34 @@ class HTTPClientSideError(httpx.HTTPStatusError):
...
+            )
+
+        return (
+            True
+            if (
+                msg_to_find in self.response.json()["message"]
+                or msg_to_find in self.response.text
+            )
+            else False
+        )
```
sileht, line 93: It looks like I miss guided you, we don't use response.text anymore, so maybe it's safer to not handle this case

```suggestion
        return msg_to_find in self.response.json()["message"]
```

sileht (changes requested):

mergify_engine/clients/http.py
```diff
@@ -64,12 +64,27 @@ class HTTPClientSideError(httpx.HTTPStatusError):
...
 
+    def contains(self, msg_to_find: str) -> bool:
+
+        if "errors" in self.response.json():
+            return any(
+                msg_to_find in error["message"]
+                for error in self.response.json()["errors"]
+            )
+
+        return True if msg_to_find in self.response.json()["message"] else False
```
sileht, line 86:
```suggestion
        return msg_to_find in self.response.json()["message"]
```

sileht (changes requested):

sileht:
@mergify squash

sileht (approved):

PR #4518, opened 2022-04-21
fix(semgrep): fix new rules

sileht:
Change-Id: I537cf17b0a191ec42998548589927b19d3bb0b2f

### DIFF
--- a/mergify_engine/tests/bridge.py
+++ b/mergify_engine/tests/bridge.py
@@ -19,7 +19,6 @@ import asyncio
 import json
 import logging
 import os
-import time
 
 from mergify_engine import config
 from mergify_engine import logs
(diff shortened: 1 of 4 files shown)


mergify_engine/tests/bridge.py
```diff
@@ -83,7 +82,7 @@ async def main() -> None:
                     )
             except Exception:
                 LOG.error("event handling failure", exc_info=True)
-            time.sleep(1)
+            asyncio.sleep(1)
```
jd, line 85: This seems unrelated @sileht ?

jd: Oh I'm surprised they have a check for this, good stuff anyway.

PR #4527, opened 2022-04-22
docs(config): remove validation endpoint

jd:
This is now barely useful as its uses an old endpoint.
If we need again we should make it on the API.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -443,26 +443,3 @@ We also provide custom Jinja2 filters:
 
    By default, the HTML comments are stripped from ``body``. To get the
    full body, you can use the ``body_raw`` attribute.
-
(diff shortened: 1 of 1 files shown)


sileht (approved):
This endpoint has not been used since a while according logs

nafg:
I was using it https://github.com/nafg/mergify-yaml/blob/master/writer/src/test/scala/io/github/nafg/mergify/demo/Test.scala#L35.

Is there any alternative?

jd:
@nafg https://docs.mergify.com/api/#/operations/simulator_repo_repos__owner___repository__simulator_post

nafg:
Sounds like that requires creating an Application and using its API key. Not very good for a unit test in an open source project.

PR #4536, opened 2022-04-22
feat(queue): change default of allow_inplace_checks

sileht:
Fixes MRGFY-1011

Change-Id: I68610fd43130a165a6ef97741c912afee698e50a


docs/source/actions/queue.rst
```diff
@@ -424,7 +424,7 @@ A ``queue_rules`` takes the following parameter:
        See :ref:`speculative checks`.
    * - ``allow_inplace_checks``
      - bool
-     - True
+     - True if ``speculative_checks: 1`` otherwise False.
```
jd, line 427:
```suggestion
     - True if ``speculative_checks: 1``, False otherwise.
```

mergify_engine/rules/__init__.py
```diff
@@ -116,6 +116,12 @@ def from_dict(cls, d: T_from_dict) -> "QueueRule":
         if allow_checks_interruption is False:
             d["disallow_checks_interruption_from_queues"].append(name)
 
+        # dynamic defaults
+        if d["allow_inplace_checks"] is None:
+            d["allow_inplace_checks"] = (
+                d["speculative_checks"] == 1 and d["batch_size"] == 1
```
jd, line 122: Why checking `batch_size`?
If `batch_size > 1` anyhow the settings will have no effect AFAICT.

sileht: When you have ``batch_size: 2``, if you only have 1 PR queued, you will do `in place` and if you have more you create draft PRs.

This change makes the behavior consistent, if you set 2 or more we always create a draft. This is the same logic as speculative_checks.

If people what the old behavior they can still set `allow_inplace_checks: True`.

sileht:
@mergifyio squash

sileht:
@Mergifyio squash

docs/source/actions/queue.rst
```diff
@@ -424,7 +424,7 @@ A ``queue_rules`` takes the following parameter:
        See :ref:`speculative checks`.
    * - ``allow_inplace_checks``
      - bool
-     - True
+     - True if ``speculative_checks: 1`` and ``batch_size: 1`, False otherwise.
```
sileht, line 427:
```suggestion
     - True if ``speculative_checks: 1`` and ``batch_size: 1``, False otherwise.
```

sileht:
@mergify squash

PR #4551, opened 2022-04-25
test: close idle Redis connections faster

sileht:
This saves between 0 and 2 second each time a test teardown.


jd (approved):

sileht:
@mergify requeue

PR #4562, opened 2022-04-26
feat(config): cache config file content into redis

Syffe:
To be able to list all the queues of a config through a future API endpoint,
we needed to cache its content into redis first to avoid requesting GitHub each
time we want to parse the config file.

Fixes MRGFY-1098


Syffe:
Might need some answers on these points: 

- not sure if my test is well enough putt together because i use calls to redis
- Might create another test to check the good deletion of the cache when `push`event is received ?
- Should we use MessagePack to cache the config file instead of Base64 ?

mergify_engine/context.py
```diff
@@ -426,18 +426,31 @@ async def get_mergify_config_file(self) -> typing.Optional[MergifyConfigFile]:
...
         async for config_file in self.iter_mergify_config_files(
             preferred_filename=cached_filename
         ):
             if cached_filename != config_file["path"]:
                 await self.installation.redis.set(
                     config_location_cache, config_file["path"], ex=60 * 60 * 24 * 31
                 )
+
+            if cached_config_file_content != config_file["content"]:
+                await self.installation.redis.set(
```
jd, line 445: I feel this should be in a pipe?

mergify_engine/context.py
```diff
@@ -426,18 +426,31 @@ async def get_mergify_config_file(self) -> typing.Optional[MergifyConfigFile]:
...
+                    config_file_content_key, config_file["content"]
+                )
+
             self._caches.mergify_config_file.set(config_file)
             return config_file
 
         await self.installation.redis.delete(config_location_cache)
+        await self.installation.redis.delete(config_file_content_key)
```
jd, line 453: this also should be in a pipe?

mergify_engine/context.py
```diff
@@ -529,6 +566,7 @@ def _users_permission_cache_key_for_repo(
             f"{cls.USERS_PERMISSION_CACHE_KEY_DELIMITER}{repo_id}"
         )
 
+    # NOTE(Syffe): this is deprecated
```
jd, line 569: why?

mergify_engine/context.py
```diff
@@ -426,18 +426,31 @@ async def get_mergify_config_file(self) -> typing.Optional[MergifyConfigFile]:
...
         ):
             if cached_filename != config_file["path"]:
                 await self.installation.redis.set(
                     config_location_cache, config_file["path"], ex=60 * 60 * 24 * 31
                 )
+
+            if cached_config_file_content != config_file["content"]:
+                await self.installation.redis.set(
+                    config_file_content_key, config_file["content"]
+                )
```
jd, line 447: I feel like we should set an expiration like we do above

mergify_engine/github_events.py
```diff
@@ -271,6 +271,10 @@ async def push_to_worker(
         elif event["repository"]["archived"]:  # pragma: no cover
             ignore_reason = "repository archived"
 
+        await context.Repository.clear_config_file_content_cache(
+            redis_cache, owner_id, repo_id
+        )
+
     elif event_type == "check_suite":
```
jd, line 278: you should also clear it on push event no?

mergify_engine/tests/unit/rules/test_rules.py
```diff
@@ -537,6 +537,57 @@ async def item(
...
+        fake_repository.installation.owner_id,
+        fake_repository.repo["id"],
+    )
+
+    cached_config_file_content = await fake_repository.installation.redis.get(
+        config_file_content_key
+    )
+    assert cached_config_file_content is None
+
+
```
jd, line 590: I'd add a test case where you actually change the configuration file and it shows you don't use the old config file from the cache. 
It'd fail right now AFAICT.

jd (changes requested):

mergify_engine/context.py
Syffe: Thought I would note it in order to delete it in a separated PR later ! How should i behave in this kind of case ? just note it for myself, create a linear ticket or just delete it into this PR ?

mergify_engine/context.py
Syffe: If this is in a pipe, should `config_location_cache` handle also be in a pipe ?

mergify_engine/tests/unit/rules/test_rules.py
Syffe: Indeed !

mergify_engine/context.py
Syffe: performance wise it would be better indeed !

mergify_engine/context.py
jd: Yes that's what I meant :)

mergify_engine/github_events.py
jd: My bad, I only saw the "archive" line above. That's fine!

mergify_engine/tests/unit/rules/test_rules.py
Syffe: I still think it would be great to have a test case where the config file changes or is cleared when a push event is received ! to be totally complete !

mergify_engine/context.py
```diff
@@ -426,19 +426,43 @@ async def get_mergify_config_file(self) -> typing.Optional[MergifyConfigFile]:
...
+                    config_file["path"],
+                    ex=60 * 60 * 24 * 31,
                 )
+
+            if cached_config_file_content != config_file["content"]:
+                await pipeline.set(
+                    config_file_content_key,
+                    config_file["content"],
+                    ex=60 * 60 * 24 * 31,
+                )
```
jd, line 456: So you cache the content in Redis… but you never use it?

jd (dismissed):

mergify_engine/context.py
```diff
@@ -423,22 +416,66 @@ async def get_mergify_config_file(self) -> typing.Optional[MergifyConfigFile]:
...
+                    "sha": config_file["sha"],
+                },
+                ex=60 * 60 * 24 * 31,
+            )
             self._caches.mergify_config_file.set(config_file)
+            await pipeline.execute()
             return config_file
 
-        await self.installation.redis.delete(config_location_cache)
+        await pipeline.delete(config_file_cache_key)
```
jd, line 455: you actually don't need a pipeline here anymore.

mergify_engine/context.py
```diff
@@ -423,22 +416,66 @@ async def get_mergify_config_file(self) -> typing.Optional[MergifyConfigFile]:
...
+    async def get_cached_config_file(
+        self,
+        repo_id: github_types.GitHubRepositoryIdType,
+    ) -> typing.Optional[MergifyConfigFile]:
+        config_file = await self.installation.redis.get(
+            self.get_config_file_cache_key(repo_id),
+        )
+
+        if config_file is not None:
+            jsonified_config_file = json.loads(config_file.replace("'", '"'))
```
jd, line 469: what is this replace sorcery?

jd (dismissed):

mergify_engine/context.py
Syffe: Ahah ! This sorcery is because Redis, when storing a `dict` is replacing somehow all the `"` with `'`, when we `redis.get()` the result is an `str` of the `dict` without any `"`, since `json.loads()` can't work without a properly formatted string, I needed to replace the `'`. But I'm sure there is a better solution, maybe formatting better before using `redis.set()` ?

jd: Did you serialize it in redis?

Syffe: Nope, did not serialised the dict format before `redis.set()`

jd: You need to call `json.dumps`!

mergify_engine/context.py
```diff
@@ -422,24 +415,68 @@ async def get_mergify_config_file(self) -> typing.Optional[MergifyConfigFile]:
...
+
+            await pipeline.set(
+                config_file_cache_key,
+                json.dumps(
+                    {
+                        "type": config_file["type"],
+                        "content": config_file["content"],
+                        "path": config_file["path"],
+                        "sha": config_file["sha"],
+                    }
```
jd, line 448: It'd be better to have a `serialize` method on the `MergifyConfigFile` that does this.

mergify_engine/context.py
```diff
@@ -422,24 +415,68 @@ async def get_mergify_config_file(self) -> typing.Optional[MergifyConfigFile]:
...
+        if config_file is not None:
+            jsonified_config_file = json.loads(config_file)
+            return MergifyConfigFile(
+                type=jsonified_config_file["type"],
+                content=jsonified_config_file["content"],
+                path=jsonified_config_file["path"],
+                sha=jsonified_config_file["sha"],
+                decoded_content=base64.b64decode(
+                    bytearray(jsonified_config_file["content"], "utf-8")
+                ).decode(),
```
jd, line 476: And here a `deserialize` method that is a classmethod.

jd (changes requested):
really small change but LGTM otherwise

jd (approved):

PR #4563, opened 2022-04-26
fix(http): adapt ClientSideError payload format changes

Syffe:
It seems like the list of errors in the payload sent from GitHub does only contain
the error's string. Github added new payload format that needs to be handled. 

Fixes MRGFY-1112
Fixes MERGIFY-ENGINE-2N4

### DIFF
--- a/mergify_engine/clients/http.py
+++ b/mergify_engine/clients/http.py
@@ -64,26 +64,20 @@ class HTTPClientSideError(httpx.HTTPStatusError):
     def message(self) -> str:
         # TODO(sileht): do something with errors and documentation_url when present
         # https://developer.github.com/v3/#client-errors
-        to_cast = (
-            self.response.json()["errors"][0]["message"]
-            if "errors" in self.response.json()
-            else self.response.json()["message"]
-        )
-        return typing.cast(str, to_cast)
+        response = self.response.json()
+        message = response.get("message", "No error message provided by GitHub")
+
+        if "errors" in response:
+            if "message" in response["errors"][0]:
+                message = response["errors"][0]["message"]
+            elif isinstance(response["errors"][0], str):
+                message = response["errors"][0]
+
+        return typing.cast(str, message)
 
     @property
     def status_code(self) -> int:
         return self.response.status_code
-
-    def contains(self, msg_to_find: str) -> bool:
-
-        if "errors" in self.response.json():
-            return any(
-                msg_to_find in error["message"]
-                for error in self.response.json()["errors"]
-            )
-
-        return msg_to_find in self.response.json()["message"]
 
 
 class HTTPForbidden(HTTPClientSideError):
--- a/mergify_engine/actions/delete_head_branch.py
+++ b/mergify_engine/actions/delete_head_branch.py
@@ -80,7 +80,7 @@ class DeleteHeadBranchAction(actions.Action):
             await ctxt.client.delete(f"{ctxt.base_url}/git/refs/heads/{ref_to_delete}")
         except http.HTTPClientSideError as e:
             if e.status_code == 404 or (
-                e.status_code == 422 and e.contains("Reference does not exist")
+                e.status_code == 422 and "Reference does not exist" in e.message
             ):
                 return check_api.Result(
                     check_api.Conclusion.SUCCESS,
--- a/mergify_engine/check_api.py
+++ b/mergify_engine/check_api.py
@@ -173,7 +173,7 @@ async def get_checks_for_ref(
             )
         ]
     except http.HTTPClientSideError as e:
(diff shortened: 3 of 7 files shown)


jd:
added `Fixes MERGIFY-ENGINE-2N4` for you

jd (changes requested):
How do we know? can we have at least one test?
How does that work with GHES?

Syffe:
Gonna invest some testing !

sileht (dismissed):

sileht:
> Gonna invest some testing !

Maybe just instance the exception with some fake `httpx.Response()` based on the two issues we got on Sentry

jd (changes requested):
Also what about GHES?

sileht:
> Also what about GHES?

Error messages are completely inconsistent... here some examples:
```
{"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/pulls#list-pull-requests-files"}
{"message":"Reference does not exist"}
{"message":"Server Error: Sorry, this diff is taking too long to generate.","errors":[{"resource":"PullRequest","field":"diff","code":"not_available"}],"documentation_url":"https://docs.github.com/rest/reference/pulls#list-pull-requests-files"}

{"message":"Validation Failed","errors":[{"resource":"Issue","code":"custom","field":"body","message":"body is too long (maximum is 65536 characters)"}],"documentation_url":"https://docs.github.com/rest/reference/pulls#create-a-pull-request"}

{"message":"No commit found for SHA: None","documentation_url":"https://docs.github.com/rest/reference/repos#get-a-commit"}

```

Syffe:
I modified the method to handle all the various formats I could find. 

Also tested them into a new test.

mergify_engine/clients/http.py
```diff
@@ -64,12 +64,16 @@ class HTTPClientSideError(httpx.HTTPStatusError):
...
         # TODO(sileht): do something with errors and documentation_url when present
         # https://developer.github.com/v3/#client-errors
-        to_cast = (
-            self.response.json()["errors"][0]["message"]
-            if "errors" in self.response.json()
-            else self.response.json()["message"]
-        )
-        return typing.cast(str, to_cast)
+        response = self.response.json()
+        message = response["message"] if "message" in response else "No message found"
```
jd, line 68:
```suggestion
        message = response.get("message", "No message found")
```

mergify_engine/clients/http.py
```diff
@@ -64,12 +64,16 @@ class HTTPClientSideError(httpx.HTTPStatusError):
...
+
+        if "errors" in response:
+            if "message" in response["errors"][0]:
+                message = response["errors"][0]["message"]
+            elif isinstance(response["errors"][0], str):
+                message = response["errors"][0]
+
```
jd, line 75:
```suggestion
        if "errors" in response:
            if "message" in response["errors"][0]:
                message = response["errors"][0]["message"]
            elif isinstance(response["errors"][0], str):
                message = response["errors"][0]
        else:
            message = response.get("message", "No message found")
```

jd (dismissed):

mergify_engine/clients/http.py
Syffe, line 68: I wasn't sure about setting the message to `None`or to `No message found`, in case there is no `message`key in the payload, although this case is practically impossible, WDYT ?

mergify_engine/clients/http.py
```diff
@@ -78,10 +82,7 @@ def status_code(self) -> int:
     def contains(self, msg_to_find: str) -> bool:
 
         if "errors" in self.response.json():
-            return any(
-                msg_to_find in error["message"]
-                for error in self.response.json()["errors"]
-            )
+            return any(msg_to_find in error for error in self.response.json()["errors"])
```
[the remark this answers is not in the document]
Syffe: Yes I didn't applied the changes, first i needed validation for the upper method

mergify_engine/clients/http.py
Syffe: If we do this we are not catching the follow format: 

`{"message":"Server Error: Sorry, this diff is taking too long to generate.","errors":[{"resource":"PullRequest","field":"diff","code":"not_available"}],"documentation_url":"https://docs.github.com/rest/reference/pulls#list-pull-requests-files"}
`

jd (commented):
LGTM
@sileht I'll leave it to you to double check

mergify_engine/clients/http.py
jd: I'd prefer to change the typing and return `None` if no message is found.
But then it's up to the caller to handle the `None` case.

jd (approved):

PR #4568, opened 2022-04-26
feat(gitter): fixup git user configuration

sileht:
We was blindly use noreply email and a random name.

This change uses the real mergify app login.
And set the noreply email of impersonated user.

Fixes MRGFY-1119

Change-Id: Ib2f6a6d013d8a3d9abf31b29812ff772b6939c8c


mergify_engine/config.py
```diff
@@ -347,6 +345,10 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
 
 globals().update(Schema(CONFIG))
 
+BOT_USER_ID = globals()["BOT_USER_ID"]  # mypy crap
+BOT_USER_LOGIN = globals()["BOT_USER_LOGIN"]  # mypy crap
```
jd, line 349:
```suggestion
BOT_USER_ID = globals()["BOT_USER_ID"]  # mypy workaround
BOT_USER_LOGIN = globals()["BOT_USER_LOGIN"]  # mypy workaround
```

mergify_engine/branch_updater.py
```diff
@@ -102,7 +103,8 @@ async def pre_rebase_check(ctxt: context.Context) -> None:
 async def _do_rebase(
     ctxt: context.Context,
     user: user_tokens.UserTokensUser,
-    bot_account_feature: subscription.Features,
+    committer_name: str,
+    committer_email: str,
```
jd, line 107:
```suggestion
    user: GitHubUser,
```
I think that would prevent repeating the `@users.noreply.github.com` in different places and the `name or login` in different places too.

jd (dismissed):

mergify_engine/gitter.py
```diff
@@ -226,14 +227,21 @@ async def cleanup(self) -> None:
...
     async def configure(
-        self, name: typing.Optional[str] = None, email: typing.Optional[str] = None
+        self, user: typing.Optional[user_tokens.UserTokensUser] = None
     ) -> None:
-        if name is None:
-            name = f"{config.CONTEXT}-bot"
-        if email is None:
-            email = config.GIT_EMAIL
+        if user is None:
+            name = config.BOT_USER_LOGIN
```
jd, line 233: I'd rather use `Mergify` here 😄

jd (dismissed):
just one detail

jd (approved):

PR #4624, opened 2022-04-29
feat(commits-behind): return only the number

sileht:
This change returns only the number of commits behind instead of sha to
leverage the GitHub commit comparaison API.

This returns the exact number in only one API calls, the API has been
introduced in 3.0.

Fixes MRGFY-1130

Change-Id: I17ae7c564d46aa89ec06427694834428544cd05d


docs/source/conditions.rst
```diff
@@ -211,9 +211,9 @@ Here's the list of pull request attribute that can be used in conditions:
      - :ref:`Timestamp <iso timestamp>` or :ref:`Relative timestamp <relative timestamp>`
      - The time the pull request mergeability checks have started at.
    * - ``commits-behind``
-     - list of string
-     - The list of commits sha between the head of the base branch and the base
-       of the pull request (maximum 100 last commits sha).
+     - Number of commits behind its base branch
+     - The number of commits between the head of the base branch and the base
+       of the pull request.
```
jd, line 216: I'd rather keep the current implementation but only makes `#commits-behind` work.
Just raise if `commits-behind` without `#` is used.

jd (changes requested):

docs/source/conditions.rst
jd, line 216:
```suggestion
     - list of commits
     - The list of commits between the head of the base branch and the base
       of the pull request. This can only be used with the length operator as `#commits-behind`.
```

mergify_engine/context.py
```diff
@@ -810,9 +809,9 @@ class ContextCaches:
     commits: cache.SingleCache[
         typing.List[github_types.CachedGitHubBranchCommit]
     ] = dataclasses.field(default_factory=cache.SingleCache)
-    commits_behind: cache.SingleCache[
-        typing.List[github_types.SHAType]
-    ] = dataclasses.field(default_factory=cache.SingleCache)
+    commits_behind: cache.SingleCache[int] = dataclasses.field(
```
jd, line 812:
```suggestion
    commits_behind_len: cache.SingleCache[int] = dataclasses.field(
```
or `_number`

mergify_engine/rules/parser.py
```diff
@@ -290,6 +292,10 @@ def parse(v: str) -> typing.Any:
 
     position = _skip_ws(v, length, position + len(attribute))
 
+    if not quantity and attribute in ATTRIBUTES_WITH_ONLY_LENGTH:
+        raise ConditionParsingError(
+            f"`#` modifier is required for attribute: `{attribute}`"
```
jd, line 297:
```suggestion
            f"`#` operator is required for attribute: `{attribute}`"
```

mergify_engine/tests/functional/actions/test_merge.py
```diff
@@ -407,9 +407,8 @@ async def test_merge_fastforward_basic(self):
         )
         assert p["head"]["sha"] == branch["commit"]["sha"]
 
-        # NOTE(sileht): GitHub seems completly lost with a manual fast-forward merge
-        assert branch["commit"]["committer"] is None
-        # assert branch["commit"]["committer"]["login"] == config.BOT_USER_LOGIN
+        assert branch["commit"]["committer"] is not None
+        assert branch["commit"]["committer"]["login"] == config.BOT_USER_LOGIN
```
jd, line 411: unrelated?

jd (changes requested):

mergify_engine/tests/functional/actions/test_merge.py
sileht: recording artefact

mergify_engine/rules/parser.py
sileht: Hum, use used modifier everywhere. Operator is reserved for > < = ...

jd (approved):

PR #4637, opened 2022-05-02
refactor(redis-clients): create an abstraction for all clients

sileht:
This introduces a new object RedisLinks that hold and lazy load
reference to all Redis clients.

Fixes MRGFY-1113

### DIFF
--- a/mergify_engine/config.py
+++ b/mergify_engine/config.py
@@ -218,9 +218,9 @@ Schema = voluptuous.Schema(
             "TESTING_FORWARDER_ENDPOINT",
             default="https://test-forwarder.mergify.io/events",
         ): str,
-        voluptuous.Required("INSTALLATION_ID", default=15398551): voluptuous.Coerce(
-            int
-        ),
+        voluptuous.Required(
+            "TESTING_INSTALLATION_ID", default=15398551
+        ): voluptuous.Coerce(int),
         voluptuous.Required(
             "TESTING_REPOSITORY_ID", default=258840104
         ): voluptuous.Coerce(int),
@@ -318,6 +318,7 @@ TESTING_MERGIFY_TEST_1_ID: int
 TESTING_MERGIFY_TEST_2_ID: int
 TESTING_GPGKEY_SECRET: bytes
 TESTING_ID_GPGKEY_SECRET: str
+TESTING_INSTALLATION_ID: int
 SAAS_MODE: bool
 
 configuration_file = os.getenv("MERGIFYENGINE_TEST_SETTINGS")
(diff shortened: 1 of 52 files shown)


mergify_engine/config.py
```diff
@@ -218,9 +218,9 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
             "TESTING_FORWARDER_ENDPOINT",
             default="https://test-forwarder.mergify.io/events",
         ): str,
-        voluptuous.Required("INSTALLATION_ID", default=15398551): voluptuous.Coerce(
-            int
-        ),
+        voluptuous.Required(
+            "TESTING_INSTALLATION_ID", default=15398551
+        ): voluptuous.Coerce(int),
```
jd, line 223: unrelated?

mergify_engine/redis_utils.py
```diff
@@ -13,17 +13,28 @@
...
+import yaaredis
 
-from mergify_engine import utils
+from mergify_engine import config
 
 
 LOG = daiquiri.getLogger(__name__)
 
+
+# NOTE(sileht): I wonder with mypy thing yaaredis.StrictRedis is Any...
```
jd, line 33:
```suggestion
# NOTE(sileht): I wonder why mypy thinks `yaaredis.StrictRedis` is `Any`...
```

mergify_engine/web/api/security.py
```diff
@@ -100,11 +100,8 @@ async def get_repository_context(
     repository: github_types.GitHubRepositoryName = fastapi.Path(  # noqa: B008
         ..., description="The name of the repository"
     ),
-    redis_cache: utils.RedisCache = fastapi.Depends(  # noqa: B008
-        redis.get_redis_cache
-    ),
-    redis_queue: utils.RedisQueue = fastapi.Depends(  # noqa: B008
-        redis.get_redis_queue
+    redis_links: redis_utils.RedisQueue = fastapi.Depends(  # noqa: B008
```
Syffe, line 103:
```suggestion
    redis_links: redis_utils.RedisLinks = fastapi.Depends(  # noqa: B008
```

jd (changes requested):

mergify_engine/config.py
sileht: I typed all tests using redis_XXX and some was complaining about missing types, when I added the type for INSTALLATION_ID I prefixed it with TESTING_ as it's testing only thing.

mergify_engine/web/api/security.py
sileht: Hum, wierd that mypy didn't cached that one...

jd (approved):

PR #4647, opened 2022-05-03
fix(worker): worker task cancellation during sleeping

sileht:
If task.cancel() was called during a work sleep, this one was ignoring
the cancellation, instead of quitting the loop.

This change fixes that.

Fixes MRGFY-1139

Change-Id: Idf48d35d3821af6227e6c3dc426f4762a2e3567e

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -392,6 +392,11 @@ class StreamProcessor:
                 if sub.has_feature(subscription.Features.DEDICATED_WORKER):
                     if self.dedicated_owner_id is None:
                         # Spawn a worker
+                        LOG.info(
+                            "shared worker got an event for dedicated worker, adding org to Redis dedicated workers set",
+                            owner_id=owner_id,
+                            owner_login=owner_login_for_tracing,
+                        )
                         await self.redis_links.stream.sadd(
                             DEDICATED_WORKERS_KEY, owner_id
                         )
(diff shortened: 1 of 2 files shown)


mergify_engine/worker.py
```diff
@@ -392,13 +392,23 @@ async def consume(
                 if sub.has_feature(subscription.Features.DEDICATED_WORKER):
                     if self.dedicated_owner_id is None:
                         # Spawn a worker
+                        LOG.info(
+                            "shared worker got an event for dedicated worker, adding org to Redis dedicated workers set.",
```
jd, line 396:
```suggestion
                            "shared worker got an event for dedicated worker, adding org to Redis dedicated workers set",
```

mergify_engine/worker.py
```diff
@@ -392,13 +392,23 @@ async def consume(
...
+                        )
                         await self.redis_links.stream.sadd(
                             DEDICATED_WORKERS_KEY, owner_id
                         )
                         return
                 else:
                     if self.dedicated_owner_id is not None:
                         # Drop this worker
+                        LOG.info(
+                            "dedicated worker got an event for shared worker, removing org from Redis dedicated workers set.",
```
jd, line 408:
```suggestion
                            "dedicated worker got an event for shared worker, removing org from Redis dedicated workers set",
```

jd (dismissed):

jd:
@mergifyio squash

jd (approved):

PR #4656, opened 2022-05-03
feat(queue): add queue configuration list API endpoint

Syffe:
In order to implement a new dashboard UI queue listing, we needed
an engine API endpoint that list all the queues configuration of a repository

Fixes MRGFY-1137


mergify_engine/web/api/queues.py
```diff
@@ -315,6 +323,33 @@ async def repository_queues(
...
+    ),
+) -> QueueNames:
+
+    config_file = await repository_ctxt.get_mergify_config_file()
+    if config_file is None:
+        raise fastapi.HTTPException(
+            status_code=404, detail="Mergify configuration file is missing."
+        )
+
+    config = get_mergify_config(config_file)
```
jd, line 348: rather return an empty list if no config

jd (dismissed):

mergify_engine/web/api/queues.py
```diff
@@ -315,6 +323,30 @@ async def repository_queues(
...
 
 
+@router.get(
+    "/repos/{owner}/{repository}/queues/names",  # noqa: FS003
+    summary="Get merge queue names",
+    description="Get the list of all merge queue names",
+    response_model=QueueNames,
+    responses={
+        **api.default_responses,  # type: ignore
+        404: {"description": "There is no queues configured for this repository"},
```
jd, line 333: hm?

jd (changes requested):

mergify_engine/web/api/queues.py
```diff
@@ -315,6 +323,29 @@ async def repository_queues(
     return queues
 
 
+@router.get(
+    "/repos/{owner}/{repository}/queues/names",  # noqa: FS003
```
jd, line 327: What about something like `/config` that returns the whole configuration of the queues, not just the names?

Syffe: If there is a need indeed it would be great ! Tho for the new dashboard UI design we only need the name, we could also integrate an info-bubble or something that shows the config for each queues on their respective tables section ?

Syffe: what would be the use case for a user to use this endpoint ?

jd: Yes, I think it'd be valuable to know that queue has spec check enabled for example, and show it in the dashboard.
And having one single endpoint and one single call to do is going to be more efficient.

mergify_engine/web/api/queues.py
```diff
@@ -315,6 +323,38 @@ async def repository_queues(
     return queues
 
 
+@router.get(
+    "/repos/{owner}/{repository}/queues/configurations",  # noqa: FS003
```
jd, line 327:
```suggestion
    "/repos/{owner}/{repository}/queues/configuration",  # noqa: FS003
```

mergify_engine/web/api/queues.py
```diff
@@ -315,6 +323,38 @@ async def repository_queues(
...
+    summary="Get merge queues configurations",
```
jd, line 328:
```suggestion
    summary="Get merge queues configuration",
```

The configuration of the queues :)

releasenotes/notes/queue-config-list-81d16c577c855672.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Added a new API endpoint (``/api/repos/<owner>/<repo>/queues/configurations``) that allows to query
```
jd, line 4:
```suggestion
    Added a new API endpoint (``/api/repos/<owner>/<repo>/queues/configuration``) that allows to query
```

mergify_engine/tests/functional/api/test_queue.py
```diff
@@ -0,0 +1,148 @@
...
+        r = await self.app.get(
+            f"/v1/repos/{config.TESTING_ORGANIZATION_NAME}/{self.RECORD_CONFIG['repository_name']}/queues/configurations",
+            headers={
+                "Authorization": f"bearer {self.api_key_admin}",
+                "Content-type": "application/json",
+            },
+        )
+
+        assert r.status_code == 200
+        assert r.json() == {"configurations": []}
```
jd, line 93:
```suggestion
        assert r.json() == {"configuration": []}
```

mergify_engine/tests/functional/api/test_queue.py
```diff
@@ -0,0 +1,148 @@
...
+                        f"base={self.main_branch_name}",
+                        "label=queue-low",
+                    ],
+                    "actions": {"queue": {"name": "low-priority"}},
+                },
+            ],
+        }
+
+        r = await self.app.get(
+            f"/v1/repos/{config.TESTING_ORGANIZATION_NAME}/{self.RECORD_CONFIG['repository_name']}/queues/configurations",
```
jd, line 85:
```suggestion
            f"/v1/repos/{config.TESTING_ORGANIZATION_NAME}/{self.RECORD_CONFIG['repository_name']}/queues/configuration",
```

mergify_engine/tests/functional/api/test_queue.py
```diff
@@ -0,0 +1,148 @@
...
+            },
+        )
+
+        assert r.status_code == 200
+        assert r.json() == {"configurations": []}
+
+        await self.setup_repo(yaml.dump(rules))
+
+        r = await self.app.get(
+            f"/v1/repos/{config.TESTING_ORGANIZATION_NAME}/{self.RECORD_CONFIG['repository_name']}/queues/configurations",
```
jd, line 98:
```suggestion
            f"/v1/repos/{config.TESTING_ORGANIZATION_NAME}/{self.RECORD_CONFIG['repository_name']}/queues/configuration",
```

jd (dismissed):

mergify_engine/web/api/queues.py
```diff
@@ -84,14 +84,14 @@ class SpeculativeCheckPullRequest:
...
-        metadata={"description": "The name of the queue rule"}
-    )
-
     config: rules.QueueConfig = dataclasses.field(
         metadata={"description": "The configuration of the queue rule"}
     )
 
+    name: rules.QueueName = dataclasses.field(
+        metadata={"description": "The name of the queue rule"}
+    )
```
jd, line 93: Why moving this?

jd (commented):
@Syffe LGTM except why you moved a piece of code?

mergify_engine/web/api/queues.py
Syffe: alphabetic order in class definition for maintainability

jd: I don't think it's a good idea: it just moves the code, break tools like `git blame` and it's not enforce by any automated check. So that's never going to stay like this forever. 😞

jd (approved):

PR #4691, opened 2022-05-09
refactor(signals): rework signal interface

sileht:
This should helps to type event logs subsystem

Fixes #4691

### DIFF
--- a/mergify_engine/signals.py
+++ b/mergify_engine/signals.py
@@ -19,24 +19,28 @@ import typing
 
 import daiquiri
 
-from mergify_engine import context
+from mergify_engine import github_types
+
+
+if typing.TYPE_CHECKING:
+    from mergify_engine import context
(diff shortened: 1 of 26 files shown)


mergify_engine/signals.py
```diff
@@ -59,9 +102,10 @@ class SignalBase(abc.ABC):
     @abc.abstractmethod
     async def __call__(
         self,
-        ctxt: context.Context,
+        repository: "context.Repository",
+        pull_request: github_types.GitHubPullRequestNumber,
```
jd, line 106: What about passing the whole PR structure?

sileht: For now, we track only think where we have the full context, but when I will add signals in Train and TrainCar, we may not have them.

jd (approved):

PR #4710, opened 2022-05-12
fix(queue): improve inplace checks default

sileht:
This change avoid creating draft PR for nothing by default is the PR is
already up to date.

Fixes MRGFY-865

Change-Id: Ib9bd09f5192dc6caa0a925d49c83b05bbb7875a8


jd (changes requested):
doc update?

jd (approved):

PR #4733, opened 2022-05-17
fix(check-runs): never cancel a check-run

sileht:
This code is obsolete, logs show we didn't fall into that case anymore.

But this doesn't fix the root cause of MRGFY-1128 as the logs show
Context.pull_engine_check_runs have duplicate check-run ids. More
logs are added to try to understand how this can happen.

Fixes MRGFY-1128

### DIFF
--- a/mergify_engine/check_api.py
+++ b/mergify_engine/check_api.py
@@ -282,20 +282,14 @@ async def set_check_run(
 
     if len(checks) >= 2:
         ctxt.log.warning(
-            "pull requests with duplicacte checks", checks=checks, skip_cache=skip_cache
+            "pull requests with duplicate checks",
+            checks=checks,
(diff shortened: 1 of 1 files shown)


mergify_engine/check_api.py
```diff
@@ -282,21 +282,15 @@ async def set_check_run(
 
     if len(checks) >= 2:
         ctxt.log.warning(
-            "pull requests with duplicacte checks", checks=checks, skip_cache=skip_cache
+            "pull requests with duplicacte checks",
```
jd, line 285:
```suggestion
            "pull requests with duplicate checks",
```

jd (dismissed):

jd:
@mergifyio squash

PR #4784, opened 2022-05-31
refactor(redis): split team/user permissions/members cache

sileht:
Fixes MRGFY-1197

Change-Id: I81625d6afe22da5a89ea6e53973155a15e944ea0

### DIFF
--- a/mergify_engine/config.py
+++ b/mergify_engine/config.py
@@ -197,9 +197,26 @@ Schema = voluptuous.Schema(
         voluptuous.Required(
             "REDIS_QUEUE_WEB_MAX_CONNECTIONS", default=None
         ): voluptuous.Any(None, voluptuous.Coerce(int)),
-        voluptuous.Required("STORAGE_URL", default="redis://localhost:6379?db=8"): str,
+        # NOTE(sileht): Unused anymore, but keep to detect legacy onpremise installation
+        voluptuous.Required("STORAGE_URL", default=None): voluptuous.Any(None, str),
+        # NOTE(sileht): Not used directly, but used to build other redis urls if not provided
+        voluptuous.Required("DEFAULT_REDIS_URL", default=None): voluptuous.Any(
+            None, str
+        ),
+        voluptuous.Required("LEGACY_CACHE_URL", default=None): voluptuous.Any(
+            None, str
+        ),
         voluptuous.Required("QUEUE_URL", default=None): voluptuous.Any(None, str),
         voluptuous.Required("STREAM_URL", default=None): voluptuous.Any(None, str),
+        voluptuous.Required("TEAM_MEMBERS_CACHE_URL", default=None): voluptuous.Any(
+            None, str
+        ),
+        voluptuous.Required("TEAM_PERMISSIONS_CACHE_URL", default=None): voluptuous.Any(
+            None, str
+        ),
+        voluptuous.Required("USER_PERMISSIONS_CACHE_URL", default=None): voluptuous.Any(
+            None, str
+        ),
         voluptuous.Required("SHARED_STREAM_PROCESSES", default=1): voluptuous.Coerce(
             int
         ),
@@ -283,9 +300,14 @@ SHARED_STREAM_TASKS_PER_PROCESS: int
 EXTERNAL_USER_PERSONAL_TOKEN: str
 BOT_USER_ID: int
 BOT_USER_LOGIN: str
-STORAGE_URL: str
+
 STREAM_URL: str
 QUEUE_URL: str
+LEGACY_CACHE_URL: str
+TEAM_PERMISSIONS_CACHE_URL: str
+TEAM_MEMBERS_CACHE_URL: str
+USER_PERMISSIONS_CACHE_URL: str
+
 BUCKET_PROCESSING_MAX_SECONDS: int
 INTEGRATION_ID: int
 SUBSCRIPTION_BASE_URL: str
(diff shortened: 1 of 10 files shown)


sileht:
Hope I didn't miss anything, the mypy typing didn't work, you can use the wrong Redis database and the yaaredis alias type didn't cache it. With redis-py it will works.

mergify_engine/config.py
```diff
@@ -359,6 +371,15 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
...
     QUEUE_URL = globals()["STORAGE_URL"]
 
+if globals()["TEAM_MEMBERS_CACHE_URL"] is None:
+    TEAM_MEMBERS_CACHE_URL = globals()["STORAGE_URL"]
+
+if globals()["TEAM_PERMISSIONS_CACHE_URL"] is None:
+    TEAM_PERMISSIONS_CACHE_URL = globals()["STORAGE_URL"]
+
+if globals()["USER_PERMISSIONS_CACHE_URL"] is None:
+    USER_PERMISSIONS_CACHE_URL = globals()["STORAGE_URL"]
```
jd, line 381: Do we want to stop doing this so we make sure to not make the same "mistake" again with onprem in the future?

mergify_engine/redis_utils.py
```diff
@@ -119,6 +122,9 @@ class RedisLinks:
     cache_max_connections: typing.Optional[int] = None
     stream_max_connections: typing.Optional[int] = None
     queue_max_connections: typing.Optional[int] = None
+    team_permissions_cache_max_connections: typing.Optional[int] = None
```
jd, line 125: there's no settings exposed in the config

sileht: I dropped them and used the cache_max_connections

mergify_engine/config.py
sileht: I proposed an alternative that should be compatible with onpremise. If it looks good to you, I will add some tests to ensure the migration will works well.

mergify_engine/config.py
```diff
@@ -353,11 +376,49 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
 
 globals().update(Schema(CONFIG))
 
-if globals()["STREAM_URL"] is None:
-    STREAM_URL = globals()["STORAGE_URL"]
+# NOTE(sileht): legacy on premise installation before thing was auto sharded in redis databases
+# before on on premise STREAM/QUEUE/LEGACY_CACHE was in the same db, so keep until manuall migration
+# as been done
```
jd, line 381:
```suggestion
# NOTE(sileht): legacy on-premise installation before things were auto sharded in redis databases
# before on-premise STREAM/QUEUE/LEGACY_CACHE was in the same db, so keep until manual migration
# as been done
```
there twice "before" so I'm not sure I understand the comment TBH

mergify_engine/config.py
```diff
@@ -353,11 +376,49 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
...
+    "LEGACY_CACHE_URL": 2,
+    "STREAM_URL": 3,
+    "QUEUE_URL": 4,
+    "TEAM_MEMBERS_CACHE_URL": 5,
+    "TEAM_PERMISSIONS_CACHE_URL": 6,
+    "USER_PERMISSIONS_CACHE_URL": 7,
+}
+
+default_redis_url_parsed = parse.urlparse(globals()["DEFAULT_REDIS_URL"])
+if default_redis_url_parsed.query and "db" in parse.parse_qs(
```
jd, line 408:
```suggestion
if default_redis_url_parsed.query and "db=" in parse.parse_qs(
```
?

mergify_engine/config.py
```diff
@@ -353,11 +376,49 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
...
+    default_redis_url_parsed.query
+):
+    print("DEFAULT_REDIS_URL must not contains db argument. Mergify can't start.")
```
jd, line 411:
```suggestion
    print("DEFAULT_REDIS_URL must not contain any db parameter. Mergify can't start.")
```

mergify_engine/config.py
```diff
@@ -197,9 +197,25 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
         voluptuous.Required(
             "REDIS_QUEUE_WEB_MAX_CONNECTIONS", default=None
         ): voluptuous.Any(None, voluptuous.Coerce(int)),
-        voluptuous.Required("STORAGE_URL", default="redis://localhost:6379?db=8"): str,
+        # NOTE(sileht): Unused anymore, but keep to detect legacy onpremise installation
+        voluptuous.Required("STORAGE_URL", default=None): voluptuous.Any(None, str),
+        voluptuous.Required(
+            "DEFAULT_REDIS_URL", default="redis://localhost:6379?db=8"
```
jd, line 203:
```suggestion
            "DEFAULT_REDIS_URL", default="redis://localhost:6379"
```
or just don't set any default; the current one is going to fail.

mergify_engine/config.py
```diff
@@ -353,11 +376,49 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
...
+        "LEGACY_CACHE_URL",
+    ):
+        if globals()[config_key] is None:
+            globals()[config_key] = globals()["STORAGE_URL"]
+
+
+# NOTE(sileht): If we reach 15, we should update onpremise installation guide
+# and add an upgrade release note section to ensure people configure their Redis
+# correctly
+REDIS_AUTO_DB_SHARDING_MAPPING = {
```
jd, line 396: I like the idea though OTOH this is not going to help if we ever decide to move a DB to a different Redis server 😢

jd (dismissed):

mergify_engine/config.py
sileht: parse_qs returns a dict

mergify_engine/config.py
sileht: It just auto shard the default Redis, you can still move a particular database to another instance by settings explicitly QUEUE_URL or TEAM_MEMBERS_CACHE_URL

mergify_engine/logs.py
```diff
@@ -95,6 +95,11 @@ def config_log() -> None:
     LOG.info("* PATH: %s", os.environ.get("PATH"))
     LOG.info("##########################################################")
 
+    if config.STORAGE_URL is not None:
+        LOG.warning(
+            "MERGIFYENGINE_STORAGE_URL is set, on-premise legacy redis database setup detected."
```
jd, line 100:
```suggestion
            "MERGIFYENGINE_STORAGE_URL is set, on-premise legacy Redis database setup detected."
```

jd (changes requested):
@sileht can we add a release note with upgrade instruction for onprem?

mergify_engine/config.py
sileht: I imagined another mechanism where we can map things like:

```
redis_servers:
  default: rediss://foo:1234
  user-defined-name: rediss://foo:1234
  queue: rediss://fish-and-chips:1234
  cache: rediss://bar-cache:1234

redis_db_mapping: // We provide defaults where server is always set to "default" and the default db is chosen by us.
  queue: 
     server: queue
     db: 3 
  legacy_cache: 
     server: cache
     db: 3
  team_members_cache: 
     server: cache
     db: 1
```

But I didn't find fancy way to have this in environment variables that works with Heroku.

jd (approved):

PR #4785, opened 2022-06-01
fix(queue): handle queue config API code 500 when config file is invalid

Syffe:
queue configuration API wasn't handling the case where a config file requested is invalid,
thus returning 500 INTERNAL SERVER ERROR. We know handle the case and return a proper payload.

Fixes MRGFY-1200


mergify_engine/web/api/queues.py
```diff
@@ -344,7 +345,13 @@ async def repository_queues_configuration(
     if config_file is None:
         return QueuesConfig()
 
-    config = get_mergify_config(config_file)
+    try:
+        config = get_mergify_config(config_file)
+    except InvalidRules:
+        raise fastapi.HTTPException(
+            status_code=409,
+            detail="The configuration file rules are invalid. Please verify they are valid or up to date.",
```
jd, line 353:
```suggestion
            detail="The configuration file is invalid.",
```

keep it simple.

mergify_engine/web/api/queues.py
```diff
@@ -344,7 +345,13 @@ async def repository_queues_configuration(
     if config_file is None:
         return QueuesConfig()
 
-    config = get_mergify_config(config_file)
+    try:
+        config = get_mergify_config(config_file)
+    except InvalidRules:
+        raise fastapi.HTTPException(
+            status_code=409,
```
jd, line 352: You need to add this return type in the documentation (see decorator).

jd (changes requested):

mergify_engine/web/api/queues.py
sileht: Shouldn't be 422?

Syffe: Indeed 409 is more for a conflict in the request, in this case the request is good, only the ressource/entity has conflict

jd (approved):

PR #4801, opened 2022-06-03
fix(ddtrace): workaround for missing ddtrace pinning

sileht:
Workaround for https://github.com/DataDog/dd-trace-py/pull/3793

Change-Id: Ibd8ce535be54bd186825127173007f2746f29a2a

### DIFF
--- a/mergify_engine/__init__.py
+++ b/mergify_engine/__init__.py
@@ -1,0 +1,5 @@
+import ddtrace.pin
+import redis.asyncio
+
+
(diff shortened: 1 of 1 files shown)


jd (commented):
Do you want to actually set a pin with a different service name for each redis connection? would be a great thing and a "better" workaround 😉

sileht:
> Do you want to actually set a pin with a different service name for each redis connection? would be a great thing and a "better" workaround 😉

We should add the client_name to the tags, but I don't understand the pin api

jd (approved):

PR #4802, opened 2022-06-03
refactor(service): use MERGIFYENGINE_VERSION to detect version

jd:
This must be passed as an environment variable to have the version set.
This removes the dependency on Heroku and leverage something more useful
such as the sha1.

Fixes MRGFY-1206


sileht (approved):

sileht:
@mergifyio requeue

PR #4804, opened 2022-06-03
refactor(config): make it testable

sileht:
This refactor the config.py to be able to test it.

Change-Id: I592d5b869a3fa3e74efafaf9dbe22b2c758c9334

### DIFF
--- a/mergify_engine/config.py
+++ b/mergify_engine/config.py
@@ -261,7 +261,7 @@ Schema = voluptuous.Schema(
     }
 )
 
-# Config variables available
+# Config variables available from voluptuous
 VERSION: str
 BASE_URL: str
 API_ENABLE: bool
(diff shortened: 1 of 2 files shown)


mergify_engine/config.py
```diff
@@ -325,56 +325,65 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
...
+    configuration_file = os.getenv("MERGIFYENGINE_TEST_SETTINGS")
 
+    if configuration_file is not None:
+        dotenv.load_dotenv(dotenv_path=configuration_file, override=True)
 
-# DASHBOARD API KEYS are required only for Saas
-if "SUBSCRIPTION_TOKEN" in CONFIG:
-    for key in ("DASHBOARD_TO_ENGINE_API_KEY", "ENGINE_TO_DASHBOARD_API_KEY"):
-        CONFIG[key] = secrets.token_hex(16)
+    raw_config: typing.Dict[str, typing.Any] = {}
```
jd, line 338: do really need `Any` here? It seems everything is a `str` in keys.

mergify_engine/config.py
```diff
@@ -325,56 +325,65 @@ def ApplicationAPIKeys(v: str) -> typing.Dict[str, ApplicationAPIKey]:
...
-if legacy_api_url is not None:
-    if legacy_api_url[-1] == "/":
-        legacy_api_url = legacy_api_url[:-1]
-    if legacy_api_url.endswith("/api/v3"):
-        CONFIG["GITHUB_REST_API_URL"] = legacy_api_url
-        CONFIG["GITHUB_GRAPHQL_API_URL"] = f"{legacy_api_url[:-3]}/graphql"
+    # DASHBOARD API KEYS are required only for Saas
+    if "SUBSCRIPTION_TOKEN" in raw_config:
+        for key in ("DASHBOARD_TO_ENGINE_API_KEY", "ENGINE_TO_DASHBOARD_API_KEY"):
+            raw_config[key] = secrets.token_hex(16)
```
jd, line 347: This is confusing because I imagine this does not work then and is useless?
Might be better to use a static string like `unused`?

mergify_engine/config.py
sileht: voluptuous is untyped, so... if I put something have to ignore it everywhere after.

mergify_engine/config.py
sileht: The thing is that the web servers' endpoints are always mounted, so I put random secrets so nobody can use them on-premise.

Known that we have `SAAS_MODE` boolean I thing we can do better. I created a ticket to fix this.

jd (approved):

PR #4822, opened 2022-06-08
feat(engine): allow mergify app to be pinged with its url

Greesb:
Fixes MRGFY-1217

### DIFF
--- a/mergify_engine/engine/commands_runner.py
+++ b/mergify_engine/engine/commands_runner.py
@@ -37,7 +37,10 @@ from mergify_engine.rules import conditions
 
 LOG = daiquiri.getLogger(__name__)
 
-COMMAND_MATCHER = re.compile(r"^@Mergify(?:|io) (\w*)(.*)", re.IGNORECASE)
+COMMAND_MATCHER = re.compile(
+    fr"^(?:{config.GITHUB_URL.rstrip('/')}/|@)Mergify(?:|io) (\w*)(.*)",  # noqa: BLK100
+    re.IGNORECASE,
+)
 COMMAND_RESULT_MATCHER_OLD = re.compile(
     r"\*Command `([^`]*)`: (pending|success|failure)\*"
 )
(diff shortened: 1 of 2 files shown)


mergify_engine/engine/commands_runner.py
```diff
@@ -37,7 +37,7 @@
 
 LOG = daiquiri.getLogger(__name__)
 
-COMMAND_MATCHER = re.compile(r"^@Mergify(?:|io) (\w*)(.*)", re.IGNORECASE)
+COMMAND_MATCHER = re.compile(r"^(?:https://github.com/|@)Mergify(?:|io) (\w*)(.*)", re.IGNORECASE)
```
jd, line 40: Hum I think `:?` makes the URL prefix or the `@` optional. I'd rather keep them mandatory.

jd (changes requested):

mergify_engine/engine/commands_runner.py
Greesb: No the `:?` is just to indicate that its a non-capturing group

jd (commented):
@sileht That looks good to me, but I imagine it won't work with GHES.
Do we want to make that work for GHES too?

We already don't look into the right username though.

mergify_engine/engine/commands_runner.py
```diff
@@ -37,7 +37,10 @@
 
 LOG = daiquiri.getLogger(__name__)
 
-COMMAND_MATCHER = re.compile(r"^@Mergify(?:|io) (\w*)(.*)", re.IGNORECASE)
+COMMAND_MATCHER = re.compile(
+    r"^(?:https://github.com/|@)Mergify(?:|io) (\w*)(.*)",
```
sileht, line 41: We should use config.GITHUB_URL instead of the hardcoded URL to make it work with GHES

sileht (dismissed):

sileht:
> @sileht That looks good to me, but I imagine it won't work with GHES. Do we want to make that work for GHES too?

We have the domain in the configuration

> We already don't look into the right username though.

People use alreadyy @mergify on GHES and this account doesn't exists on them, so... I'm not sure we can do better.

mergify_engine/engine/commands_runner.py
Greesb: Done @sileht

sileht (approved):

PR #4866, opened 2022-06-18
fix(seats-count): allow to use a dedicated db

sileht:
Fixes MRGFY-819

Change-Id: Ided4b76e680140b57d1c0afb772c4253922676ed

### DIFF
--- a/mergify_engine/config.py
+++ b/mergify_engine/config.py
@@ -220,6 +220,9 @@ Schema = voluptuous.Schema(
         voluptuous.Required("USER_PERMISSIONS_CACHE_URL", default=None): voluptuous.Any(
             None, str
         ),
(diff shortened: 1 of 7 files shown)


sileht:
I didn't though how to move data to the new DB yet.

jd (changes requested):
There seems to be no plan for migration?

jd (approved):

sileht:
Lua scripts to move the data:

```
redis.call("SELECT", 2)
local matches = redis.call('KEYS', 'active-users~*')
local total = 0
local success = 0
for _,key in ipairs(matches) do
    local ret = redis.call('COPY', key, key, "DB", 8, "REPLACE")
    success = success + tonumber(ret)
    total = total + 1
end
return "Number of keys migrated: " .. success .. "/" .. total
```

Old data will expire itself.

sileht:
@mergifyio rebase

PR #4869, opened 2022-06-19
fix(worker): handle too big pull request

sileht:
Fixes MRGFY-233
Fixes MERGIFY-ENGINE-2PW

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -241,13 +241,28 @@ async def run_engine(
                 wait_background_github_processing=True,
             )
         except http.HTTPNotFound:
-            # NOTE(sileht): Don't fail if we received even on repo/pull that doesn't exists anymore
+            # NOTE(sileht): Don't fail if we received an event on repo/pull that doesn't exists anymore
             logger.debug("pull request doesn't exists, skipping it")
-            return None
+            return
+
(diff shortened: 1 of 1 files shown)


mergify_engine/worker.py
```diff
@@ -243,11 +243,26 @@ async def run_engine(
...
+                and "The request could not be processed because too many files changed"
+                in e.message
+            ):
+                logger.warning(
+                    "This pull request cannot be evaluated by Mergify", exc_info=True
+                )
+                return
+
+            # NOTE(sileht): Reraise to retry with worker or create a Sentry
+            # issue
```
jd, line 260: how a Sentry issue can be useful here?

sileht: Mainly to track and learn what kind of GitHub issue we can have here and not blindly ignore all 4XX.

mergify_engine/worker.py
```diff
@@ -243,11 +243,26 @@ async def run_engine(
         except http.HTTPNotFound:
             # NOTE(sileht): Don't fail if we received even on repo/pull that doesn't exists anymore
```
Syffe, line 244:
```suggestion
            # NOTE(sileht): Don't fail if we received an event on repo/pull that doesn't exists anymore
```

PR #4870, opened 2022-06-19
fix(delayed-refresh): decrease priority of delayed-refresh

sileht:
Currently when a time conditions will change its status, we will refresh
all opened pull refresh to reflect that in the Summary.

This meantime that at certain hours Mergify may looks laggy.

Delayed refresh doesn't really need to be on time, we already
document that delayed-refresh can be 5 minutes late, so.

This introduces a new level of priority: medium. By default GitHub events
stay high as before. delayed-refresh are now medium. And pull request
refresh due to push events stay low.

Fixes MRGFY-1234

### DIFF
--- a/mergify_engine/worker.py
+++ b/mergify_engine/worker.py
@@ -46,6 +46,7 @@ import collections
 import contextlib
 import dataclasses
 import datetime
+import enum
 import functools
 import hashlib
 import itertools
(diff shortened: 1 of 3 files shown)


mergify_engine/worker.py
```diff
@@ -142,11 +142,35 @@ class UnexpectedPullRetry(Exception):
...
+
+
+def get_low_priority_score() -> float:
+    return date.utcnow().timestamp() * 100
+
+
+def get_priority_level_from_score(
+    score: float,
+) -> typing.Literal["high", "medium", "low"]:
+    low_prio_threshold = get_low_priority_score() * 2
```
Greesb, line 166: So if the score is has just been set to `date.utcnow().timestamp() * 100` (low priority score), when going through this function it will have a priority level of "medium". I might be missing something but i don't see any cases where an event could have a level of "low" in this, is this what is wanted ?

jd: I'm a bit concerned also by the lack of unit tests here. 😄

sileht: What is stored in the ordered set is not the list of events, but just the PR to refresh. The events list is stored in another Redis key (one by PR). 
Here is a beautiful ASCII art of the data structures: https://github.com/Mergifyio/mergify-engine/blob/main/mergify_engine/worker.py#L15-L40 :D

So if a PR is already in the set with a higher score (a lower prior), and a new event comes for the same PR but with a lower score (higher prior), we add this event in the other key and just override the score (the position) in the set.

If we use 3 lists/sets for this, if the priority to refreshing a PR change, we need to clean up the two other lists/sets.

But even if there is a little overhead as maintaining 3 lists/sets instead of 1, maybe this is worth it for readability... What does other think?

I'm just a bit fearful to change how the priority system currently works (Lua scripts and the polling system), which is a bigger task.

Here, I just wanted to introduce a new level of priority.

Syffe: Using an offset+divider already makes priorities clearer to handle and read. Like @sileht said, further changes would lead to a bigger rework than what we actually need for this PR to fit, even though leveraging sets is a good idea for handling priorities!

Greesb: Yeah but in 5138 Mergify would go boom with this solution :(

mergify_engine/tests/functional/commands/test_copy.py
```diff
@@ -65,9 +65,9 @@ async def test_command_copy(self) -> None:
         ]
 
         await self.merge_pull(pulls_feature[0]["number"])
-        await self.wait_for("pull_request", {"action": "closed"})
         await self.merge_pull(pulls_stable[0]["number"])
         await self.wait_for("pull_request", {"action": "closed"})
+        await self.wait_for("pull_request", {"action": "closed"})
```
jd, line 70: different PR?

mergify_engine/worker.py
```diff
@@ -142,11 +142,36 @@ class UnexpectedPullRetry(Exception):
...
-    # 2 to split the bucket
-    return date.utcnow().timestamp() * 2
+PriorityT = typing.Literal["high", "medium", "low"]
+
+PRIORITY_BY_NAMES: typing.Dict[PriorityT, int] = {
+    "high": 1,
+    "medium": 3,
+    "low": 5,
+}
+PRIORITY_BY_SCORES = {v: k for k, v in PRIORITY_BY_NAMES.items()}
```
jd, line 152: use `enum.Enum`?

mergify_engine/tests/functional/commands/test_copy.py
sileht: useless.

sileht:
@mergifyio rebase

PR #4881, opened 2022-06-20
feat(tools): Add a custom version of updatecli

Greesb:
There are not as many features as the original updatecli, but it is able
to handle unicode characters correctly.

Fixes MRGFY-1052


.github/get-latest-python-version.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_superior_version_tag(tag1: str, tag2: str) -> str:
+    tag1_split = tag1.split(".")
+    tag2_split = tag2.split(".")
+    for i in range(len(tag1_split)):
+        if int(tag1_split[i]) > int(tag2_split[i]):
+            return tag1
+        elif int(tag1_split[i]) < int(tag2_split[i]):
+            return tag2
```
jd, line 15: Might be safer to use https://packaging.pypa.io/en/latest/version.html#packaging.version.parse ?

.github/get-latest-python-version.py
```diff
@@ -0,0 +1,41 @@
...
+    json_tags = json.load(tags_request)["results"]
+
+    latest_tag = None
+    for tag in json_tags:
+        if re.fullmatch(r"^\d+\.\d+\.\d+-slim", tag["name"]):
+            tag_name = tag["name"].replace("-slim", "")
+            if latest_tag is None:
+                latest_tag = tag_name
+            else:
+                latest_tag = get_superior_version_tag(tag_name, latest_tag)
```
jd, line 32: or simply `sort` the tag list and take the first one, might be easier to read :)

tools/updatecli.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python3
+
```
jd, line 2: Might be better to use a different name than `updatecli` or it might be confusing.

jd (changes requested):

.github/get-latest-python-version.py
Greesb: Indeed, done !

tools/updatecli.py
Greesb: Renamed it but i was not really inspired

jd (dismissed):
What's the story if we want to use this in another repo?

tools/update_files_from_cli.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def get_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(description="UpdateCLI")
+
+    parser.add_argument(
+        "-c",
+        "--config",
+        action="store",
+        type=file_type,
```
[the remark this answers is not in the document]
Greesb: Not really because `argparse.FileType` requires a mode and directly open that file with that mode, but in the script i need to open it at different times with different modes.

tools/update_files_from_cli.py
```diff
@@ -0,0 +1,226 @@
...
+    path = os.path.expanduser(filepath)
+    if not os.path.isfile(path):
+        raise argparse.ArgumentTypeError(
+            f"The specified file path '{filepath}' is invalid"
+        )
+    return path
+
+
+def get_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(description="UpdateCLI")
```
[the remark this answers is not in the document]
Greesb: A real name for what ? You mean the `prog` argument for the ArgumentParser ?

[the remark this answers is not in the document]
jd: Yes and for the tool 😉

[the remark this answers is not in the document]
Greesb: Yeah but naming is hard 😢

Greesb:
> What's the story if we want to use this in another repo?

@jd I don't know, do we ? I can put it in another repo or even put it on pip if that's what we want, but i've never uploaded something on pip so it might take some time.

tools/update_files_from_cli.py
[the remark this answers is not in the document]
Greesb: Came up with `clifus` for `CLI File Update Strategy`. It looks a bit like a dog name but i can't find anything better x)

tools/clifus.py
```diff
@@ -0,0 +1,229 @@
...
+        value = value.strip()
+        print(f"Value of source '{source_id}': {value}")
+        return value
+
+
+def get_replacer_value_with_transfomers(
+    source_value: typing.Any, transformers: list[dict[str, typing.Any]]
+) -> str:
+    """
+    Go through all the `transformers` indexs of the target to transform the value
```
Syffe, line 66:
```suggestion
    Go through all the `transformers` indexes of the target to transform the value
```

tools/clifus.py
```diff
@@ -0,0 +1,229 @@
...
+    we will write to the file
+    - find: Regex to apply to the source_value, all the other transformers will be applied to this if specified
```
Syffe, line 68:
```suggestion
    - find: Regex to apply to the source_value, will be applied to all transformers if specified
```

tools/clifus.py
```diff
@@ -0,0 +1,229 @@
...
+                transformer_dict["replacer"]["to"],
+            )
+
+        if "addPrefix" in transformer_dict:
+            value = f"{transformer_dict['addPrefix']}{value}"
+
+        if "addSuffix" in transformer_dict:
+            value = f"{value}{transformer_dict['addSuffix']}"
+
+    print(f"Transformed value : {value}")
```
Syffe, line 99:
```suggestion
    print(f"Transformed value: {value}")
```

tools/clifus.py
```diff
@@ -0,0 +1,229 @@
...
+    except FileExistsError:
+        pass
+
+    # TODO:
+    # For file.lineMatchPattern and yaml, gather all the targets modifying the same
+    # file to only have to rewrite the whole file once.
+    if target_config["kind"] == "file":
+        execute_file_target(target_name, target_config, source_value, output_file_path)
+    elif target_config["kind"] == "yaml":
+        # TODO
```
Syffe, line 180:
```suggestion
        # TODO(greesb)
```

.github/workflows/updatecli.yaml
```diff
@@ -15,31 +15,12 @@ jobs:
       - name: Checkout
```
sileht, line 15: You can enable the schedule L9 since it works now :)

sileht (changes requested):
Maybe we should rename the configuration file to `clifus.yml` too?

I would love to have a test and something that check the configuration file is valid (with pydantic for example).

But I don't like the idea to add a dependency. And the workflow that will run it daily may be enough as a test.

So just enabling the schedule and fixing the configuration file name, would be enough for me.

sileht (approved):

.github/workflows/clifus.yaml
```diff
@@ -0,0 +1,36 @@
+name: Clifus
+permissions: write-all
+
+on:
+  workflow_dispatch:
+    - cron: '0 0 * * *'
```
sileht, line 6:
```suggestion
    schedule:
    - cron: '0 0 * * *'
```

sileht (changes requested):

sileht:
@mergifyio squash

sileht (approved):

PR #4896, opened 2022-06-21
fix(worker): wait for redis to be up before starting worker

Syffe:
Worker was starting without asserting that Redis connections are up,
causing starting failures.

Fixes MRGFY-1225


mergify_engine/worker.py
```diff
@@ -907,6 +907,24 @@ def get_process_index_from_env() -> int:
...
+    redis_name: str,
+) -> None:
+    def retry_log(retry_state: tenacity.RetryCallState) -> None:
+        statsd.increment("redis.client.connection.errors")
+        LOG.warning(
+            f"Couldn't connect to Redis {redis_name}, retrying...",
+        )
+
+    r = tenacity.retry(
+        wait=tenacity.wait_exponential(multiplier=0.2),
```
sileht, line 921: We need to set a max= otherwise the exponential can become too big, like max 5 seconds

mergify_engine/worker.py
```diff
@@ -907,6 +907,24 @@ def get_process_index_from_env() -> int:
...
 
 
+async def ping_redis(
+    redis: typing.Union[redis_utils.RedisStream, redis_utils.RedisCache],
+    redis_name: str,
+) -> None:
+    def retry_log(retry_state: tenacity.RetryCallState) -> None:
+        statsd.increment("redis.client.connection.errors")
+        LOG.warning(
+            f"Couldn't connect to Redis {redis_name}, retrying...",
```
sileht, line 917: if we can tell in how many time we will retry :)
```suggestion
            f"Couldn't connect to Redis {redis_name}, retrying in X seconds...",
```

sileht (dismissed):

mergify_engine/worker.py
```diff
@@ -1160,6 +1184,9 @@ def get_shared_worker_ids(self) -> typing.List[int]:
     async def start(self) -> None:
         self._stopping.clear()
 
+        await ping_redis(self._redis_links.stream, "Stream")
+        await ping_redis(self._redis_links.cache, "Cache")
+
```
jd, line 1189: what if redis crashes right here?
is the following `load_scripts` going to retry too?

jd: Yes: and if you retry load_scripts "for ever" you basically have a better version of "ping redis" IMHO.

It's probably a bit orthogonal to this PR, but handling script loading dynamically might also be worth it.

Basically, making sure that if you restart Redis *at any point in time* Mergify falls back on its feet is a good way to think about the problem.

sileht: We were doing `load_script()` (note no ending 's') dynamically before but when `eval_sha()` is called from a pipeline it always succeeds and then an unrelated error is raised later when the pipeline is executed, making it hard to know which one is missing.

This is why we added this "load_scripts" to preload them on startup.

Maybe we can leverage `Client.on_connect()` to preload/reload them instead?

For migration, maybe we can just drop them for now, as all on-prem has run them.

I'm still in favor of pinging Redis on startup, as this improves the on-premise first installation experience.

sileht (approved):
I merging this as ticket has been created for the other refactor.

PR #4901, opened 2022-06-21
feat(.mergify.yml): Add automerge for python version bump

Greesb:


jd (approved):

.mergify.yml
```diff
@@ -72,6 +72,22 @@ pull_request_rules:
...
 
+  - name: automatic merge for python version bump
+    conditions:
+      - and: *CheckRuns
+      - check-success=semantic-pull-request
+      - check-success=constraints
+      - author=mergify-ci-bot
+      - label!=work-in-progress
+      - label!=manual merge
+      - "title~=^chore: bump python version"
```
sileht, line 83: I would replace this with `head: python-version-update`

sileht (changes requested):

.mergify.yml
Greesb: True, added it. I let the title check as well just for more security

PR #4902, opened 2022-06-21
feat(eventlogs): add what trigger the action

sileht:
This tracks and restitute what has triggered the action.

Fixes MRGFY-1241

Change-Id: I0f484d8f6fc2866134e0c1e3c4af687b1f2231bb


Greesb (commented):
LGTM but i'll let somebody else review in case i missed something business side

PR #4915, opened 2022-06-22
feat(signals): track more queue events

sileht:
This adds queue.check_start, queue.check_end, queue.enter and queue.leave.

Fixes MRGFY-1258
Fixes MRGFY-1239

Change-Id: Ie59c15ca292dc8e99be210b4fc2ba7270c83cfe8

### DIFF
--- a/mergify_engine/queue/merge_train.py
+++ b/mergify_engine/queue/merge_train.py
@@ -35,6 +35,7 @@ from mergify_engine import date
 from mergify_engine import github_types
 from mergify_engine import json
 from mergify_engine import queue
+from mergify_engine import signals
 from mergify_engine import utils
 from mergify_engine.clients import http
 from mergify_engine.dashboard import subscription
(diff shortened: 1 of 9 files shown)


mergify_engine/queue/merge_train.py
```diff
@@ -846,6 +872,41 @@ async def delete_pull(
                 )
                 tmp_pull_ctxt.log.info("train car deleted", reason=reason)
 
+        if reason is None:
+            if self.has_timed_out:
+                aborted = True
+                abort_reason = "Checks has timed out"
```
jd, line 878:
```suggestion
                abort_reason = "Checks have timed out"
```

mergify_engine/signals.py
```diff
@@ -39,6 +40,10 @@
     "action.label",
     "action.merge",
     "action.post_check",
+    "action.queue.enter",
+    "action.queue.check_start",
```
jd, line 44: check or checks?

jd (changes requested):

mergify_engine/signals.py
sileht: checks is used every where so

mergify_engine/queue/merge_train.py
```diff
@@ -1574,11 +1635,14 @@ async def _slice_cars(
...
-    ) -> typing.Optional[EmbarkedPullWithCar]:
-        return first.first(
-            self._iter_embarked_pulls(),
-            key=lambda c: c.embarked_pull.user_pull_request_number == pull_number,
-        )
+    ) -> typing.Union[
+        typing.Tuple[int, EmbarkedPullWithCar],
+        typing.Tuple[typing.Literal[None], typing.Literal[None]],
+    ]:
+        for position, c in enumerate(self._iter_embarked_pulls()):
```
Greesb, line 1642: Isn't there a better name than just `c` ? Would be better for understanding the code more easily

mergify_engine/queue/merge_train.py
```diff
@@ -1689,26 +1755,45 @@ async def add_pull(
             )
 
         await self.save()
+
+        final_position = await self.get_position(ctxt)
+        if final_position is None:
+            raise RuntimeError("just added pull request in queue not found")
```
Greesb, line 1761:
```suggestion
            raise RuntimeError("Could not find the pull request we just added in the queue")
```

mergify_engine/signals.py
```diff
@@ -234,6 +274,50 @@ async def send(
...
+    pull_request: github_types.GitHubPullRequestNumber,
+    event: typing.Literal["action.queue.checks_end"],
+    metadata: EventQueueChecksEndMetadata,
+    trigger: str,
+) -> None:
+    ...
+
+
+@typing.overload
+async def send(
```
Greesb, line 311: What is the point of all those overload ? Since every metadata specified is a child of EventMetadata and the event name can be added in the EventName literal type at the top, calling the base `send` at the bottom of this file should work without having to make all those overloads, no ?

Greesb (changes requested):

jd (approved):

PR #4929, opened 2022-06-23
fix(queue): get rid of stored queue_config

sileht:
Using the stored one led to race conditions when the user just changed it
configuration and expected the queue to behave like the new config.

As soon the new config is there, logic should read the config only from
the one in the configuration file, not an old one.

Fixes MRGFY-1261

Change-Id: I29d4fb4a78bba86a3e50978e7e76dcb8e8a678bc


mergify_engine/queue/merge_train.py
```diff
@@ -1439,6 +1422,14 @@ async def get_queue_rules(self) -> typing.Optional["rules.QueueRules"]:
 
         return mergify_config["queue_rules"]
 
+    async def get_queue_rule(self, queue_name: "rules.QueueName") -> "rules.QueueRule":
+        rules = await self.get_queue_rules()
+
+        if not rules or queue_name not in [rule.name for rule in rules]:
+            raise RuntimeError("Rules are missing from configuration")
```
Greesb, line 1429: Can't we add more details on this error ? Like the queue name and/or the rule name ?

Greesb (changes requested):

mergify_engine/tests/functional/actions/test_queue.py
```diff
@@ -3007,6 +3008,174 @@ async def test_queue_manual_merge(self):
...
+                    [p2["number"]],
+                    [p1["number"]],
+                    p_new_config["merge_commit_sha"],
+                    "created",
+                    tmp_mq_p2["number"],
+                ),
+            ],
+        )
+
+        # Put second PR at the begining of the queue via pr priority Checks
```
Syffe, line 3139:
```suggestion
        # Put third PR at the begining of the queue via pr priority Checks
```
it is the third right ? otherwise LGTM

sileht: I will fix that into another PR

Greesb (approved):

PR #4931, opened 2022-06-23
feat(worker): preload/reload redis scripts dynamically on client connection

Syffe:
To ensure the good loading of redis scripts all along the engine workflow, at any time,
and avoid worker failures, we needed to setup a way of loading them dynamically.
These changes have also lead to the removing of all migrations scripts and tests.

Fixes MRGFY-1259

### DIFF
--- a/mergify_engine/redis_utils.py
+++ b/mergify_engine/redis_utils.py
@@ -45,7 +45,7 @@ RedisEventLogs = typing.NewType("RedisEventLogs", "redispy.Redis[bytes]")
 
 ScriptIdT = typing.NewType("ScriptIdT", uuid.UUID)
 
-SCRIPTS: typing.Dict[ScriptIdT, typing.Tuple[str, str]] = {}
+SCRIPTS: typing.Dict[ScriptIdT, typing.Tuple[bytes, str]] = {}
 
 
 # TODO(sileht): Redis script management can be moved back to Redis.register_script() mechanism
@@ -57,7 +57,9 @@ def register_script(script: str) -> ScriptIdT:
         # NOTE(sileht): SHA1 is imposed by Redis itself
         hashlib.sha1(  # nosemgrep contrib.dlint.dlint-equivalent.insecure-hashlib-use, python.lang.security.insecure-hash-algorithms.insecure-hash-algorithm-sha1
             script.encode("utf8")
-        ).hexdigest(),
+        )
+        .hexdigest()
+        .encode(),
         script,
     )
     return script_id
@@ -67,12 +69,12 @@ def register_script(script: str) -> ScriptIdT:
 # it works but if a script is loaded into two redis, this won't works as expected
 # as the app will think it's already loaded while it's not...
 async def load_script(
-    redis: typing.Union[RedisCache, RedisStream], script_id: ScriptIdT
+    redis: "redispy.connection.Connection", script_id: ScriptIdT
 ) -> None:
     global SCRIPTS
     sha, script = SCRIPTS[script_id]
-    # FIXME(sileht): weird, this method is typed on redis-py
-    newsha = await redis.script_load(script)  # type: ignore[no-untyped-call]
+    await redis.send_command("SCRIPT LOAD", script)
+    newsha = await redis.read_response()
     if newsha != sha:
         LOG.error(
             "wrong redis script sha cached",
@@ -83,7 +85,7 @@ async def load_script(
         SCRIPTS[script_id] = (newsha, script)
 
 
-async def load_scripts(redis: typing.Union[RedisCache, RedisStream]) -> None:
+async def load_stream_scripts(redis: "redispy.connection.Connection") -> None:
     # TODO(sileht): cleanup unused script, this is tricky, because during
     # deployment we have running in parallel due to the rolling upgrade:
     # * an old version of the asgi server
@@ -93,9 +95,13 @@ async def load_scripts(redis: typing.Union[RedisCache, RedisStream]) -> None:
     scripts = list(SCRIPTS.items())  # order matter for zip bellow
     shas = [sha for _, (sha, _) in scripts]
     ids = [_id for _id, _ in scripts]
-    exists = await redis.script_exists(*shas)  # type: ignore[no-untyped-call]
+    await redis.on_connect()
+    await redis.send_command("SCRIPT EXISTS", *shas)
+
+    # exists is a list of 0 and/or 1, notifying the existence of each script
+    exists = await redis.read_response()
     for script_id, exist in zip(ids, exists):
-        if not exist:
+        if exist == 0:
             await load_script(redis, script_id)
 
 
@@ -137,11 +143,15 @@ class RedisLinks:
 
     @functools.cached_property
     def stream(self) -> RedisStream:
-        client = self.redis_from_url(
+        # Note(Syffe): mypy struggles to recognize the type of load_scripts because typeshed is missing
+        # typing on the objects we use here.
+        # cf: https://github.com/python/typeshed/pull/8147
+        client = self.redis_from_url(  # type: ignore[call-overload]
             "stream",
             config.STREAM_URL,
             decode_responses=False,
             max_connections=self.stream_max_connections,
+            redis_connect_func=load_stream_scripts,
         )
         return RedisStream(client)
 
@@ -191,7 +201,6 @@ class RedisLinks:
             "active_users",
             config.ACTIVE_USERS_URL,
             decode_responses=False,
-            max_connections=None,
         )
         return RedisActiveUsers(client)
 
@@ -212,6 +221,9 @@ class RedisLinks:
         url: str,
         decode_responses: typing.Literal[True],
         max_connections: typing.Optional[int] = None,
+        redis_connect_func: typing.Optional[
+            "redispy.connection.ConnectCallbackT"
+        ] = None,
     ) -> "redispy.Redis[str]":
         ...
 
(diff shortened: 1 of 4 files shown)


jd (dismissed):
I'm scared this is going to break in production. The code here:
https://github.com/redis/redis-py/blob/master/redis/connection.py#L621-L626

seems to indicate no authentication will be done if we pass our own custom function.

mergify_engine/redis_utils.py
```diff
@@ -231,6 +250,12 @@ def redis_from_url(
         url: str,
         decode_responses: bool,
         max_connections: typing.Optional[int] = None,
+        redis_connect_func: typing.Optional[
```
Greesb, line 253: Maybe you can create a custom TypeVar to only have to write that once ?

Greesb (changes requested):

mergify_engine/redis_utils.py
```diff
@@ -66,13 +66,14 @@ def register_script(script: str) -> ScriptIdT:
...
-    redis: typing.Union[RedisCache, RedisStream], script_id: ScriptIdT
-) -> None:
+async def load_script(redis: redispy.Connection, script_id: ScriptIdT) -> None:
     global SCRIPTS
     sha, script = SCRIPTS[script_id]
     # FIXME(sileht): weird, this method is typed on redis-py
-    newsha = await redis.script_load(script)  # type: ignore[no-untyped-call]
+    await redis.send_command("SCRIPT LOAD", script)
+    response = await redis.read_response()
+    newsha = response if isinstance(response, str) else response.decode("utf-8")
```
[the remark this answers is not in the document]
sileht: it depends of if the connection is opened with decode_responses=True or not, just guessing.

mergify_engine/redis_utils.py
```diff
@@ -66,13 +66,14 @@ def register_script(script: str) -> ScriptIdT:
 # FIXME(sileht): We store Cache and Stream script into the same global object
 # it works but if a script is loaded into two redis, this won't works as expected
 # as the app will think it's already loaded while it's not...
-async def load_script(
-    redis: typing.Union[RedisCache, RedisStream], script_id: ScriptIdT
-) -> None:
+async def load_script(redis: redispy.Connection, script_id: ScriptIdT) -> None:
     global SCRIPTS
     sha, script = SCRIPTS[script_id]
     # FIXME(sileht): weird, this method is typed on redis-py
```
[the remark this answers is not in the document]
Syffe: Nope indeed, I removed it !

mergify_engine/redis_utils.py
Syffe: Indeed good idea, too many repetitions !

mergify_engine/redis_utils.py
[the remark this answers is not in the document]
Syffe: Indeed we can also check it with `decode_responses`. It depends on what connection is being processed here, (RedisCache which is a type Redis[str] or RedisStream which is a type Redis[bytes]). We still need to check which one is being processed in order to get the `newsha` though. 

I thought about creating new types for each connection, but still we would need to check which one is being processed here.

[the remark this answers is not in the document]
Syffe: I changed the way we check it using the parameter `decode_responses` instead

jd (commented):
This loads all scripts in every Redis db, I'm not sure we really need to do this. Do we want to only load scripts where there are needed?

mergify_engine/redis_utils.py
```diff
@@ -212,6 +224,7 @@ def redis_from_url(
         url: str,
         decode_responses: typing.Literal[True],
         max_connections: typing.Optional[int] = None,
+        redis_connect_func: LoadScripts = None,
```
sileht, line 227:
```suggestion
        redis_connect_func: typing.Optional["redis.asyncio.connection.ConnectCallbackT"] = None,
```

mergify_engine/redis_utils.py
```diff
@@ -222,6 +235,7 @@ def redis_from_url(
         url: str,
         decode_responses: typing.Literal[False],
         max_connections: typing.Optional[int] = None,
+        redis_connect_func: LoadScripts = None,
```
sileht, line 238:
```suggestion
        redis_connect_func: typing.Optional["redis.asyncio.connection.ConnectCallbackT"] = None,
```

mergify_engine/redis_utils.py
```diff
@@ -231,6 +245,7 @@ def redis_from_url(
         url: str,
         decode_responses: bool,
         max_connections: typing.Optional[int] = None,
+        redis_connect_func: LoadScripts = None,
```
sileht, line 248:
```suggestion
        redis_connect_func: typing.Optional["redis.asyncio.connection.ConnectCallbackT"] = None,
```

mergify_engine/redis_utils.py
```diff
@@ -202,6 +213,7 @@ def cache(self) -> RedisCache:
             config.LEGACY_CACHE_URL,
             decode_responses=True,
             max_connections=self.cache_max_connections,
+            redis_connect_func=load_scripts,
```
sileht, line 216: I don't think this is required anymore, since we drop the migration scripts, we don't need them here anymore.

sileht (changes requested):

Syffe:
> This loads all scripts in every Redis db, I'm not sure we really need to do this. Do we want to only load scripts where there are needed?

In the current case, we only load scripts for RedisStream, each time a connection is made to it, the other Redis db are not concerned. 

If we load the scripts where they are needed, how can we load them dynamically then ?

mergify_engine/redis_utils.py
```diff
@@ -67,12 +67,13 @@ def register_script(script: str) -> ScriptIdT:
...
-    redis: typing.Union[RedisCache, RedisStream], script_id: ScriptIdT
+    redis: "redispy.connection.Connection", script_id: ScriptIdT
 ) -> None:
     global SCRIPTS
     sha, script = SCRIPTS[script_id]
-    # FIXME(sileht): weird, this method is typed on redis-py
-    newsha = await redis.script_load(script)  # type: ignore[no-untyped-call]
+    await redis.send_command("SCRIPT LOAD", script)
+    response = await redis.read_response()
+    newsha = response.decode("utf-8")
```
jd, line 76: 1. decoding as ASCII might be enough
2. you could rather encode the sha1 computed in L60; that'd make encoding once and compare N times, rather than encoding+compare N times

mergify_engine/redis_utils.py
```diff
@@ -67,12 +67,13 @@ def register_script(script: str) -> ScriptIdT:
...
 # as the app will think it's already loaded while it's not...
 async def load_script(
-    redis: typing.Union[RedisCache, RedisStream], script_id: ScriptIdT
+    redis: "redispy.connection.Connection", script_id: ScriptIdT
 ) -> None:
     global SCRIPTS
     sha, script = SCRIPTS[script_id]
-    # FIXME(sileht): weird, this method is typed on redis-py
-    newsha = await redis.script_load(script)  # type: ignore[no-untyped-call]
+    await redis.send_command("SCRIPT LOAD", script)
```
jd, line 74: what's the rational for not using `script_load` anymore?

mergify_engine/redis_utils.py
```diff
@@ -93,9 +94,13 @@ async def load_scripts(redis: typing.Union[RedisCache, RedisStream]) -> None:
     scripts = list(SCRIPTS.items())  # order matter for zip bellow
     shas = [sha for _, (sha, _) in scripts]
     ids = [_id for _id, _ in scripts]
-    exists = await redis.script_exists(*shas)  # type: ignore[no-untyped-call]
+    await redis.on_connect()
+    await redis.send_command("SCRIPT EXISTS", *shas)
```
jd, line 98: what's the rational for not using `script_exists` anymore?

mergify_engine/redis_utils.py
```diff
@@ -132,16 +137,21 @@ def queue(self) -> RedisQueue:
...
-        client = self.redis_from_url(
+        # Note(Syffe): mypy struggles to recognize the type of load_scripts because typeshed is missing
+        # typing on the objects we use here.
+        # cf: https://github.com/python/typeshed/pull/8147
+        client = self.redis_from_url(  # type: ignore[call-overload]
             "stream",
             config.STREAM_URL,
             decode_responses=False,
             max_connections=self.stream_max_connections,
+            redis_connect_func=load_scripts,
```
jd, line 154:
```suggestion
            redis_connect_func=load_stream_scripts,
```
might be clearer so we know later we need to have a function per connection

mergify_engine/redis_utils.py
```diff
@@ -162,6 +173,7 @@ def team_permissions_cache(self) -> RedisTeamPermissionsCache:
             config.TEAM_PERMISSIONS_CACHE_URL,
             decode_responses=False,
             max_connections=self.cache_max_connections,
+            redis_connect_func=None,
```
jd, line 176: this is the default, why setting it?

jd (dismissed):
LGTM, just a few details

mergify_engine/redis_utils.py
Syffe: because when using `redis_connect_func` parameter, what is passed as an argument to the function is a redis `Connection` object and not a redis `Client`. So script_load is not defined for a `Connection` object, instead, we leverage to send direct commands through the `Connection`

mergify_engine/redis_utils.py
Syffe: same as the upper explanation

mergify_engine/redis_utils.py
sileht: I never found how to please mypy when the method uses @overload and not (re)set the default on every calls.

sileht (commented):
Looks good to me

mergify_engine/redis_utils.py
sileht: Cool!

mergify_engine/redis_utils.py
sileht: Done

jd (approved):

PR #4942, opened 2022-06-25
ci(build): replace setuptools by poetry

sileht:
This change moves build system to poetry. The base package has been
converted by dephell tool. Then tox.ini has been adapted to install
locked deps with poetry instead of pip.

Version of tox and poetry are tracked in requirements-dev.txt

Change-Id: I23e681f8d1f71156bfb7cfeba6bc9fa5c7afda26


requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+#NOTE(sileht): we separe them, so Docker image can only pick poetry-version.txt
```
jd, line 1:
```suggestion
# NOTE(sileht): we separate them, so Docker image can only pick poetry-version.txt
```

jd (approved):

jd:
@mergifyio squash

PR #4964, opened 2022-06-28
feat(eventlogs) add pagination

sileht:
Fixes MRGFY-1190

Change-Id: Ia4c12e51b4fc3cf3580b8f1cad419e0ac5e99a4a


mergify_engine/eventlogs.py
```diff
@@ -297,76 +303,137 @@ async def get(
...
     older_event = date.utcnow() - retention
+    older_event_id = f"{int(older_event.timestamp())}"
+
+    pipe = await redis.pipeline()
+
+    if not page.cursor:
+        # first page
+        from_ = "+"
+        to_ = older_event_id
+        look_behind = False
```
jd, line 314: did you mean backward rather than behind?

jd (approved):

PR #4978, opened 2022-06-29
fix(summary): handle base changeable attributes in rules conditions

Syffe:
Due to a change from GitHub, now allowing the change of the base branch in a PR,
we had to modify how we handle these new types of changeable base attributes. Those
changes now avoid to break the summary when those rule conditions are not met.

Fixes MRGFY-1183


mergify_engine/rules/__init__.py
```diff
@@ -170,11 +170,17 @@ class GenericRulesEvaluator(typing.Generic[T_Rule, T_EvaluatedRule]):
...
         "author",
         "merged_by",
     )
 
+    # special attributes that are looked for the next matching rules, but
+    # should not appear in summary.
+    # Note(Syffe): this list was created after GitHub integrated the possibility to change
+    # the base branch of a PR. It invalidated the unchangeable state of this base attribute.
+    # and thus we needed to process it separately from the other base attributes.
+    SPECIAL_ATTRIBUTES = ("base",)
```
jd, line 182:
```suggestion
    # Special attributes that are looked for in the next matching rules, but
    # should not appear in summary.
    # Note(Syffe): this list was created after GitHub integrated the possibility to change
    # the base branch of a PR. It invalidated the unchangeable state of this base attribute.
    # and thus we needed to process it separately from the other base attributes.
    BASE_CHANGEABLE_ATTRIBUTES = ("base",)
```

jd (dismissed):
"special" is not really explicit 😁

mergify_engine/rules/__init__.py
```diff
@@ -256,19 +267,23 @@ async def create(
                     attr = condition.get_attribute_name()
                     if attr not in self.BASE_ATTRIBUTES:
                         condition.update("number>0")
+                    if attr in self.SPECIAL_ATTRIBUTES and (
+                        evaluated_rule not in self.special_attributes_rules
+                    ):
+                        self.special_attributes_rules.append(evaluated_rule)
```
sileht, line 273: We should use the same logic as BASE_ATTRIBUTES for SPECIAL_ATTRIBUTES, to not just check if the special attribute is present, but if the binary tree can be true if the tree only has these attributes.

Imagine this tree: 
```
{"or": [
    "label: automerge",
    "base: foobar"
 ]}
 ```
 This rules should not be categorized in special_attributes_rules as we have a `or` with a label.

File: mergify_engine/engine/actions_runner.py [hunk 3]
```diff
@@ -105,26 +105,53 @@ async def gen_summary(
...
+
+    if special_attributes_rules > 0:
+        for rule in match.special_attributes_rules:
+            if rule in match.matching_rules:
+                base_conditions = rule.conditions.copy()
+                for condition in base_conditions.walk():
+                    attr = condition.get_attribute_name()
+                    if attr in match.SPECIAL_ATTRIBUTES and not condition.match:
+                        not_applicable_special_attributes_rules_list.append(rule)
+                        match.matching_rules.remove(rule)
```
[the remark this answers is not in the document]
Syffe: Okay ! 
There will be duplicates with rules that are in `potential_rules` and `not_applicable` though

mergify_engine/engine/actions_runner.py
```diff
@@ -522,8 +558,8 @@ async def cleanup_pending_actions_with_no_associated_rules(
             check_name,
             check_api.Result(
                 check_api.Conclusion.CANCELLED,
-                "The rule/action does not exist anymore",
-                "",
+                "The rule doesn't match anymore",
+                "This action has been cancelled.",
             ),
```
sileht, line 563: You can import and use the check_api.Result directly from actions.__init__

mergify_engine/engine/actions_runner.py
```diff
@@ -105,26 +105,62 @@ async def gen_summary(
     match: rules.RulesEvaluator,
 ) -> typing.Tuple[str, str]:
     summary = ""
-    summary += await get_already_merged_summary(ctxt, match)
+    base_changeable_attributes_rules = len(match.base_changeable_attributes_rules)
+    not_applicable_base_changeable_attributes_rules = []
```
sileht, line 109: Can we build this is in RuleEvaluator.create like other lists ?

mergify_engine/rules/__init__.py
```diff
@@ -247,28 +258,31 @@ async def create(
...
+                        attr not in self.BASE_CHANGEABLE_ATTRIBUTES
+                    ):
                         condition.update("number>0")
 
                 await base_conditions(pulls)
 
                 if not base_conditions.match:
-                    self.ignored_rules.append(typing.cast(T_EvaluatedRule, rule))
+                    self.base_changeable_attributes_rules.append(evaluated_rule)
+                    self.ignored_rules.append(evaluated_rule)
```
sileht, line 277: The two lists look to be the same, no?

I think you can directly build the `not_applicable_base_changeable_attributes_rules` here?

sileht (changes requested):

mergify_engine/engine/actions_runner.py
```diff
@@ -106,25 +106,42 @@ async def gen_summary(
...
+        if not_applicable_count == 1:
+            summary += (
+                f"<summary>{not_applicable_count} not applicable rule</summary>\n\n"
+            )
         else:
-            summary += f"<summary>{ignored_rules} not applicable rules</summary>\n\n"
-        summary += await gen_summary_rules(ctxt, match.ignored_rules)
+            summary += (
+                f"<summary>{not_applicable_count} not applicable rules</summary>\n\n"
+            )
```
Greesb, line 135: I don't see any difference between the `if` and the `else`'s content

Greesb (changes requested):

mergify_engine/engine/actions_runner.py
Greesb, line 135: I don't see any differences between the if and the else's content

Greesb (changes requested):

mergify_engine/rules/__init__.py
```diff
@@ -247,28 +258,46 @@ async def create(
...
             if hide_rule and not rule.conditions.match:
                 # NOTE(sileht): Replace non-base attribute by true, if it
                 # still matches it's a potential rule otherwise hide it.
                 base_conditions = rule.conditions.copy()
                 for condition in base_conditions.walk():
                     attr = condition.get_attribute_name()
-                    if attr not in self.BASE_ATTRIBUTES:
+                    if (attr not in self.BASE_ATTRIBUTES) and (
+                        attr not in self.BASE_CHANGEABLE_ATTRIBUTES
+                    ):
```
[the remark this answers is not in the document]
Greesb: Otherwise LGTM

Greesb (commented):
LGTM

sileht (dismissed):

mergify_engine/tests/functional/test_summary.py
```diff
@@ -0,0 +1,107 @@
...
+        p = await self.create_pr()
+        await self.add_label(p["number"], "test")
+        await self.run_engine()
+
+        ctxt = await context.Context.create(self.repository_ctxt, p, [])
+
+        summary = await ctxt.get_engine_check_run(constants.SUMMARY_NAME)
+        assert summary is not None
+        assert summary["output"]["title"] == "1 rule matches"
+        assert "1 not applicable rule" in summary["output"]["summary"]
```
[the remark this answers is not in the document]
Syffe: The rule with label does apply here, the rule that doesn't apply is the one named `invalid rule dummy`

mergify_engine/rules/__init__.py
```diff
@@ -241,34 +252,48 @@ async def create(
...
+                await base_changeable_conditions(pulls)
                 await base_conditions(pulls)
 
-                if not base_conditions.match:
-                    self.ignored_rules.append(typing.cast(T_EvaluatedRule, rule))
+                if not base_changeable_conditions.match:
+                    self.not_applicable_base_changeable_attributes_rules.append(
+                        evaluated_rule
+                    )
+                    categorized = True
```
sileht, line 285: We should not categorize them as we want them to be in the matching_rules set to have their actions running. This set is here only to the Summary UI to not show them if they don't pass.

sileht (changes requested):

mergify_engine/rules/__init__.py
Syffe: I don't understand, in this context if the rule is in the `not_applicable_base_changeable_attributes_rules` we are sure that it is not gonna pass due to the false `base_changeable_attributes` right ?

mergify_engine/engine/actions_runner.py
```diff
@@ -107,38 +107,70 @@ async def gen_summary(
...
+    for rule in match.matching_rules:
+        if rule in match.not_applicable_base_changeable_attributes_rules:
+            conditions = rule.conditions.copy()
+            for condition in conditions.walk():
+                attr = condition.get_attribute_name()
+                if attr in match.BASE_CHANGEABLE_ATTRIBUTES and not condition.match:
+                    matching_rules_to_display.remove(rule)
+                    not_applicable_base_changeable_attributes_rules_to_display.append(
+                        rule
+                    )
```
sileht, line 121: I don't see why we check BASE_CHANGEABLE_ATTRIBUTES, I think this should be enough:

```suggestion
    for rule in match.matching_rules:
        if rule in match.not_applicable_base_changeable_attributes_rules:
                matching_rules_to_display.remove(rule)
                not_applicable_base_changeable_attributes_rules_to_display.append(
                    rule
                )
```

sileht (changes requested):

PR #4980, opened 2022-06-29
fix(eventlogs): Don't enforce that all fields should be there

sileht:
All recorded events may not have all fields we currently have in the
code.

This change makes all of them optional.

Fixes MERGIFY-ENGINE-2R1

### DIFF
--- a/mergify_engine/signals.py
+++ b/mergify_engine/signals.py
@@ -56,7 +56,7 @@ EventName = typing.Literal[
 ]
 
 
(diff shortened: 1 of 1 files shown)


jd (approved):

sileht:
I'm not a huge fan of the solution. Still looking for some idea that doesn't force us to do that.
Using default to field missing data doesn't really work either, for queued_at or branch_name I can put anything logic.

PR #4985, opened 2022-06-30
feat(review): signal who we impersonate

sileht:
Fixes MRGFY-1275

Change-Id: I35c719bb11e6a5a03e10b84fd23d9538d781cff6


jd (dismissed):
no test/assert?

jd (approved):

PR #5001, opened 2022-07-04
fix(command): add protection against user edition

sileht:
This changes may increase a bit load as we will now handle delete and
updated comments.

Fixes MRGFY-1256

Change-Id: I7255c35c8a30c6815c937a2460b5f40f1a4ed79d


mergify_engine/engine/commands_runner.py
```diff
@@ -196,12 +225,24 @@ async def run_pending_commands_tasks(
...
+                pendings.add(command)
+            else:
+                edited.add(command)
+        else:
+            edited.discard(command)
+            pendings.discard(command)
+
+    for edit in edited:
+        message = prepare_message(
+            f"@Mergifyio {edit}",
```
Greesb, line 238: Don't you have the whole text message above ? You could use it instead of reformatting the message here, no ?

Greesb (changes requested):

jd (approved):

mergify_engine/engine/commands_runner.py
sileht: Good catch, this was even a bug.

PR #5007, opened 2022-07-05
fix: clean config file location cache only when needed

sileht:
Instead of clearing the cache on all push (even each time a pull request
from the same repo is created/updated) we should only clear it when the
default branch change.

Fixes MRGFY-1289

Change-Id: I6ee997a8d83586af296ca90c03809e1146a9efb7


mergify_engine/github_events.py
```diff
@@ -280,7 +291,30 @@ async def push_to_worker(
         elif event["repository"]["archived"]:  # pragma: no cover
             ignore_reason = "repository archived"
 
-        await context.Repository.clear_config_file_cache(redis_links.cache, repo_id)
+        if f"refs/heads/{event['repository']['default_branch']}" == event["ref"]:
+
+            if event["forced"]:
+                mergify_configuration_changed = True
+            else:
+                mergify_configuration_changed = False
```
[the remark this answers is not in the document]
sileht: The `else` has more than just setting it to `false`.

mergify_engine/github_events.py
```diff
@@ -280,7 +291,30 @@ async def push_to_worker(
         elif event["repository"]["archived"]:  # pragma: no cover
             ignore_reason = "repository archived"
 
-        await context.Repository.clear_config_file_cache(redis_links.cache, repo_id)
+        if f"refs/heads/{event['repository']['default_branch']}" == event["ref"]:
+
+            if event["forced"] or len(event["commits"]) > 250:
```
jd, line 296: why 250?

sileht: GitHub API limit.

jd (dismissed):

mergify_engine/github_events.py
```diff
@@ -280,7 +291,31 @@ async def push_to_worker(
...
+                    itertools.chain.from_iterable(
+                        itertools.chain(
+                            commit["added"], commit["removed"], commit["modified"]
+                        )
+                        for commit in event["commits"]
+                    )
+                )
+                mergify_configuration_changed = bool(
+                    set(constants.MERGIFY_CONFIG_FILENAMES) & files
+                )
```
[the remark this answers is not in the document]
sileht: We don't care 250*3

mergify_engine/github_events.py
```diff
@@ -280,7 +290,30 @@ async def push_to_worker(
...
+                mergify_configuration_changed = True
+            else:
+                mergify_configuration_changed = False
+
+                mergify_config_filenames = set(constants.MERGIFY_CONFIG_FILENAMES)
+                for commit in event["commits"]:
+                    if (
+                        set(commit["added"] + commit["removed"] + commit["modified"])
+                        & mergify_config_filenames
+                    ):
```
jd, line 309:
```suggestion
                    if (
                        set(commit["added"] & mergify_config_filenames or commit["removed"] & mergify_config_filenames or set(commit["modified"]) & mergify_config_filenames
                    ):
```
if you want to optimize further

jd (dismissed):

mergify_engine/github_events.py
sileht: even better

jd (dismissed):

jd (approved):

mergify_engine/tests/functional/test_engine.py
```diff
@@ -798,11 +798,12 @@ async def test_pull_request_init_summary(self) -> None:
         await self.setup_repo(yaml.dump(rules))
 
         # Run the engine once, to initialiaze the config location cache
-        await self.create_pr()
+        p = await self.create_pr()
```
jd, line 801: unused?

PR #5014, opened 2022-07-06
chore(mergify): remove approval on new commits

jd:
Now that there more regular contributors, there's no need to keep
approval for ever. We can re-request a review.

This should prevent issues where wrong push could be merged by
mistake.


sileht (approved):
Welcome to the new world!

PR #5025, opened 2022-07-08
fix: decrease verbosity of "pull request train car status update" message

DouglasBlackwood:
Log useful attributes, instead of the whole report, to be correctly parsed by Datadog.

Fixes MRGFY-879

### DIFF
--- a/mergify_engine/queue/merge_train.py
+++ b/mergify_engine/queue/merge_train.py
@@ -1258,21 +1258,30 @@ You don't need to do anything. Mergify will close this pull request automaticall
         if self.has_timed_out:
             conclusion = check_api.Conclusion.FAILURE
 
-        report = check_api.Result(
-            conclusion,
-            title=original_pull_title,
-            summary=unexpected_change_summary
+        check_api_summary = (
+            unexpected_change_summary
             + queue_summary
             + "\n"
             + checks_copy_summary
             + "\n"
-            + batch_failure_summary,
+            + batch_failure_summary
         )
+
+        report = check_api.Result(
+            conclusion,
+            title=original_pull_title,
+            summary=check_api_summary,
+        )
+
         for original_ctxt in original_ctxts:
             original_ctxt.log.info(
                 "pull request train car status update",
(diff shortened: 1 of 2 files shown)


Greesb (commented):
LGTM

mergify_engine/queue/merge_train.py
```diff
@@ -1258,21 +1258,30 @@ async def update_summaries(
...
         for original_ctxt in original_ctxts:
             original_ctxt.log.info(
                 "pull request train car status update",
                 conclusion=conclusion.value,
-                report=report,
+                original_pull_title=original_pull_title,
+                checked_pull=checked_pull,
+                unexpected_change_summary=unexpected_change_summary.strip(),
+                queue_summary=queue_summary.strip(),
+                batch_failure_summary=batch_failure_summary.strip(),
```
sileht, line 1284: We should replace the `checks_copy_summary` part by something`{check.name: check.state for check in self.last_checks}` to get the same content but without the html.

mergify_engine/queue/merge_train.py
```diff
@@ -1258,21 +1258,30 @@ async def update_summaries(
...
         )
+
+        report = check_api.Result(
+            conclusion,
+            title=original_pull_title,
+            summary=check_api_summary,
+        )
+
         for original_ctxt in original_ctxts:
             original_ctxt.log.info(
```
sileht, line 1277: We should also move the logging out of the loop (you can use `self.train.log` instead).

mergify_engine/queue/merge_train.py
```diff
@@ -1258,21 +1258,30 @@ async def update_summaries(
         if self.has_timed_out:
             conclusion = check_api.Conclusion.FAILURE
 
-        report = check_api.Result(
-            conclusion,
-            title=original_pull_title,
-            summary=unexpected_change_summary
+        check_api_summary = (
```
sileht, line 1261:
```suggestion
        original_pull_summary = (
```

sileht (dismissed):

